PR #101, opened 2019-03-06
Make this app immune against any exploit(fixing #61).

### LINKED ISSUE #61: Update Calculator engine’s use of smart pointers
The Calculator engine allocates and frees its own memory which leads to issues such as double-frees that are hard to investigate.  We should clean this code and switch to modern constructs such as smart_ptrs to eliminate these issues.

jens1o:
- Removed all source files.

Well, I guess it is fine that it doesn't built anymore?

[In the recent years, many Proof of Concepts in regards to exploits where presented with the calculator app launching (or being launched by) an arbitrary application](https://www.youtube.com/watch?v=iaIkc203Fd0). When we no longer have the calculator, many PoCs are broken, thus we also need to [worry less about exploits](https://twitter.com/Mythic3d/status/1103364806579372032). :woman_shrugging::man_shrugging:

---

Please note that this Pull Request is not meant at all for merging, it is just meant to maybe draw some smiles. I find it generally awesome that you are (apparently) committed to Free Software by making core programs of an operating system open to the public, but reading about this exact repository the first time immediately brought me the idea of saving many lifes. (That is why I only removed the C++/header files, to leave the general idea behind. `rm -rf *` would have been easier.)

I think this is kind of the best way to reach out to you. I'll eventually delete my fork, as a Linux-based user I'm not able to use this project.

:Kappa:


Volxz:
I think this should be merged. Big security @jens1o

shengslogar:
While your sense of humor is appreciated, this just doesn't seem to be the right place for it.

Airblader:
This kind of humor might make you a target for hackers. Better delete your GitHub account now to defend against that... :-)

PR #110, opened 2019-03-06
Fix scientific notation for crypto currencies

uhliksk:
## Fixes #.
It's not helpful to calculate crypto currency values with scientific notation. All values up to 7 decimals should be in standard format.

- Changed maximum number of zeros after decimal to 7 (smallest unit "1 satoshi" in crypto currencies have 7 zeros).

- manual

### DIFF
--- a/src/CalcManager/Ratpack/conv.cpp
+++ b/src/CalcManager/Ratpack/conv.cpp
@@ -22,7 +22,7 @@
 
 using namespace std;
 
-static constexpr int MAX_ZEROS_AFTER_DECIMAL = 2;
+static constexpr int MAX_ZEROS_AFTER_DECIMAL = 7;
 
 // digits 0..64 used by bases 2 .. 64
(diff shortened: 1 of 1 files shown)


Agusx1211:
This change seems big and specific to work only with Bitcoin-related amounts, and ignoring the fact that some cryptos have more than 8 decimals, Ethereum has 18.

I may recommend changing this to a feature like "disable scientific notation", if not already present.

grochocki:
Thanks so much for the submission. This looks like a useful improvement for cryptocurrency scnearios, however due to the scope of impact that any new change might have in app given its large userbase, we've established a [New Feature Process](https://github.com/Microsoft/calculator/blob/master/docs/NewFeatureProcess.md) to help us maintain our quality bar and a consistent user experience. We're going to close out this pull request for now and give you a chance to create an issue so the community can discuss it further.

uhliksk:
I know userbase is wide but there is "F-E" button to enable scientific notation and I think it is counterproductive to force scientific notation with disabled scientific notation just because there is more than predefined number of neros. I just wanted to make it little bit less annoying when scientific notation is disabled. Thank you for your opinion.

PR #121, opened 2019-03-06
Add explicit [[fallthrough]] attribute

janisozaur:
Instead of relying on comment, have the C++17's attribute ensure fallthrough case is recognized.


janisozaur:
This is the only one I found in CEngine.

bwaldbaum:
Great, thank you for checking.

PR #126, opened 2019-03-06
Modify how default units are selected for conversions

rudyhuyn:
## Fixes #.
Improve how the Calculator selects the default units used for conversions.

- remove GetRegion() and UKSource/USSource/Source
- include all countries using US Customary units
- separate how we compute if the current country uses `US Customary units` and `Fahrenheit` (not part of the definition and used by more countries)
- update the value of `Volume_TeaspoonUS` and `Volume_TablespoonUS` with more precise values.

- Tested in French, US, UK and Liberian


src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -156,27 +156,32 @@ void UnitConverterDataLoader::GetCategories(_In_ shared_ptr<vector<UCM::Category
 
 void UnitConverterDataLoader::GetUnits(_In_ unordered_map<ViewMode, vector<OrderedUnit>>& unitMap)
 {
-    bool USSource, USTarget;
-    bool UKSource, UKTarget;
-    bool Source, Target;
+    // US + Federated States of Micronesia, Marshall Islands, Palau
+    bool isPartOfCOFA = m_currentRegionCode == L"US" ||
```
mcooley: I don't think you need the isPartOfCOFA variable--you can define useUSCustomary directly, and useFarenheit = useUSCustomary or BS, KY, LR

(Avoiding references to international agreements and politicial entities in code can be a good idea, since those things tend to change over time.)

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -156,27 +156,32 @@ void UnitConverterDataLoader::GetCategories(_In_ shared_ptr<vector<UCM::Category
...
+        m_currentRegionCode == L"FM" ||
+        m_currentRegionCode == L"MH" ||
+        m_currentRegionCode == L"PW";
 
-    USSource = (GetRegion() == L"US") ? true : false;
-    USTarget = USSource;
+    // COPA + Liberia
+    // Source: https://en.wikipedia.org/wiki/Metrication
+    bool useUSCustomary = isPartOfCOFA || m_currentRegionCode == L"LR";
+    bool useISU = !useUSCustomary;
```
mcooley: nit: I don't think "ISU" is a commonly-used abbreviation. Consider "useSI" or "useMetric"

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -156,27 +156,32 @@ void UnitConverterDataLoader::GetCategories(_In_ shared_ptr<vector<UCM::Category
...
 
-    UKSource = (GetRegion() == L"UK") ? true : false;
-    UKTarget = UKSource;
+    // COPA + the Bahamas, the Cayman Islands and Liberia
+    // Source: http://en.wikipedia.org/wiki/Fahrenheit
+    bool useFahrenheit = isPartOfCOFA || m_currentRegionCode == "BS" || m_currentRegionCode == "KY" || m_currentRegionCode == "LR";
 
-    Source = (GetRegion() == L"Others") ? true : false;
-    Target = Source;
+    bool useWattInsteadOfKilowatt = m_currentRegionCode == "GB";
```
mcooley: I know the code already works this way, but none of the (American) people our team who have looked at this are sure why this behavior exists. We'll have to dig through the history and see if we can find why it was originally added.

Just in case there are any UK-based spectators out there--does anyone have an idea why watts might be preferred over kilowatts?

mcooley:
Good improvements overall, thank you! I had a fun time reading all about the history of metrication to verify your work. :)

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
rudyhuyn: I really hesitated about it, but I also don't want to assume that all countries using US Customary units also use Farenheit (this list will probably change with the times). I would prefer to rename `isPartOfCOFA` to `useUSCustomaryAndFahrenheit`.

mcooley: useUSCustomaryAndFarenheit sounds good to me.

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
rudyhuyn: Note that I did my best to not use a French acronym!

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
rudyhuyn: Same reaction, we should file an issue, it will get more visibility.

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
jsoref: SI (~ System International) is known in math/science even in the US...

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
jsoref: So, for some reason, this doesn't shock me. I know in the US we* use kW. But for some reason killowatts doesn't seem natural for the UK (even as someone who has only visited).

The best quick hit I can find supporting this is:
https://www.frequencycast.co.uk/howmanywatts.html

Note that it list things like 1800 watts.

*Yes, I live in Canada now, and lived in Europe most recently, but I grew up in the US, so my default units are miles-farenheit-kW-pounds-inches-feet...

mcooley: Internally, I traced this behavior back to 2013, in the Windows 8.1 source code (BLUE bug 115814, change 16507). I can't find many details from that time which explain why it was done this way. I'll open a tracking issue to see if we can justify this or remove it.

mcooley: #167

Issue #141, opened 2019-03-07
Multiple issues in German translation (de-DE/Resources.resw)


ferzkopp:
There are a number of translation issues in the German resource file
https://github.com/Microsoft/calculator/blob/master/src/Calculator/Resources/de-DE/Resources.resw
as follows:

**(1)**

```
  <data name="andButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>And</value>
    <comment>Screen reader prompt for the Calculator And button</comment>
  </data>
  <data name="orButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>Or</value>
    <comment>Screen reader prompt for the Calculator Or button</comment>
  </data>
  <data name="notButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>Not</value>
    <comment>Screen reader prompt for the Calculator Not button</comment>
  </data>
```

- should use UND, ODER, NICHT

similar to:

```
  </data>
  <data name="xorButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>Exklusives ODER</value>
    <comment>Screen reader prompt for the Calculator XOR button</comment>
</data>
```

**(2)**

```
  <data name="rolButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>Nach links drehen</value>
    <comment>Screen reader prompt for the Calculator ROL button</comment>
  </data>
  <data name="rorButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>Nach rechts drehen</value>
    <comment>Screen reader prompt for the Calculator ROR button</comment>
</data>
```

- should use "schieben" or "verschieben" instead of "drehen"
- see also: https://de.wikipedia.org/wiki/Maschinensprache#Kategorisierung_der_Befehle

**(3)**

```
  <data name="equalButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
    <value>Gleich</value>
    <comment>Screen reader prompt for the invert button on the scientific operator keypad</comment>
</data>
```

- incorrect comment text; should not be "invert" but "equal" for equalButton.[]

**(4)**

```
[...]

PR #146, opened 2019-03-07
Fix some wording translation in german (de-DE/Resources.resw) #141

michilehr:
## Fixes #.

Corrects wrong translations for german (de-DE/Resources.resw) like explained in #141

Some corrections in the translations which were no-brainers.

### DIFF
--- a/src/Calculator/Resources/de-DE/Resources.resw
+++ b/src/Calculator/Resources/de-DE/Resources.resw
@@ -938,7 +938,7 @@
     <comment>Screen reader prompt for the x squared on the scientific operator keypad. </comment>
   </data>
   <data name="xpower3Button.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Hoch 3</value>
+    <value>Kubik</value>
     <comment>Screen reader prompt for the x cubed on the scientific operator keypad. </comment>
   </data>
   <data name="invsinButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
--- a/src/Calculator/Resources/de-DE/Resources.resw
+++ b/src/Calculator/Resources/de-DE/Resources.resw
@@ -770,23 +770,23 @@
     <comment>Screen reader prompt for the Calculator number "F" button</comment>
   </data>
   <data name="andButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>And</value>
+    <value>UND</value>
     <comment>Screen reader prompt for the Calculator And button</comment>
   </data>
   <data name="orButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Or</value>
+    <value>ODER</value>
     <comment>Screen reader prompt for the Calculator Or button</comment>
   </data>
   <data name="notButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Not</value>
+    <value>NICHT</value>
     <comment>Screen reader prompt for the Calculator Not button</comment>
   </data>
   <data name="rolButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Nach links drehen</value>
+    <value>Nach links verschieben</value>
     <comment>Screen reader prompt for the Calculator ROL button</comment>
   </data>
   <data name="rorButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Nach rechts drehen</value>
+    <value>Nach rechts verschieben</value>
     <comment>Screen reader prompt for the Calculator ROR button</comment>
   </data>
   <data name="lshButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
@@ -847,7 +847,7 @@
   </data>
   <data name="equalButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
     <value>Gleich</value>
-    <comment>Screen reader prompt for the invert button on the scientific operator keypad</comment>
+    <comment>Screen reader prompt for the equal button on the scientific operator keypad</comment>
   </data>
   <data name="shiftButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
     <value>Inverse Function</value>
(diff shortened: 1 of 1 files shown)


src/Calculator/Resources/de-DE/Resources.resw
```diff
@@ -931,26 +931,26 @@
...
+    <comment>Screen reader prompt for the Calculator tanh button on the scientific operator keypad</comment>
   </data>
   <data name="xpower2Button.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Quadrat</value>
+    <value>Quadrieren</value>
     <comment>Screen reader prompt for the x squared on the scientific operator keypad. </comment>
   </data>
   <data name="xpower3Button.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Würfel</value>
+    <value>Hoch 3</value>
```
jbartlau:
```suggestion
    <value>Kubik</value>
```

src/Calculator/Resources/de-DE/Resources.resw
```diff
@@ -931,26 +931,26 @@
   </data>
   <data name="tanhButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
     <value>Hyperbolischer Tangens</value>
-    <comment>Screen reader prompt for the Calculator tanh button  on the scientific operator keypad</comment>
+    <comment>Screen reader prompt for the Calculator tanh button on the scientific operator keypad</comment>
   </data>
   <data name="xpower2Button.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Quadrat</value>
+    <value>Quadrieren</value>
```
jbartlau:
```suggestion
    <value>Quadrat</value>
```

xyzzer:
If nothing's changed since I last checked - loc fixes should be suggested using the Language Community app or Feedback Hub. If it's still the case - perhaps there should be a readme.md in the resources directory explaining that since it will keep happening otherwise.
image

ferzkopp:
@xyzzer - I don't have a Language Community button in my Feedback Hub. Also, why would one need to go through such a complicated process? The app code and loc resources are on github now and can be reviewed and fixed there.

HowardWolosky:
@ferzkopp -- Thanks for the submission.  We definitely want to get these translations right for all of our customers.  I've reached out to our localization team to find out what the best way to handle these types changes, and will update this PR and our documentation once I hear back.

xyzzer:
@ferzkopp My understanding is that localized resources are currently automatically checked in from a separate localization service. If you change them manually with a PR - they will be overwritten the next time the build pulls new resources, so if you want to update them - you need to push the changes to the loc system first.

michilehr:
If the translations were somewhere else, then they should not be checked into the repository and loaded from the source during the build.

xyzzer:
@michilehr I see your point. Perhaps they shouldn't. It's somewhat valuable for a dev though to be able to see what they are and how they changed over time.

fforjan:
@michilehr 
How would someone then work disconnected from internet, and try to build/reproduce an issue in French locally? You should have a local copy into your git ?

michilehr:
@xyzzer that's also a good point. But it does not make sense to check them in where they got overwritten automatically from time to time. 

@fforjan there should be an install script which fetches external resources when the developer clones the repository the first time and which can be updated.
Create a second repository for the language files for example which gets updated automatically which will be included in here.
The fun fact with your problem is that maybe you will be able to reproduce the problem, but you wouldn't be able to fix it because language files are somewhere else.

HowardWolosky:
Update: I'm still working through the details with our localization team on how we'll be handling resource change submissions coming in through this repo.  I'm hopeful that we'll have clarity on this by tomorrow, and at that time I'll update this PR as well as our repo's documentation.  Thanks for your contribution and continued patience.

Regarding the rest of this discussion: @xyzzer is correct in that the localized files that we have published in this repo are the output of a separate localization process that happens behind the scenes, and that process is not currently structured to accept direct changes to the localized output.

When taking Calculator open, as a team we had a decision to make with whether we'd only share the English strings (and avoid this problem entirely), or share the strings knowing that we wouldn't be able to directly accept changes to them.  We thought the community would be best served with us sharing as much as we could, and thus we have the files as part of our repo.  Now we just need to figure out the smoothest way we can merge feedback to our localized strings moving forward.

@ferzkopp - According to our localization team, the language quality tab in Feedback Hub will only be displayed if the OS display language is different from `English (United States)`.

ferzkopp:
@HowardWolosky - Thanks for the follow up. I've reported an issue via Feedback Hub against Feedback Hub to enable the Language Quality tab all the time. After all, about 20% of the US population is speaking a second language (see: https://en.wikipedia.org/wiki/Languages_of_the_United_States).

src/Calculator/Resources/de-DE/Resources.resw
```diff
@@ -847,7 +847,7 @@
   </data>
   <data name="equalButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
     <value>Gleich</value>
-    <comment>Screen reader prompt for the invert button on the scientific operator keypad</comment>
+    <comment>Screen reader prompt for the equal button on the scientific operator keypad</comment>
```
HowardWolosky: This is a good catch.  Can you make _this_ change to the en-us resource?  The comment will then get fixed across all of the resources the next time it round-trips through our localization process.

src/Calculator/Resources/de-DE/Resources.resw
```diff
@@ -931,26 +931,26 @@
   </data>
   <data name="tanhButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
     <value>Hyperbolischer Tangens</value>
-    <comment>Screen reader prompt for the Calculator tanh button  on the scientific operator keypad</comment>
```
HowardWolosky: You can make this change in the `en-US` resource as well (if you wish)

HowardWolosky:
As mentioned before, we cannot directly accept localization changes directly through this repo.  I've worked with our Localization team to identify a temporary workaround to make sure that this contribution isn't lost.  I have filed an issue in our internal tracking system (`20784911`) which tracks all of your suggested changes, and the Localization team will that internal issue to review and apply your changes via their system.

Your participation is very much appreciated, and your work _will_ get applied.  Unfortunately, I cannot directly merge this change into this repo due to the way our system currently works.  As I mentioned in the PR feedback, if you submit those two other changes against the `en-US` file, I will accept those changes.

Thanks again.

PR #151, opened 2019-03-07
Removed duplicate zeros in Margin and Paddings

MoienTajik:
Removed unuseful duplicated zeros for Margin and Paddings which they're same, like `0,0,0,0` and `0`.

1. As said [here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.margin?view=netframework-4.7.2#xaml-values) in Microsoft Docs:

> For instance, Margin="20" will be interpreted to mean a Thickness with all properties set to 20.

2. Double checked all the views manually.

### DIFF
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -13,7 +13,7 @@
             </ResourceDictionary.MergedDictionaries>
             <ResourceDictionary.ThemeDictionaries>
                 <ResourceDictionary x:Key="Default">
-                    <Thickness x:Key="HighContrastThicknessTop">0,0,0,0</Thickness>
+                    <Thickness x:Key="HighContrastThicknessTop">0</Thickness>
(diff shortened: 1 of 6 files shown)


MoienTajik:
What happened to this ?!

HowardWolosky:
Thanks for the contribution.  This change comes down to a style preference.  We have chosen to defer to [XAMLStyler](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler) as an automated solution for keeping our style consistent across all of our XAML files as mentioned in our [README](https://github.com/Microsoft/calculator#getting-started).

If you disagree with this current style, I would recommend that you [open an issue against XAMLStyler](https://github.com/Xavalon/XamlStyler/issues) so that we can be sure that it applies this style consistently across our XAML files, regardless of it's collapsing `0,0,0,0` to `0` or `2,2,2,2` to `2`.

PR #135, opened 2019-03-07
Spelling

jsoref:
### Description of the changes:
- Fixes spelling

### How changes were validated:

- No validation

PR #160, opened 2019-03-07
Update ManualTests.md

0xflotus:
I only fixed a typo


mcooley:
Thanks @0xflotus, looks like this was already fixed in #135.

PR #176, opened 2019-03-07
Added support for pasting of monetary values

### LINKED ISSUE #52: Does not support paste of monetary values
When pasting input that contains a prefix currency symbol (e.g., $5), the display shows "Invalid input". We should not treat this as invalid input, but instead trim away prefix currency symbols and treat them as numerical values.

willibrandon:
## Fixes #52.
Fixes #52 

Added support for pasting of prefix currency symbols supported by the Windows keyboard.

    yen or yuan (¥)
    unspecified currency sign (¤)
    Ghanaian cedi (₵)
    dollar or peso ($)
    colón (₡)
    won (₩)
    shekel (₪)
    naira (₦)
    Indian rupee (₹)
    pound (£)
    euro (€)

- Manually tested each prefix currency symbol supported by the Windows keyboard and ran unit tests.


src/CalcViewModel/Common/Utils.cpp
```diff
@@ -86,11 +86,11 @@ bool Utils::IsLastCharacterTarget(_In_ wstring const &input, _In_ wchar_t target
...
 }
 
-//return wstring after removing characters like space, comma, and double quotes
+//return wstring after removing characters like space, comma, double quotes, and monetary prefix currency symbols supported by the Windows keyboard
 wstring Utils::RemoveUnwantedCharsFromWstring(wstring input)
 {
-    wchar_t unWantedChars[] = { L' ', L',', L'"', 8234, 8235, 8236, 8237 };
-    return RemoveUnwantedCharsFromWstring(input, unWantedChars, 6);
+    wchar_t unWantedChars[] = { L' ', L',', L'"', 165, 164, 8373, 36, 8353, 8361, 8362, 8358, 8377, 163, 8364, 8234, 8235, 8236, 8237 };
+    return RemoveUnwantedCharsFromWstring(input, unWantedChars, 18);
```
mlaventure: may as well compute the size automatically:
```
sizeof(unWantedChars) / sizeof(*unWantedChars)
```

src/CalcViewModel/Common/Utils.cpp
```diff
@@ -86,11 +86,11 @@ bool Utils::IsLastCharacterTarget(_In_ wstring const &input, _In_ wchar_t target
     return !input.empty() && input.back() == target;
 }
 
-//return wstring after removing characters like space, comma, and double quotes
+//return wstring after removing characters like space, comma, double quotes, and monetary prefix currency symbols supported by the Windows keyboard
 wstring Utils::RemoveUnwantedCharsFromWstring(wstring input)
 {
-    wchar_t unWantedChars[] = { L' ', L',', L'"', 8234, 8235, 8236, 8237 };
-    return RemoveUnwantedCharsFromWstring(input, unWantedChars, 6);
+    wchar_t unWantedChars[] = { L' ', L',', L'"', 165, 164, 8373, 36, 8353, 8361, 8362, 8358, 8377, 163, 8364, 8234, 8235, 8236, 8237 };
```
jlaanstra: Let's also modernize this and use std::array<wchar_t>.

mikaelmello: Why would this be better as opposed to keep the current conventions?

What are the benefits of changing it?

jlaanstra: std::array provides you with the size, so you don't have to worry about calculating the length correctly. static constexpr causes the value to be calculated at compile-time vs runtime.

mikaelmello: I've also read a little and it does feels better. It's been a while since I've used C++ though. Would we still be able to pass `std::array<wchar_t>' as a parameter where the function expects a `wchar_t[]`

jlaanstra: You should still be able to get a pointer to the first element. However if that doesn't work we should just update the function so it does understand std::array. Ideally we would use something like ahn array_view, but that's not part of the standard yet.

danbelcher-MSFT: @mikaelmello, is there a specific example function you're thinking of?  Functions don't typically accept primitive arrays as the parameter type, ie, `wchar_t[]`.  When declared that way, the type of the variable [decays](https://stackoverflow.com/questions/29346552/passing-an-array-to-a-function-formal-parameter) to `wchar_t*`.

Instead of passing the `std::array` to another function, the best approach would be to pass begin and end iterators over a sequence.  Then the function can accept input from other types, such as `vector`

src/CalcViewModel/Common/Utils.cpp
danbelcher-MSFT: Addressed below.  We should prefer using the standard containers and the `size` member function.

src/CalcViewModel/Common/Utils.cpp
danbelcher-MSFT: This file supports unicode characters, meaning you can embed the character you want to ignore, ie `L'€'`.  Please update the values here so the array is readable.

src/CalcViewModel/Common/Utils.cpp
```diff
@@ -86,11 +86,11 @@ bool Utils::IsLastCharacterTarget(_In_ wstring const &input, _In_ wchar_t target
     return !input.empty() && input.back() == target;
 }
 
-//return wstring after removing characters like space, comma, and double quotes
+//return wstring after removing characters like space, comma, double quotes, and monetary prefix currency symbols supported by the Windows keyboard
 wstring Utils::RemoveUnwantedCharsFromWstring(wstring input)
```
danbelcher-MSFT: Can you update the function so the parameter type is `const wstring&`?

src/CalcViewModel/Common/Utils.cpp
mikaelmello: @jlaanstra, I see, thanks for the explanation!

@danbelcher-MSFT, no, you're right about the function expecting a * (and not a []). I just thought that maybe the way that `std::array` was built could be somewhat compatible? Apparently not. I like the approach you said, specially the "Then the function can accept input from other types, such as vector" part. Thank you also for taking your time and explaining!

willibrandon: When I embed the character `L'€'` into the array, during runtime it becomes `â`.

mikaelmello: If I had to guess I'd say that your output does not support unicode or is using a different encoding

danbelcher-MSFT: @willibrandon, thanks for trying my suggestion.  When I try locally, I get a warning.  Let's ignore my suggestion and fall back to using unicode literals, ie `L'\u20AC'` is the Euro symbol.  Then please add a comment for each unicode symbol.  What I'm trying to get out of this is that the array should be readable so that other developers don't need to go look up each symbol.

src/CalcViewModel/Common/Utils.cpp
danbelcher-MSFT: This function is only used by `CopyPasteManager`.  Let's remove it from `Utils` and move it to that file.

src/CalcViewModel/Common/Utils.cpp
willibrandon: @danbelcher-MSFT , may I ask what warning you are receiving? The reason I ask is because I'm having the same problem with writing unit tests to validate the behavior. When I write the unit tests using unicode characters, some of the currency symbols get messed up in the same way. File encoding is set to Unicode (UTF-8 without signature) - Codepage 65001 so I'm not entirely sure why this is happening. One thing I have noticed is if I use the /utf-8 option for the CalculatorUnitTests project [as specified here](https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8) then the unicode characters do not get messed up.

This is not a topic I'm very experienced in so any help would be very appreciated.

willibrandon: @danbelcher-MSFT , it looks like a fix for this is already in the works #295.

danbelcher-MSFT: Yep, the PR for that was merged, so you can merge from master and see if you can get the embedded Euro characters etc working again.

danbelcher-MSFT: Sorry, I confused the nature of that PR.  We still need to add the /utf8 compiler flag, so this won't work yet.

danbelcher-MSFT: Hey @willibrandon, I don't want to block your PR waiting for us to figure out encoding problems.  What I care most about is readability of the code.  I'm fine with what you had for your first iteration (the raw numerical value of the monetary characters), but please add comments so that other readers can know what the values represent.  Then we can get your change checked-in.

willibrandon: @danbelcher-MSFT , I have added comments and made the requested changes. Please let me know if anything else needs to be done.

PR #182, opened 2019-03-07
Fix #178: Wrong Date Calculation on last Saturday to Sunday in March

### LINKED ISSUE #178: Wrong Date Calculation on last Saturday to Sunday in March
**Describe the bug**
The Date Calculator does not factor in the last Saturday to Sunday every March (ex. 2017/03/25 - 2017/03/26).
Because of that the Calculation of the whole month is also wrong.

**Steps To Reproduce**
1. Go to the Date Calculator
2. Select 30. March 2019 and 31. March 2019
3. The Calculator says its the same Date.

**Expected behavior**
If you do the steps above it should show a 1Day difference.
If you select the whole Month of March it says its 29 Days. The right answer would be 30 Days.

**Screenshots**
image
image

 - OS Build: 10.0.17763.0
 - Architecture: X64

patrick-wassmer:
## Fixes #178.

Solve issue [#178](https://github.com/Microsoft/calculator/issues/178)

- Include daylight time support in date difference calculation (when switching from winter to summer)

Manual testing

**Before change :** 

image

image

**After Change:**

image

image

### DIFF
--- a/src/CalcViewModel/Common/DateCalculator.cpp
+++ b/src/CalcViewModel/Common/DateCalculator.cpp
@@ -206,7 +206,8 @@ void DateCalculationEngine::GetDateDifference(_In_ DateTime date1, _In_ DateTime
 int DateCalculationEngine::GetDifferenceInDays(DateTime date1, DateTime date2)
 {
     // A tick is defined as the number of 100 nanoseconds
-    long long ticksDifference = date2.UniversalTime - date1.UniversalTime;
+    // Add one hour to the difference to have correct number of days when change of daylight time (winter to summer) happens between date1 and date2
+    long long ticksDifference = date2.UniversalTime - date1.UniversalTime + c_hour;
     return static_cast<int>(ticksDifference / static_cast<long long>(c_day));
 }
 


patrick-wassmer:
> It works but it's not enough robust, we should use the UTC timezone when we "clip" dates and in `DateCalculator.cpp`. I can send a diff tonight.

They already use date in UniversalTime before calling the fonction in
void DateCalculationEngine::GetDateDifference(_In_ DateTime date1, _In_ DateTime date2, _In_ DateUnit outputFormat, _Out_ DateDifference *difference)

rudyhuyn:
The issue is before that they build the DateTime using local timezone
`datetime.Hour = 12`, etc...  sowitg daylight saving, the difference
between 2 dates can be 23 or 25, creating the bug you saw. If you build the
DateTime directly in UTC, you won't have the issue. I will CC you in my diff

rudyhuyn:
Here is my diff:

https://github.com/Microsoft/calculator/pull/193

jlaanstra:
We don't do time today. #193 is probably the better fix as it doesn't change the behavior.

HowardWolosky:
Thanks for trying to fix #178.  Very much appreciated!  After reviewing both this change and @rudyhuyn's #193, #193 is a more comprehensive fix.

I'm going to close this PR out and we'll stick with Rudy's.  Sorry that this change didn't get in, but looking forward to seeing your next submission.

PR #183, opened 2019-03-07
Compare locale strings, not their pointers

janisozaur:
I counted number of characters in `en-US` manually and came up with magic value
`5`.

### DIFF
--- a/src/CalcViewModel/Common/LocalizationSettings.h
+++ b/src/CalcViewModel/Common/LocalizationSettings.h
@@ -26,8 +26,9 @@ namespace CalculatorApp
                     m_digitSymbols.at(i) = formatter->FormatUInt(i)->Data()[0];
                 }
 
+                wchar_t resolvedName[LOCALE_NAME_MAX_LENGTH];
                 result = ResolveLocaleName(formatter->ResolvedLanguage->Data(),
-                    m_resolvedName,
+                    resolvedName,
                     LOCALE_NAME_MAX_LENGTH);
                 if (result == 0)
                 {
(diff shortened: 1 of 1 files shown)


Mindavi:
Pretty sure the comparison with == is valid in c++ (see http://www.cplusplus.com/reference/string/basic_string/operators/)

janisozaur:
https://github.com/Microsoft/calculator/blob/057401f5f2b4bb1ea143da02c773ac18d1bb9a2e/src/CalcViewModel/Common/LocalizationSettings.h#L380

src/CalcViewModel/Common/LocalizationSettings.h
```diff
@@ -177,7 +177,7 @@ namespace CalculatorApp
 
             Platform::String^ GetEnglishValueFromLocalizedDigits(const std::wstring& localizedString) const
             {
-                if (m_resolvedName == L"en-US")
+                if (wcsncmp(m_resolvedName, L"en-US", 5) == 0)
```
jlaanstra: Let's change the code to not be a wchar_t[] so the comparison will work :).

src/CalcViewModel/Common/LocalizationSettings.h
```diff
@@ -177,7 +177,7 @@ namespace CalculatorApp
 
             Platform::String^ GetEnglishValueFromLocalizedDigits(const std::wstring& localizedString) const
             {
-                if (m_resolvedName == L"en-US")
```
danbelcher-MSFT: Agree with @jlaanstra here.  Please update the type of `m_resolvedName` to be `wstring` and then the comparison will work here.

PR #186, opened 2019-03-07
Fix declarations of functions

janisozaur:
* Add `static` where needed
* Explicitly name inheritance type
* Add missing include
* Drop `#pragma once` from .cpp files


src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -30,7 +30,7 @@ using namespace CalcEngine;
 //
 //  returns a virtual number for precedence for the operator. We expect binary operator only, otherwise the lowest number
 // 0 is returned. Higher the number, higher the precedence of the operator.
-INT NPrecedenceOfOp(int nopCode)
+static INT NPrecedenceOfOp(int nopCode)
```
danbelcher-MSFT: I think the better fix here is to move these functions to an anonymous namespace.  Same for the other instances below.

janisozaur:
Updated

src/CalcManager/CalcException.h
```diff
@@ -15,12 +15,3 @@ class CalcException : std::exception
 private:
     HRESULT m_hr;
 };
-
-void IFT(HRESULT hr)
```
janisozaur: This was moved into `History.cpp`, the only consumer of `CalcException` is also in there, should I move it into `History.cpp` as well and remove this file?

danbelcher-MSFT: Yep, I think that's a good change.  Let's limit visibility of functions to places where they are needed.

janisozaur: Updated

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -30,24 +30,26 @@ using namespace CalcEngine;
 //
 //  returns a virtual number for precedence for the operator. We expect binary operator only, otherwise the lowest number
 // 0 is returned. Higher the number, higher the precedence of the operator.
```
danbelcher-MSFT: Move this comment to be with the `NPrecedenceOfOp` function.

src/CalcManager/CEngine/History.cpp
```diff
@@ -2,18 +2,41 @@
...
 #include "CalculatorVector.h"
 #include "ExpressionCommand.h"
-#include "CalcException.h"
 
 constexpr int ASCII_0 = 48;
 
 using namespace std;
 using namespace CalcEngine;
 
+class CalcException : public std::exception
```
danbelcher-MSFT: On second look, this class isn't needed.  It doesn't override `std::exception::what` and nowhere in the codebase explicitly catches a `CalcException`.

src/CalcManager/CEngine/scicomm.cpp
janisozaur: done

src/CalcManager/CEngine/History.cpp
janisozaur: done

danbelcher-MSFT:
Hey @janisozaur, looks like there's merge conflicts now.  If you fix them up we can get the change merged.  Thanks!

janisozaur:
I'll try updating this in the evening, some 10h from now.

janisozaur:
Updated

PR #188, opened 2019-03-07
Add `override` keyword where applicable

janisozaur:


danbelcher-MSFT:
I have a similar comment here as with #189.  These improvements are great, but we can also add them to other areas of the codebase outside of just the engine.

PR #189, opened 2019-03-07
Mark classes as `final` where applicable

janisozaur:
I checked and there should be no conflict with https://github.com/Microsoft/calculator/pull/188


danbelcher-MSFT:
I'm fine with this change, but I noticed that this improvement is limited only to pure C++ classes.  For C++/CX, there is the equivalent [`sealed`](https://docs.microsoft.com/en-us/cpp/windows/sealed-cpp-component-extensions?view=vs-2017) keyword and, if we think classes should be final/sealed where possible, then we should also seal the CX classes.

PR #190, opened 2019-03-07
Use consistent path separators in include paths

janisozaur:


saviour07:
That isn't an escape sequence in the #include's, that is a legal directory separator.
Makes no difference having the existing "\" or the change you've made to "/".

marekr:
This seems like a subtle linux troll attempt

danbelcher-MSFT:
Approved because this establishes consistency across the codebase and sets a precedence for style moving forward.

PR #192, opened 2019-03-08
Fixes Issue #162 Pasting long decimal number results in Invalid input

### LINKED ISSUE #162: Pasting long decimal number results in Invalid input
**Describe the bug**
When pasting a decimal number with more than 16 decimal digits, the calculator outputs `Invalid input`.

**Steps To Reproduce**
Paste `0.00012345678987654` into the calculator.

**Expected behavior**
The input to be accepted. If it is too long, trim it, like the Windows 7 calculator does.

**Screenshots**

 - OS Build: 18334.1.amd64fre.19h1_release.190205-1505
 - Architecture: x64

naveenspace7:
## Fixes #162.
Change has been made to retrieve a substring of the given string by considering its maximum length.

### Summary of the issue 
When pasting a long decimal number, the calculator display shows "Invalid Input".

Added the following in a function to trim.
`operand = operand.substr(0, operand.length() - exceededBy);`

- Tested with long operands, whose values will be trimmed when the length exceeds the maximum value. 
- Also tested by giving long invalid numbers (like 0.23459288484823nj) to make sure they are still treated as "Invalid Inputs".

Note: Further modifications to the UnitTests will be made once this change is agreed.

### DIFF
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -307,11 +307,12 @@ bool CopyPasteManager::ExpressionRegExMatch(vector<wstring> operands, ViewMode m
             // or which will break conversion from string-to-ULL.
             wstring operandValue = SanitizeOperand(operand);
 
-            // If an operand exceeds the maximum length allowed, break and return.
-            if (OperandLength(operandValue, mode, modeType, programmerNumberBase) > maxOperandLength)
+            // If an operand exceeds the maximum length allowed, trim it to its allowed length.
+            size_t OperandLen = OperandLength(operandValue, mode, modeType, programmerNumberBase);
+            if (OperandLen > maxOperandLength)
             {
-                expMatched = false;
-                break;
+                // Try to trim the length to allowed length
+                TrimOperand(operandValue, OperandLen - maxOperandLength);
             }
 
             // If maxOperandValue is set and the operandValue exceeds it, break and return.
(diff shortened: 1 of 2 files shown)


rudyhuyn:
If the calculator isn't able to handle a number, I would expect the calculator to display an error message (but a better one).

Can you think of a use case where working with a truncated number makes sense?

The only one I can think about is a number a lot of decimals, but in this case, you should also check the number of decimals

jlaanstra:
I don't think changing user input is the right thing to do. If you want greater precision you can use Scientific mode. Error message could be improved though.

naveenspace7:
Oh, I see your point, which made me realize that this fix was rather narrow-sighted. 

I can still go ahead with your suggestions and display something like "Invalid Length". What are your comments?

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -339,6 +340,11 @@ bool CopyPasteManager::ExpressionRegExMatch(vector<wstring> operands, ViewMode m
     return expMatched;
 }
 
+void CopyPasteManager::TrimOperand(std::wstring& operand, size_t exceededBy)
```
HowardWolosky: In general, we want to avoid having methods directly modifying input values.  A safer, more testable, and more standard design is to leave the input alone and return the modified version as the result.  For an example, see [`wstring CopyPasteManager::SanitizeOperand(const wstring& operand)`](https://github.com/Microsoft/calculator/blob/500d84ab5ee0eef5d198012a73f13486662be7f4/src/CalcViewModel/Common/CopyPasteManager.cpp#L410).

HowardWolosky:
Thanks for the fix!  Unfortunately, I'm going to close out this implementation for now for a couple reasons:
 1. This PR doesn't increase the max length of supported input in `Standard` mode (which the bug indicates it should).
 2. More importantly, after reviewing the feedback in this PR and elsewhere on the idea of just blindly truncating user input, @grochocki is going to be updating #162 with a proposal on how we can truncate with an asynchronous notification to the user so that they _know_ that their input has been truncated (allowing them to decide if they want to proceed with the truncated value or not).

After #162 has been updated with the new design, I'd welcome an updated PR from you.

Again, thanks for your interested in the repo!

src/CalcViewModel/Common/CopyPasteManager.cpp
naveenspace7: Will keep that in mind. Thanks.

PR #193, opened 2019-03-08
Fix issue with Date diff when it includes a Daylight Saving Time

### LINKED ISSUE #178: Wrong Date Calculation on last Saturday to Sunday in March

rudyhuyn:
## Fixes #178 

The application uses local time to calculate the number of days between 2 dates. If a Daylight Saving Time takes place during this period of time (only Clocks Forward 2am->3am), the application will miss 1 day and fail to calculate the number of days/weeks/months between the 2 dates.
image

`DateCalculationEngine` uses local time to modify dates, however, `AddDays`, `AddWeeks`,... won't add 24 hours or 7 days if DST happens between the 2 dates, but instead add ~23.9/24.1 hours or ~6.9/7.1 days (depends if it's the DST clock backward or clock forward). When the DST is clock forward, `DateCalculationEngine` will miss one day.

### Solution
- use UTC dates to calculate date difference.
Extra Fix:
- use `calendar->FirstPeriodInThisDay` and `calendar->FirstHourInThisPeriod` in ClipTime (else it will set the time to 12PM (noon) in some regions.
- replace `OBSERVABLE_PROPERTY_RW` by `OBSERVABLE_PROPERTY_R` when possible.
- remove  the definition of `CheckClipTimeSameDay` (implementation missing)

Tested manually with different regions (FR, US, ES, JP).


rudyhuyn:
It probably also explains this comment in the code:

>   In the ClipTime function, we used to change timezone to UTC before clipping the time.
     The comment from the previous delopers said this was done to eliminate the effects of
     Daylight Savings Time. We can't think of a good reason why this change in timezone is
     necessary and did find bugs related to the change, therefore, we have removed the
     change. Just in case, we will see if the clipped time is ever a different day from the
     original day, which would hopefully indicate the change in timezone was actually
     necessary. We will collect telemetry if we find this case. If we don't see any
     telemetry events after the application has been used for some time, we will feel safe
     and can remove this function.

jlaanstra:
This change makes sense to me. We don't deal with time and therefore should do these calculations in UTC.

michael-hawker:
Should some Unit Tests be added as a safety for this logic?

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -378,13 +372,14 @@ String^ DateCalculatorViewModel::GetLocalizedNumberString(int value) const
     return ref new String(numberStr.c_str());
 }
 
-// Adjusts the given DateTime to 12AM of the same day
+// Adjusts the given DateTime to 12AM (UTC) of the same day
```
jsoref: Personally I'd suggest avoiding adding trailing whitespace here

src/Calculator/Views/DateCalculator.xaml.cpp
```diff
@@ -36,8 +36,6 @@ using namespace Windows::UI::Xaml::Input;
 using namespace Windows::UI::Xaml::Media;
 using namespace Windows::UI::Xaml::Navigation;
 
-// The User Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
```
jsoref: This change _feels_ unrelated

src/CalculatorUnitTests/DateCalculatorUnitTests.cpp
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jsoref: This change _feels_ unrelated (and I think I've seen it i a number of PRs)

src/CalculatorUnitTests/DateCalculatorUnitTests.cpp
```diff
@@ -387,6 +387,65 @@ namespace DateCalculationUnitTests
...
+          viewModel->FromDate = DateUtils::SystemTimeToDateTime(datetimeDifftest[5].startDate);
+          /*31.03.2008*/
+          viewModel->ToDate = DateUtils::SystemTimeToDateTime(datetimeDifftest[5].endDate);
+
+          //// Assert for the result
+          VERIFY_IS_FALSE(viewModel->IsDiffInDays);
+          VERIFY_ARE_EQUAL(StringReference(L"31 days"), viewModel->StrDateDiffResultInDays);
+          VERIFY_ARE_EQUAL(StringReference(L"1 month, 2 days"), viewModel->StrDateDiffResult);
+
+          // Daylight saving time - Clock Forward
```
jsoref: `Daylight Saving Time` is a proper noun, please capitalize all 3 items :-)

src/CalculatorUnitTests/DateCalculatorUnitTests.cpp
```diff
@@ -387,6 +387,65 @@ namespace DateCalculationUnitTests
...
+
+          viewModel->IsDateDiffMode = true;
+          VERIFY_IS_TRUE(viewModel->IsDateDiffMode);
+
+          /*29.02.2008*/
+          viewModel->FromDate = DateUtils::SystemTimeToDateTime(datetimeDifftest[5].startDate);
+          /*31.03.2008*/
+          viewModel->ToDate = DateUtils::SystemTimeToDateTime(datetimeDifftest[5].endDate);
+
+          //// Assert for the result
```
jsoref: I'm not sure about the style in calc, but I'd half expect this comment to be unnecessary (it should be obvious that VERIFY_... is asserting).

src/CalculatorUnitTests/DateCalculatorUnitTests.cpp
```diff
@@ -387,6 +387,65 @@ namespace DateCalculationUnitTests
...
         }
 
+        TEST_METHOD(DateCalcViewModelDateDiffDaylightSavingTimeTest)
+        {
+          auto viewModel = ref new DateCalculatorViewModel();
+
+          viewModel->IsDateDiffMode = true;
+          VERIFY_IS_TRUE(viewModel->IsDateDiffMode);
+
+          /*29.02.2008*/
```
jsoref: I'm not sure what date format calc normally uses internally, but i'd encourage YYYY-MM-DD as opposed to DD-MM-YYYY. The standard US format is MM-DD-YYYY and thus using DD-MM-YYYY is fairly unexpected and at times ambiguous. Since no one writes YYYY-DD-MM, using YYYY-MM-DD is unambiguous and fairly easy for people to understand.

src/Calculator/Views/DateCalculator.xaml.cpp
```diff
@@ -36,6 +36,7 @@ using namespace Windows::UI::Xaml::Input;
 using namespace Windows::UI::Xaml::Media;
 using namespace Windows::UI::Xaml::Navigation;
 
+// The User Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
```
jlaanstra: Remove.

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -43,11 +43,7 @@ DateCalculatorViewModel::DateCalculatorViewModel() :
     m_StrDateDiffResultAutomationName(L""),
     m_StrDateDiffResultInDays(L""),
     m_StrDateResult(L""),
-    m_StrDateResultAutomationName(L""),
-    m_fromDate({ 0 }),
```
HowardWolosky: What's your argument for removing the explicit initialization here?

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -378,13 +372,14 @@ String^ DateCalculatorViewModel::GetLocalizedNumberString(int value) const
...
-// Adjusts the given DateTime to 12AM of the same day
+// Adjusts the given DateTime to 12AM (UTC) of the same day 
 DateTime DateCalculatorViewModel::ClipTime(DateTime dateTime)
 {
     auto calendar = ref new Calendar();
+    calendar->ChangeTimeZone("UTC");
     calendar->SetDateTime(dateTime);
-    calendar->Period = 1;
-    calendar->Hour = 12;
+    calendar->Period = calendar->FirstPeriodInThisDay;
```
HowardWolosky: Nice. Didn't know these properties existed. 👍

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -56,19 +52,17 @@ DateCalculatorViewModel::DateCalculatorViewModel() :
 
     // Initialize Date Calc engine
     m_dateCalcEngine = make_shared<DateCalculationEngine>(localizationSettings.GetCalendarIdentifier());
-
     // Initialize dates of DatePicker controls to today's date
     auto calendar = ref new Calendar();
+    calendar->ChangeTimeZone("UTC");
```
HowardWolosky: The comment at line 83 (that you referenced in another comment) should be removed now.  Leaving it as is will be confusing, because you now _are_ changing he timezone to `UTC` before calling `ClipTime`.  Can you just add a simpler comment here that it's being done to avoid issues around Daylight Savings Time?

src/CalcViewModel/DateCalculatorViewModel.h
```diff
@@ -23,7 +23,7 @@ namespace CalculatorApp
             // Input Properties
             OBSERVABLE_PROPERTY_RW(bool, IsDateDiffMode);
             OBSERVABLE_PROPERTY_RW(bool, IsAddMode);
-            OBSERVABLE_PROPERTY_RW(bool, IsDiffInDays);     // If diff is only in days or the dates are the same,
+            OBSERVABLE_PROPERTY_R(bool, IsDiffInDays);     // If diff is only in days or the dates are the same,
```
HowardWolosky: nit: Annoying, but can you re-line up the split-line comment here with the one below it?

src/Calculator/Views/DateCalculator.xaml.cpp
```diff
@@ -37,7 +37,6 @@ using namespace Windows::UI::Xaml::Media;
 using namespace Windows::UI::Xaml::Navigation;
 
 // The User Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
```
HowardWolosky: Seems like this file can be reverted.

src/CalcViewModel/Common/DateCalculator.cpp
```diff
@@ -12,6 +12,7 @@ using namespace CalculatorApp::Common::DateCalculation;
 DateCalculationEngine::DateCalculationEngine(_In_ String^ calendarIdentifier)
 {
     m_calendar = ref new Calendar();
+    m_calendar->ChangeTimeZone("UTC");
```
HowardWolosky: Worthwhile to add a comment into here as to why.  This has clearly been a question that has come up in the code in the past.

src/CalcViewModel/DateCalculatorViewModel.cpp
rudyhuyn: These 4 objects are initialized in the constructor lines 61-65, these 4 lines weren't necessary

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -54,6 +54,8 @@ DateCalculatorViewModel::DateCalculatorViewModel() :
     m_dateCalcEngine = make_shared<DateCalculationEngine>(localizationSettings.GetCalendarIdentifier());
     // Initialize dates of DatePicker controls to today's date
     auto calendar = ref new Calendar();
+    // We force the timezone to UTC, to be sure to not being affected by Daylight Saving Time
```
HowardWolosky: `to be sure to not being` -> `to avoid being`

src/CalculatorUnitTests/DateCalculatorUnitTests.cpp
```diff
@@ -387,6 +387,63 @@ namespace DateCalculationUnitTests
...
         }
 
+        TEST_METHOD(DateCalcViewModelDateDiffDaylightSavingTimeTest)
+        {
+          auto viewModel = ref new DateCalculatorViewModel();
+
+          viewModel->IsDateDiffMode = true;
+          VERIFY_IS_TRUE(viewModel->IsDateDiffMode);
+
+          // 29.02.2008
```
HowardWolosky: It looks like you forgot to make these two dates  be M-D-Y like the others that you had changed in a previous update.

rudyhuyn: all the dates in this file are dd.mm.yyyy (see line 275). The few dates I added use the same format

HowardWolosky: Right you are.  I misread the code.  Apologies.

rogersachan:
@HowardWolosky I think this bug has resurfaced according to this reddit post <https://www.reddit.com/r/Windows10/comments/c14fu5/wait_what/>

image

Poster below claims it can be reproduced on the 1905 version.

rogersachan:
image
`Calculator 10.1904.42.0
© 2019 Microsoft. All rights reserved.`

HowardWolosky:
Thanks for the heads-up, @rogersachan.  The two issues aren't actually related, but the reddit thread is completely right about there being a bug.  @Bigpet opened #552 which is tracking the issue you're calling attention to (and also managed to supply a proposed fix with #553 as well!), so we'll keep _this_ issue closed for now and continue using #552 to track this new bug.  Thanks for the report though!

PR #194, opened 2019-03-08
Updating comments per the C++ core guidelines and removing trailing whitespace

WilliamWelsh:
## Fixes
Fixed comments that were inconsistent with the style guidelines described in [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) and the [modern C++/WinRT language projections](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/) and removed trailing whitespace.

- Inserted a space after the beginning of the comment so the text wasn't touching the `//` on all occurrences.
- Removed all occurrences of trailing whitespace
-

- Manual
-
-


WilliamWelsh:
@janisozaur Sure. I've removed the trailing whitespace.

@HowardWolosky Sure. I've now fixed all occurrences of the issue. You are absolutely right, I should have done that in the first place. Also, I am unable to find a specific part in the guidelines that say a comment should be ``// Example`` instead of ``//Example`` explicitly, but throughout the guidelines there are no examples where comments are without a space. Although, the documents also preach about consistency, so I believe the style of the comment should stay consistent.

src/CalcManager/Ratpack/basex.cpp
```diff
@@ -126,28 +126,28 @@ void _mulnumx( PNUMBER *pa, PNUMBER b )
...
+
                 // update carries from
                 mcy >>= BASEXPWR;
                 cy >>= BASEXPWR;
             }
-            
+
             ptrb++;
             ptrc++;
-
```
HowardWolosky: It looks like you're introducing some trailing whitespace.  I may have missed others... you may want to re-review,

WilliamWelsh: There's no whitespace here, just an empty line. I can't find any introduced whitespace.

HowardWolosky: Weird.  I've confirmed you're right by looking at the raw file in your branch.  Not sure why my diffing tool was showing the spaces on this line (and on a couple, otherwise blank, lines as well).

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
```diff
@@ -72,24 +72,30 @@ namespace CalculatorApp
...
             if (dynamic_cast<ToggleButton^>(button))
             {
                 return;
             }
 
-            // The button will go into the visual Pressed state with this call 
+            // The button will go into the visual Pressed state with this call
             VisualStateManager::GoToState(button, "Pressed", true);
 
+            // This timer will fire after c_lightUpTime and make the button
```
HowardWolosky: It's confusing to see this being added here.  Is this just a result of an out-of-date branch?

HowardWolosky:
@WilliamWelsh -- Can you update this change with the latest from `master`?  There are some conflicts right now, and it looks like some odd stuff added (like in KeyboardShortcutManager.cpp).  Will complete my review once that's been updated.  Thanks!

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
WilliamWelsh: I believe, so. I fixed it.

WilliamWelsh:
@HowardWolosky Sure. I've updated it and fixed the conflicts.

PR #195, opened 2019-03-08
DateDiff: Optimize how we build the string used when we calculate the difference between 2 dates

rudyhuyn:
Optimize how we build the result of `GetDateDiffString` and `GetDateDiffStringInDays`, using `std::wstring` (mutuable) instead of `Platform::String` (immutable)

### Removed from the initial diff
- minus sign


jlaanstra:
-2 week, 1 day should be 15 days instead of 14 right?

rudyhuyn:
> -2 week, 1 day should be 15 days instead of 14 right? Looks like that's an existing bug. Not sure about the minus here. I'll leave that for the feature team to decide on.

Yes, I already sent a diff to fix the Daylight saving issue here: https://github.com/Microsoft/calculator/pull/193

dopusteam:
Btw, how difference between dates can be negative?
The minus sign in this case just shows, that your 'from' date more than 'to' date, but difference between dates is still 14 days, no minus 14 days

rudyhuyn:
Yes it was more an indicator than 'To' was before 'From', but I agree, it can be confusing. I will remove the sign part and PR the string optimization only

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -320,43 +332,56 @@ String^ DateCalculatorViewModel::GetDateDiffString() const
...
     }
 
-    return result;
+    return ref new Platform::String(result.data());
 }
 
 String^ DateCalculatorViewModel::GetDateDiffStringInDays() const
 {
-    String^ strDateUnit;
+    wstring result;
```
danbelcher-MSFT: Why not use a `wstringstream`?

rudyhuyn: For this kind of operation (limited number of small wchar_t[] to concatenate), std::(w)string::append is a lot faster than std::(w)stringstream.

See here: https://stackoverflow.com/questions/18892281/most-optimized-way-of-concatenation-in-strings

danbelcher-MSFT: Yes, I should read up on these things before commenting :)  I have been using `stringstream` because I thought it was standard.  TIL!

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -320,43 +323,45 @@ String^ DateCalculatorViewModel::GetDateDiffString() const
...
         }
         else
         {
-            result = String::Concat(result, resourceLoader.GetResourceString(L"Date_Day"));
+            result += resourceLoader.GetResourceString(L"Date_Day")->Data();
         }
     }
 
-    return result;
+    return ref new Platform::String(result.data());
```
danbelcher-MSFT: Don't need `Platform::` here.

src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -320,43 +323,45 @@ String^ DateCalculatorViewModel::GetDateDiffString() const
...
+        result += AppResourceProvider::GetInstance().GetResourceString(L"Date_Days")->Data();
     }
     else
     {
-        strDateUnit = AppResourceProvider::GetInstance().GetResourceString(L"Date_Day");
+        result += AppResourceProvider::GetInstance().GetResourceString(L"Date_Day")->Data();
     }
 
-    return String::Concat(GetLocalizedNumberString(m_dateDiffResultInDays.day), String::Concat(L" ", strDateUnit));
+    return ref new Platform::String(result.data());
```
danbelcher-MSFT: Same here

src/CalculatorUnitTests/DateCalculatorUnitTests.cpp
```diff
@@ -494,6 +494,31 @@ namespace DateCalculationUnitTests
...
+            //// check negative difference
+            VERIFY_IS_FALSE(viewModel->IsDiffInDays);
+            VERIFY_ARE_EQUAL(StringReference(L"305 days"), viewModel->StrDateDiffResultInDays);
+            VERIFY_ARE_EQUAL(StringReference(L"10 months"), viewModel->StrDateDiffResult);
+
+            //// check positive difference
+            viewModel->FromDate = DateUtils::SystemTimeToDateTime(date[6]);
+            viewModel->ToDate = DateUtils::SystemTimeToDateTime(date[10]);
+            VERIFY_IS_FALSE(viewModel->IsDiffInDays);
+            VERIFY_ARE_EQUAL(StringReference(L"- 305 days"), viewModel->StrDateDiffResultInDays);
```
danbelcher-MSFT: Are these tests still valid?

rudyhuyn: You are right, I forgot to remove them!

rudyhuyn: It can still be interesting to test when FromDate < ToDate, I will keep them but modify expected results

danbelcher-MSFT:
@rudyhuyn, there's some conflicts to fix up, probably from one of your earlier changes :)

danbelcher-MSFT:
> @rudyhuyn, there's some conflicts to fix up, probably from one of your earlier changes :)

Nevermind, apparently you can resolve conflicts in GitHub!

PR #197, opened 2019-03-08
Build with /W4

### LINKED ISSUE #142: Not compiling at /W4 - multiple issues
**Describe the bug**

calculator compiles at warning level 3. I can understand that nobody other than me wants to compile at warning level all, but compiling at warning level 4 is a good idea.

This produces a bunch of warnings, most of which involve sketchy code that probably isn't buggy (seemingly hundreds of unreferenced parameters/local variables), but a few of which are highly questionable. One of which is an operator== comparison between a wchar_t[] and a literal string, which is an address comparison instead of a string comparison (`warning C4130: '==': logical operation on address of string constant` in `calculator\src\calcviewmodel\common\localizationsettings.h(180)`).

Of the warnings that I don't think are bugs, but that somebody should still look at:

- Unreferenced parameters/variables
- Variable shadowing
- Lots of implicit `int` to `wchar_t` conversions around the various cstdlib string handling functions
- `CalcException` derives inaccessibly from `std::exception`, which means catching [won't work](https://stackoverflow.com/a/2569819/625687) like it presumably should
- A bunch of variables that are initialized in all reasonably possible branches, but lack a `default` `switch` case, a `NOTREACHED()` kind of macro, or some other way to shut the compiler up.

**Screenshots**
compare_wchar_by_addr

**Additional context**

bwaldbaum:
Fixes #142 - Not compiling at /W4

- All projects are built with warning level 4 (/W4) and treat warnings as errors (/WX)
- Fixed build errors resulting from enabling these compiler flags

- Manual smoke tests of the app
- CI build


src/CalcViewModel/Common/KeyboardShortcutManager.cpp
```diff
@@ -83,11 +82,13 @@ namespace CalculatorApp
...
             VisualStateManager::GoToState(button, "Pressed", true);
 
-            // This timer will fire after c_lightUpTime and make the button
+            // This timer will fire after lightUpTime and make the button
             // go back to the normal state. 
             // This timer will only fire once after which it will be destroyed
             auto timer = ref new DispatcherTimer();
-            timer->Interval = c_lightUpTime;
+            TimeSpan lightUpTime{};
+            lightUpTime.Duration = 500000L; // Half second (in 100-ns units)
```
HowardWolosky: This change is surprising -- what was the warning that necessitated removing the constant for the magic number?

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
```diff
@@ -579,9 +580,9 @@ void KeyboardShortcutManager::OnKeyDownHandler(CoreWindow^ sender, KeyEventArgs^
             // Writer lock for the static maps
             reader_writer_lock::scoped_lock lock(s_keyboardShortcutMapLock);
 
-            auto currentControlKeyPressed = s_ControlKeyPressed.find(viewId);
```
HowardWolosky: What was the warning that desired this variable name change?

src/CalcManager/CEngine/History.cpp
```diff
@@ -227,7 +227,7 @@ void CHistoryCollector::AddUnaryOpToHistory(int nOpCode, bool fInv, ANGLE_TYPE a
             {
                 angleOpCode = CalculationManager::Command::CommandRAD;
             }
-            if (angletype == ANGLE_GRAD)
+            else // (angletype == ANGLE_GRAD)
```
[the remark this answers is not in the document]
bwaldbaum: you're right; I've fixed in the next iteration.

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
bwaldbaum: It was complaining about changing the long to a value-type TimeSpan.

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
bwaldbaum: The variable name was already declared earlier in the function, and declaring the same name in the nested block hides the variable from earlier in the function.

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -1016,7 +1013,7 @@ void StandardCalculatorViewModel::OnMemoryButtonPressed()
     m_standardCalculatorManager.MemorizeNumber();
 
     int windowId = Utils::GetWindowId();
-    TraceLogger::GetInstance().InsertIntoMemoryMap(windowId, IsStandard, IsScientific, IsProgrammer);
+    TraceLogger::GetInstance().InsertIntoMemoryMap(windowId, IsScientific, IsProgrammer);
```
danbelcher-MSFT: Instead of removing `IsStandard`, we should update `InsertIntoMemoryMap` to use the bool.

src/Calculator/Views/DateCalculator.xaml.cpp
```diff
@@ -102,7 +102,7 @@ void DateCalculator::FromDate_DateChanged(_In_ CalendarDatePicker^ sender, _In_
     {
         auto dateCalcViewModel = safe_cast<DateCalculatorViewModel^>(this->DataContext);
         dateCalcViewModel->FromDate = e->NewDate->Value;
-        TraceLogger::GetInstance().LogDateDifferenceModeUsed(ApplicationView::GetApplicationViewIdForWindow(CoreWindow::GetForCurrentThread()));
+        TraceLogger::GetInstance().LogDateDifferenceModeUsed();
```
danbelcher-MSFT: We should update the `LogDateDifferenceModeUsed` to use the `windowId` value.  For an example, see `TraceLogger::LogHypButtonUsed`.

src/CalculatorUnitTests/HistoryTests.cpp
```diff
@@ -42,15 +42,15 @@ namespace CalculatorFunctionalTests
         HistoryViewModel^ m_historyViewModel;
         StandardCalculatorViewModel^ m_standardViewModel;
 
-        void Initialize(unsigned int windowId = 0)
+        void Initialize(unsigned int /*windowId*/ = 0)
```
danbelcher-MSFT: If this parameter isn't used, let's just remove it entirely.

src/CalculatorUnitTests/HistoryTests.cpp
```diff
@@ -42,15 +42,15 @@ namespace CalculatorFunctionalTests
...
         {
             m_standardViewModel = ref new StandardCalculatorViewModel();
             m_standardViewModel->IsStandard = true;
             m_historyViewModel = ref new HistoryViewModel(&(m_standardViewModel->m_standardCalculatorManager));
             m_historyViewModel->SetCalculatorDisplay(m_standardViewModel->m_calculatorDisplay);
         }
 
-        void Cleanup(unsigned int windowId = 0)
+        void Cleanup(unsigned int /*windowId*/ = 0)
```
danbelcher-MSFT: Remove the parameter

src/CalcManager/CalcManager.vcxproj.filters
```diff
@@ -136,9 +136,6 @@
     <ClInclude Include="Header Files\History.h">
       <Filter>Header Files</Filter>
     </ClInclude>
-    <ClInclude Include="Header Files\scimath.h">
```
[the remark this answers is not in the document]
bwaldbaum: I didn't explicitly do it; maybe VS automatically did since that file doesn't exist?

src/Calculator/Views/DateCalculator.xaml.cpp
bwaldbaum: updated.

src/CalcViewModel/StandardCalculatorViewModel.cpp
bwaldbaum: That param is not needed; looking in the function you'll see that it defaults to Standard if it's not Scientific or Programmer mode.

src/CalculatorUnitTests/HistoryTests.cpp
bwaldbaum: I wasn't sure if it was needed as part of the test framework.  It looks like it's not needed, so I'll remove it.

src/CalculatorUnitTests/HistoryTests.cpp
bwaldbaum: done.

src/CalcManager/CalcManager.vcxproj.filters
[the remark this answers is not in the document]
danbelcher-MSFT: I think this was missed when @joshkoon switched over to `RationalMath.h`.  Josh, is that right?  Sometimes I feel like VS doesn't update the filters, even if I save and build.  Then the update ends up in a later commit.

src/CalcViewModel/StandardCalculatorViewModel.cpp
danbelcher-MSFT: I understand.  I'm asking that we update the TraceLogger function to use the bool as part of an `else if`.  Then if all three are false, the function will no-op.  The current change would default to assuming Standard, which isn't necessarily correct.

bwaldbaum: got it.  Updated.

src/CalculatorUnitTests/CurrencyConverterUnitTests.cpp
```diff
@@ -407,11 +407,11 @@ namespace CalculatorUnitTests
             const UCM::Unit usdUnit = GetUnit(unitList, L"USD");
             const UCM::Unit eurUnit = GetUnit(unitList, L"EUR");
 
-            VERIFY_ARE_EQUAL(StringReference(L"United States - Dollar"), ref new String(usdUnit.name.c_str()));
-            VERIFY_ARE_EQUAL(StringReference(L"USD"), ref new String(usdUnit.abbreviation.c_str()));
+            VERIFY_ARE_EQUAL(wstring(L"United States - Dollar"), usdUnit.name);
```
danbelcher-MSFT: Thanks for fixing these up!  For future reference, you can use [operator""s](https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s) to cleanly write a string literal that is converted to a `wstring`.  For example, this is equivalent `L"United States - Dollar"s`.

src/CalculatorUnitTests/HistoryTests.cpp
```diff
@@ -104,14 +104,14 @@ namespace CalculatorFunctionalTests
...
-            for (int i = 1; i < m_standardViewModel->m_standardCalculatorManager.MaxHistorySize(); i++)
+            for (size_t i = 1; i < m_standardViewModel->m_standardCalculatorManager.MaxHistorySize(); i++)
             {
                 m_standardViewModel->m_standardCalculatorManager.SendCommand(Command::Command1);
                 m_standardViewModel->m_standardCalculatorManager.SendCommand(Command::CommandADD);
                 m_standardViewModel->m_standardCalculatorManager.SendCommand(Command::Command2);
                 m_standardViewModel->m_standardCalculatorManager.SendCommand(Command::CommandEQU);
             }
-            VERIFY_ARE_EQUAL(m_historyViewModel->ItemSize, m_standardViewModel->m_standardCalculatorManager.MaxHistorySize());
+            VERIFY_ARE_EQUAL((size_t)m_historyViewModel->ItemSize, m_standardViewModel->m_standardCalculatorManager.MaxHistorySize());
```
danbelcher-MSFT: Prefer `static_cast`

src/CalculatorUnitTests/CurrencyConverterUnitTests.cpp
bwaldbaum: The VERIFY_ARE_EQUAL macro doesn't specify a type for the param, so wstring needs to be specified in order for the == operator to succeed.

src/CalculatorUnitTests/HistoryTests.cpp
bwaldbaum: agreed.

PR #203, opened 2019-03-08
The application displays faulty results with unary operations and numbers when a binary action was previously used

rudyhuyn:
2 major issues due to `CCalcEngine::ProcessCommandWorker` not managing its states correctly.

### Unary actions display faulty results when a binary operation was previously used
a2019-03-08_4-36-01

- Type a random binary operation: `3 + 5` (for example)
- press Enter
- Use an unary action (Power or square root for example)
- press Enter

or

- Type a random binary operation: `3 x 5` (for example)
- press Enter
- Type a new number 
- Use an unary action (Power or square root for example)
- press Enter

### Bug when users enter a number when a binary operation was previously used
a2019-03-08_4-39-13

- Type a random binary operation: `3 x 5` (for example)
- press Enter
- Type a new number 
- press Enter

### Reason:
- `m_bNoPrevEqu`, `m_nOpCode` and `m_nOpCode` aren't correctly reset when necessary
- the code assuming that the last operation was a binary one when users click on Enter more than 1 time.


uhliksk:
Dear Rudyhuyn, you don't understand how scientific calculator work. This is not bug. This is precisely designed feature. For me personally it is very usefull to not type whole expression for repeated calculations. I suggest to not change that behavior.

rudyhuyn:
> Dear Rudyhuyn, you don't understand how scientific calculator work. This is not bug. This is precisely designed feature. For me personally it is very usefull to not type whole expression for repeated calculations. I suggest to not change that behavior.

The issue isn't repeated operations (you enter `2+3` then `Enter` (app displays 5), `Enter` again (app displays 8), the code doesn't change this behavior.

The issue is when you type: `2+3`, then `Enter`, **then you type 4 to start a new calculation and Enter** -> the app will display `4+3  = 7` for no reasons (and if you type `Enter` again, it will display `12 + 3 = 15`.

uhliksk:
> The issue isn't repeated operations (you enter 2+3 then Enter (app displays 5), Enter again (app displays 8), the code doesn't change this behavior.
> The issue is when you type: 2+3, then Enter, then you type 4 to start a new calculation and Enter -> the app will display 4+3  = 7 for no reasons (and if you type Enter again, it will display 12 + 3 = 15.

I know. I'm using exactly that behavior. If I need to multiplicate lot of numbers with same multiplier I just type that once and than I just type number and Enter, then another number and Enter, etc. It's much easier than typing same multiplier again and again. I'm using this behavior since my first "physical" digital calculator to calculate numerical series. It is implemented right way in Microsoft calculator and this one really should not be changed.

uhliksk:
> The issue isn't repeated operations (you enter 2+3 then Enter (app displays 5), Enter again (app displays 8), the code doesn't change this behavior.
> The issue is when you type: 2+3, then Enter, then you type 4 to start a new calculation and Enter -> the app will display 4+3  = 7 for no reasons (and if you type Enter again, it will display 12 + 3 = 15.

I know. I'm using exactly that behavior. If I need to multiplicate lot of numbers with same multiplier I just type that once and then I just type number and Enter, then another number and Enter, etc. It's much easier than typing same multiplier again and again. I'm using this behavior since my first "physical" digital calculator to calculate numerical series. It is implemented right way in Microsoft calculator and this one really should not be changed.

rudyhuyn:
> > The issue isn't repeated operations (you enter 2+3 then Enter (app displays 5), Enter again (app displays 8), the code doesn't change this behavior.
> > The issue is when you type: 2+3, then Enter, then you type 4 to start a new calculation and Enter -> the app will display 4+3  = 7 for no reasons (and if you type Enter again, it will display 12 + 3 = 15.
> 
> I know. I'm using exactly that behavior. If I need to multiplicate lot of numbers with same multiplier I just type that once and then I just type number and Enter, then another number and Enter, etc. It's much easier than typing same multiplier again and again. I'm using this behavior since my first "physical" digital calculator to calculate numerical series. It is implemented right way in Microsoft calculator and this one really should not be changed.

Interesting feedback. This is not the behavior I have with my own calculator, but I was able to reproduce this behavior with the Win7 Calculator (it makes sense, it's based on the same lib).

I will let Microsoft decide if it's a "Feature" or a "Bug" :D

HowardWolosky:
Let's move this over to an Issue as we're doing for your other PR's right now that involve a _behavior change_ to the app.

Our team is standardizing on a flow where a _behavior change_ needs to have an associated Issue which will provide the opportunity for discussion and approval before we'll consider merging in the corresponding code change.

I recognize that there has been a lot of discussion already in this PR thread, which is _great_.  I think the benefit of the Issue is that it will now be able to summarize how things have evolved so far, and we'll take the conversation forward there.

PR #205, opened 2019-03-08
Fix punctuation in README.md

whutddk:
- some punctuation in wrong place, and some of them missing

-
-
-


ben174:
Oh come on! put in a little effort if you want to commit to this repo.

whutddk:
@ben174 Hello,, may be you are right. But there are some mistakes right? Compare to open an issue to tell developer, correct them directly will be better I guess.

PR #208, opened 2019-03-08
Replace custom ARRAYSIZE macro with std::size

janisozaur:


src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -16,6 +16,8 @@
 #include "Header Files/CalcEngine.h"
 #include "Header Files/CalcUtils.h"
 
+#include <iterator>
```
danbelcher-MSFT: Move standard headers to the pch so they can be compiled once and used throughout the project.

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -38,11 +40,11 @@ INT NPrecedenceOfOp(int nopCode)
     int iPrec;
 
     iPrec = 0;
-    while ((iPrec < ARRAYSIZE(rgbPrec)) && (nopCode != rgbPrec[iPrec]))
+    while ((iPrec < std::size(rgbPrec)) && (nopCode != rgbPrec[iPrec]))
```
danbelcher-MSFT: Add `using namespace std;` for cpp files and drop the fully qualified name when standard functions.

src/CalcViewModel/Common/LocalizationSettings.h
```diff
@@ -39,7 +41,7 @@ namespace CalculatorApp
                     result = GetLocaleInfoEx(m_resolvedName,
                         LOCALE_SDECIMAL,
                         decimalString,
-                        ARRAYSIZE(decimalString));
+                        std::size(decimalString));
```
danbelcher-MSFT: This (and others in this file/change) introduce new compilation warnings.  #197 turns on 'Treat warnings as errors' and then this will be a problem.

`1>calculator\src\calcviewmodel\common\localizationsettings.h(44): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data (compiling source file Common\EngineResourceProvider.cpp)`

Resolve with `static_cast`.

src/CalcManager/CEngine/scicomm.cpp
janisozaur: I'm not in favour of that. PCH are commonly hard to work with and include unnecessary items that you wouldn't otherwise have.

danbelcher-MSFT: Why is it hard to work with?  The STL headers aren't changing between compilation runs.  These projects already use PCH and so this change is abnormal.  I'm open to a community discussion for whether we should use PCH or not, but it's working fine right now and I don't see an immediate reason to move away.

danbelcher-MSFT: ^ Sorry, meant standard library headers in general

janisozaur: The problematic part is many (most?) build systems don't support them adequately and forcing users to compile everything C++ and the project has to offer for each translation unit is a huge overkill.

Knowing how PCHs work, I would rather imagine the project to be set up so that only the minimal required set of headers is included, then the build system supporting PCH injects it into translation unit.

janisozaur:
Updated

src/CalcManager/CEngine/scicomm.cpp
danbelcher-MSFT: I'm happy to have a community discussion on how we can properly support pch or move to no pch.  That conversation should not be decided by this PR.  Until we have that conversation, we should follow the standards of the codebase, and the convention here is to keep standard headers in the pch.  In the new discussion, we can compare compilation times, how they are improved by pch, and what the impact would be for us if we move off of pch and update source files to only include what is needed (as you are suggesting here).

janisozaur: I have updated it as requested, will file a ticket for discussing PCH later.

janisozaur:
Ping @danbelcher-MSFT, this should be ready for merging

src/CalcManager/CEngine/scicomm.cpp
janisozaur: I filed a ticket for reworking PCHs: https://github.com/Microsoft/calculator/issues/324

PR #210, opened 2019-03-08
fix: minor Azerbaijani translation error

Maharramoff:
- Single word

- Unsuitable for testing

### DIFF
--- a/src/Calculator/Resources/az-Latn-AZ/CEngineStrings.resw
+++ b/src/Calculator/Resources/az-Latn-AZ/CEngineStrings.resw
@@ -150,7 +150,7 @@
     <comment>Same 107</comment>
   </data>
   <data name="99" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


HowardWolosky:
Thanks for the submission. We definitely want to get these translations right for all of our customers. I've still discussing with our localization team on their preferred way to handle these types of changes, and will update this PR and our documentation once I hear back.

HowardWolosky:
As mentioned before, we cannot directly accept localization changes directly through this repo.  I've worked with our Localization team to identify a temporary workaround to make sure that this contribution isn't lost.  I have filed an issue in our internal tracking system (`20785608`) which tracks your suggested change, and the Localization team will use that internal issue to review and apply your change via their system.

Your participation is very much appreciated, and your work _will_ get applied.  Unfortunately, I cannot directly merge this change into this repo due to the way our system currently works.

Thanks again.

PR #211, opened 2019-03-08
Add CMake/Clang support for the Calculator Engine

### LINKED ISSUE #109: Alternative compilers
Support Clang and/or GCC

fwcd:
- Added CMake with a library target for the engine
- Added cross-platform headers for common Win32 error codes
- Added typealiases for Win32 data types

- Manual testing using `Apple LLVM version 10.0.0 (clang-1000.11.45.5)` on macOS 10.13.6 High Sierra


CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_minimum_required(VERSION 2.6)
+project(calculator)
```
janisozaur:
```suggestion
project(calculator CXX)
```

src/CalcManager/CEngine/calc.cpp
```diff
@@ -19,8 +26,8 @@ static constexpr long DEFAULT_RADIX = 10;
 
 static constexpr wchar_t DEFAULT_DEC_SEPARATOR = L'.';
 static constexpr wchar_t DEFAULT_GRP_SEPARATOR = L',';
-static constexpr wstring_view DEFAULT_GRP_STR = L"3;0";
-static constexpr wstring_view DEFAULT_NUMBER_STR = L"0";
+static const wstring_view DEFAULT_GRP_STR = L"3;0";
```
janisozaur: why?

src/CalcManager/win_data_types_cross_platform.h
```diff
@@ -0,0 +1,26 @@
...
+typedef unsigned short WORD;
+typedef UINT_PTR WPARAM;
+typedef ULONG_PTR DWORD_PTR;
+
+#define LOWORD(dw) ((WORD)(((DWORD_PTR)(dw)) & 0xffff))
+#define HIWORD(dw) ((WORD)((((DWORD_PTR)(dw)) >> 16) & 0xffff))
+#define LODWORD(qw) ((DWORD)(qw))
+#define HIDWORD(qw) ((DWORD)(((qw) >> 32) & 0xffffffff))
+
+#define ARRAYSIZE(a) (sizeof(a) / sizeof(*a))
```
janisozaur: https://github.com/Microsoft/calculator/pull/208

src/CalcManager/win_data_types_cross_platform.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <cstdint>
+
+typedef uint64_t ULONGLONG;
+typedef int32_t INT;
+typedef uint8_t CHAR;
+typedef int32_t LONG;
+typedef uint8_t BYTE;
+typedef unsigned int UINT_PTR;
```
janisozaur: That's incorrect, should be `uintptr_t`

src/CalcManager/CEngine/calc.cpp
fwcd: Explained in 9bfb78e: Clang does not consider `char_traits<wchar_t>::length` a `constexpr` function:

```
/.../WindowsCalculator/src/CalcManager/CEngine/calc.cpp:29:31: error: constexpr variable 'DEFAULT_GRP_STR' must be initialized by a
      constant expression
static constexpr wstring_view DEFAULT_GRP_STR = L"3;0";
                              ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /.../WindowsCalculator/src/CalcManager/CEngine/calc.cpp:7:
In file included from /.../WindowsCalculator/src/CalcManager/pch_cross_platform.h:4:
In file included from /.../usr/include/c++/v1/sstream:174:
In file included from /.../usr/include/c++/v1/ostream:138:
In file included from /.../usr/include/c++/v1/ios:216:
In file included from /.../usr/include/c++/v1/__locale:15:
In file included from /.../usr/include/c++/v1/string:477:
/.../usr/include/c++/v1/string_view:236:31: note: non-constexpr function 'length'
      cannot be used in a constant expression
        : __data(__s), __size(_Traits::length(__s)) {}
                              ^
/.../WindowsCalculator/src/CalcManager/CEngine/calc.cpp:29:49: note: in call to 'basic_string_view(&L"3;0"[0])'
static constexpr wstring_view DEFAULT_GRP_STR = L"3;0";
                                                ^
In file included from /.../WindowsCalculator/src/CalcManager/CEngine/calc.cpp:7:
In file included from /.../WindowsCalculator/src/CalcManager/pch_cross_platform.h:4:
In file included from /.../usr/include/c++/v1/sstream:174:
In file included from /.../usr/include/c++/v1/ostream:138:
In file included from /.../usr/include/c++/v1/ios:216:
In file included from /.../usr/include/c++/v1/__locale:15:
In file included from /.../usr/include/c++/v1/string:477:
In file included from /.../usr/include/c++/v1/string_view:176:
/.../usr/include/c++/v1/__string:356:23: note: declared here
char_traits<wchar_t>::length(const char_type* __s) _NOEXCEPT
                      ^
3 warnings and 1 error generated.

** BUILD FAILED **
```

janisozaur:
@fwcd please see https://github.com/Microsoft/calculator/pull/212, I was halfway through that when you submitted your changes. I asked Microsoft team to explain how would they want to pursue these changes.

fwcd:
@janisozaur Looks great, that is probably an even better solution than mine!

src/CalcManager/win_data_types_cross_platform.h
fwcd: Resolved with the newest commit, thanks

src/CalcManager/win_data_types_cross_platform.h
```diff
@@ -7,14 +7,14 @@ typedef int32_t INT;
...
-typedef unsigned long DWORD;
-typedef unsigned long ULONG;
-typedef unsigned short USHORT;
+typedef uint32_t ULONG32;
+typedef uint32_t DWORD;
+typedef uint32_t ULONG;
+typedef uint16_t USHORT;
 typedef wchar_t WCHAR;
-typedef unsigned short WORD;
+typedef uint16_t WORD;
```
fwcd: These values are based on https://docs.microsoft.com/en-us/windows/desktop/WinProg/windows-data-types

janisozaur: I know, I used the same page (and filed feedback regarding the errors there)

CMakeLists.txt
fwcd: Thanks

fwcd:
This branch now successfully builds both on Windows with `msbuild` and on macOS using `xcodebuild` (through CMake).

cup:
https://github.com/Microsoft/calculator/issues/109
https://github.com/Microsoft/calculator/pull/211
https://github.com/Microsoft/calculator/pull/212

HowardWolosky:
@mcooley -- What would it take to get the pipeline updated as part of this change so that we could have PR/CI validation moving forward to ensure that a change doesn't _break_ this support?

src/CalcManager/CEngine/calc.cpp
chuggafan: https://github.com/llvm-mirror/libcxx/blob/master/include/__string#L130 (and for the wchar_t specialization: https://github.com/llvm-mirror/libcxx/blob/master/include/__string#L356 ) https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/char_traits.h#L110 Both libcxx and gcc say that after c++14 char_traits::length is constexpr and is required for std::string_view for it to be so, so either you're doing this on mac with it's REALLY outdated libcxx, you're on an old version of clang, or you forgot to compile with --std=c++17.

fwcd: @chuggafan You are correct, I committed this before bumping the version to C++17. Fixed with the latest commit, thanks!

mcooley:
> @mcooley -- What would it take to get the pipeline updated as part of this change so that we could have PR/CI validation moving forward to ensure that a change doesn't _break_ this support?

Agreed, we should have some CI support so we don't break this. We currently have a job per configuration and currently one configuration per architecture

src/CalcManager/win_data_types_cross_platform.h
tdmllc1: looks good

tdmllc1:
Thank You

danbelcher-MSFT:
Just to recap, there are two PRs right now that start to address this issue, #211 and #212.  We're holding off on reviewing and merging those PRs until we can build in a macOS or Linux environment as part of our CI, so we don't regress this capability later.

src/CalcManager/sal_cross_platform.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+// Empty macro definitions for source annotations
+
+#define _In_opt_
+#define _Out_opt_
+#define _In_
+#define _Out_
+#define _Inout_
+#define __in_opt
```
janisozaur: There's another one being added in #304: `_Frees_ptr_opt_`

src/CalcManager/win_data_types_cross_platform.h
janisozaur: #208 had landed

src/CalcManager/CEngine/Number.cpp
```diff
@@ -59,6 +59,13 @@ namespace CalcEngine
...
     {
-        return all_of(m_mantissa.begin(), m_mantissa.end(), [](auto &&i) { return i == 0; });
+        for (auto const& digit : Mantissa())
+        {
+            if (digit != 0)
+            {
+                return false;
+            }
+        }
+        return true;
```
janisozaur: Why did this `all_of` have to change?

fwcd: See: 49a291d9198045bdde2e16a7f7ef1b30f48c6103

chuggafan: https://godbolt.org/z/eMGp-e Clang-5 has all_of capabilities, so most likely you're having problems with apple's clang or libcxx being bad....  it might have had something to do with the "not using c++17 standard" thing I mentioned in a previous comment

src/CalcManager/sal_cross_platform.h
fwcd: Added, thanks.

src/CalcManager/CEngine/Number.cpp
janisozaur: @fwcd are you sure? I'm able to compile this with clang as ancient as 3.4 and GCC 4.9: https://godbolt.org/z/FflRkU

CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_minimum_required(VERSION 2.6)
```
janisozaur: You need at least 3.8 to support
```
set(CMAKE_CXX_STANDARD 17)
```
Which you call a few lines lower.
```suggestion
cmake_minimum_required(VERSION 3.8)
```

src/CalcManager/CEngine/Number.cpp
fwcd: @chuggafan That was indeed the issue, thanks.

src/CalcManager/win_data_types_cross_platform.h
```diff
@@ -0,0 +1,26 @@
...
+typedef uint32_t DWORD;
+typedef uint32_t ULONG;
+typedef uint16_t USHORT;
+typedef wchar_t WCHAR;
+typedef uint16_t WORD;
+typedef UINT_PTR WPARAM;
+typedef ULONG_PTR DWORD_PTR;
+
+#define LOWORD(dw) ((WORD)(((DWORD_PTR)(dw)) & 0xffff))
+#define HIWORD(dw) ((WORD)((((DWORD_PTR)(dw)) >> 16) & 0xffff))
```
janisozaur: Unused
```suggestion
```

src/CalcManager/win_data_types_cross_platform.h
janisozaur:
```suggestion
```
Not needed anymore

src/CalcManager/win_data_types_cross_platform.h
janisozaur: This whole block should not be needed anymore once https://github.com/Microsoft/calculator/pull/212 lands

src/CalcManager/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+	CEngine/Number.cpp
+	CEngine/Rational.cpp
+	CEngine/RationalMath.cpp
+	CEngine/scicomm.cpp
+	CEngine/scidisp.cpp
+	CEngine/scifunc.cpp
+	CEngine/scioper.cpp
+	CEngine/sciset.cpp
+)
+target_compile_options(CalculatorEngine PRIVATE "-fms-extensions")
```
janisozaur: Is there anything that still requires `-fms-extensions`?

src/CalcManager/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+add_library(CalculatorEngine
```
janisozaur: When #212 lands, you should also add `Ratpack` library.

janisozaur:
#212 was just merged, @fwcd will you update this PR?

fwcd:
I have rebased this branch onto the current master, which continues to build successfully under macOS.

src/CalcManager/CMakeLists.txt
fwcd: @janisozaur `Ratpack` support is non-trivial due to dependencies on `intsafe.h` and should be added in a separate PR.

fwcd: The C++ standard does not support flexible array members inside structs:

https://github.com/Microsoft/calculator/blob/7a48f66807fe87c98af9781fda5a526faa13aa86/src/CalcManager/Ratpack/ratpak.h#L53-L66

`Ratpack`, however, heavily relies on this nonstandard extension, pointer arithmetic and memory reordering, which makes the port a non-trivial change that should be handled in a separate PR instead.

I have created an experimental version using `std::vector` in https://github.com/fwcd/WindowsCalculator/tree/ratpack that simply exposes the underlying data array. This compiles with clang, although it is far from being an ideal solution.

fwcd: Tracked in https://github.com/fwcd/WindowsCalculator/pull/2.

DeveloperPaul123:
I'll that this also build in the VS 2019 preview on Windows 10.

janisozaur:
@DeveloperPaul123 you seem to have missed a verb there.

src/CalcManager/win_data_types_cross_platform.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <cstdint>
```
janisozaur: With the typedefs below removed, it won't be required

src/CalcManager/win_data_types_cross_platform.h
```diff
@@ -0,0 +1,26 @@
...
+typedef uint16_t USHORT;
+typedef wchar_t WCHAR;
+typedef uint16_t WORD;
+typedef UINT_PTR WPARAM;
+typedef ULONG_PTR DWORD_PTR;
+
+#define LOWORD(dw) ((WORD)(((DWORD_PTR)(dw)) & 0xffff))
+#define HIWORD(dw) ((WORD)((((DWORD_PTR)(dw)) >> 16) & 0xffff))
+#define LODWORD(qw) ((DWORD)(qw))
+#define HIDWORD(qw) ((DWORD)(((qw) >> 32) & 0xffffffff))
```
janisozaur: These two macros should be the only thing required from this file.

src/CalcManager/winerror_cross_platform.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <cstdint>
+
+typedef int32_t HRESULT;
```
janisozaur: Should not be needed anymore

src/CalcManager/winerror_cross_platform.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <cstdint>
```
janisozaur: Should not be needed anymore, once the typedef below is gone.

src/CalcManager/pch.h
```diff
@@ -24,6 +22,29 @@
...
 #include <unordered_map>
-#include <intsafe.h>
 #include <array>
+#include <string_view>
+
+#if defined(_WIN32) && defined(_MSC_VER)
+
+#include <windows.h>
+#include <winerror.h>
+#include <intsafe.h>
```
janisozaur:
```suggestion
```
This should not be required anymore, if you provide the `HIDWORD` and `LODWORD` macros for all the platforms

This should also let you add Ratpack library. See https://github.com/Microsoft/calculator/pull/398 where I remove this include

src/CalcManager/pch.h
```diff
@@ -24,6 +22,29 @@
...
 #include <ppltasks.h>
+
+#else
+
+// #include <intsafe.h>
+// #include <ppltasks.h>
```
janisozaur: remove

src/CalcManager/pch.h
fwcd: @janisozaur Have so seen https://github.com/Microsoft/calculator/pull/211#discussion_r269383211 and https://github.com/fwcd/WindowsCalculator/pull/2? The issue with Ratpack ist actually a bit deeper than that.

fwcd: `-fms-extension` does not provide support for C++ flexible array members

fwcd:
@janisozaur I have squashed the commits at a granularity level of roughly 7 commits each and updated the commit messages to present tense. Maybe it would make more sense to squash the entire PR into one commit once ready?

fwcd:
Since using the cross-platform versions of the headers causes trouble with MSVC due to duplicate declarations, I have reverted `pch.h` to use the Windows headers again.

src/CalcManager/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+	CEngine/History.cpp
+	CEngine/Number.cpp
+	CEngine/Rational.cpp
+	CEngine/RationalMath.cpp
+	CEngine/scicomm.cpp
+	CEngine/scidisp.cpp
+	CEngine/scifunc.cpp
+	CEngine/scioper.cpp
+	CEngine/sciset.cpp
+)
```
danbelcher-MSFT: I'd like to avoid fragmentation between what it means to compile this project with MSVC or Clang.  With MSVC, we fully compile CalcManager and produce CalcManager.lib.  With Clang/CMake, we'd only be compiling CEngine and producing CEngine.lib.  If that's the case, I don't think think putting this file in the CalcManager folder is the right place for it, because it's not a goal for this file to compile CalcManager.  It seems like this should be moved to the CEngine folder.

Separately, I'm curious if your intention is to add support for building other parts of the CalcManager folder in a separate change, or are you only interested in compiling CEngine?  @janisozaur, I'm wondering what your interest is as well.  If people only want CEngine, then we should move it to its own project that we can compile consistently with Clang or MSVC.

src/CalcManager/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+	CEngine/CalcUtils.cpp
+	CEngine/History.cpp
+	CEngine/Number.cpp
+	CEngine/Rational.cpp
+	CEngine/RationalMath.cpp
+	CEngine/scicomm.cpp
+	CEngine/scidisp.cpp
+	CEngine/scifunc.cpp
+	CEngine/scioper.cpp
+	CEngine/sciset.cpp
```
danbelcher-MSFT: How is the resulting lib functional if you haven't compiled Ratpak?  There is no definition for the various extern functions that are compiled into this library through the included ratpak headers.  Is that intentional?  Are you wanting to add support in pieces?  Each update that we merge into master should be functional so it seems like this file needs more work.

src/CalcManager/pch.h
```diff
@@ -24,6 +22,26 @@
 #include <list>
 #include <regex>
 #include <unordered_map>
-#include <intsafe.h>
 #include <array>
+#include <string_view>
+
+#include "win_data_types_cross_platform.h"
+
+#if defined(_WIN32) && defined(_MSC_VER)
```
danbelcher-MSFT: I think logic like this should be moved into each `_cross_platform.h` file.  For example, `winerror_cross_platform.h` could have:
```C++
#if defined(_WIN32) && defined(_MSC_VER)
#include <winerror.h>
#else 
typedef int32_t HRESULT;
...
#endif 
```

Then, any file in the project that needs error code types can `#include winerror_cross_platform.h` and the right thing will happen.

README.md
```diff
@@ -17,6 +17,17 @@ Calculator ships regularly with new features and bug fixes. You can get the late
 - Currency conversion based on data retrieved from [Bing](https://www.bing.com).
 
 ## Getting started
+
+### On macOS
```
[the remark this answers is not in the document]
fwcd: Thanks for pointing this out, that was not intentional.

src/CalcManager/Header Files/EngineStrings.h
```diff
@@ -214,7 +219,7 @@
 #define SIDS_ERR_INPUT_OVERFLOW L"119"
 #define SIDS_ERR_OUTPUT_OVERFLOW L"120"
 
-__declspec(selectany) std::wstring g_sids[] =
+DECLSPEC_SELECTANY std::wstring g_sids[] =
```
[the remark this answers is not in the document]
fwcd: Fixed by 55e12f0.

DeveloperPaul123:
On a slightly unrelated note, is there any sort of documentation for the calculator engine?

fwcd:
@DeveloperPaul123 There is a section about the engine here: https://github.com/Microsoft/calculator/blob/master/docs/ApplicationArchitecture.md#model

src/CalcManager/winerror_cross_platform.h
fwcd: @danbelcher-MSFT I forgot to open it again after reverting the change that originally fixed it, my mistake.

Currently, however, the include is required since the `SUCCEEDED` depends on `HRESULT`, which in turn depends on `uint32_t`.

src/CalcManager/winerror_cross_platform.h
fwcd: @janisozaur Should this really be removed even though `SUCCEEDED` and `FAILED` depend on it?

src/CalcManager/CMakeLists.txt
fwcd: This should be fixed by the commits 752c3b5 and 8baac74, which package `CalcManager` into a single library and move the inclusions of `CEngine` and `Ratpack` source files into their own directories.

I will, however, leave this conversation open for discussion, since the question remains whether the engine would deserve its own project.

src/CalcManager/CMakeLists.txt
fwcd: The resulting library was indeed not functional without Ratpack, which should be included now though. (Ratpack support was previously tracked by https://github.com/fwcd/WindowsCalculator/pull/2)

fwcd:
Currently, the calculator builds fine on Windows:

image

Ratpack, however, does not build on Linux yet, since Clang does not support direct initialization of flexible array members, as used in `ratconst.h`:

https://github.com/Microsoft/calculator/blob/a32a5e4ce41901fc444c48ba360ab35f5cb791d3/src/CalcManager/Ratpack/ratconst.h#L7-L12

Possible solutions for this problem would include migrating to a more standards-compliant way of allocating the mantissa (for example with an `std::vector`). This would require rewriting large parts of Ratpack though.

Any thoughts on this?

danbelcher-MSFT:
From @fwcd
> Possible solutions for this problem would include migrating to a more standards-compliant way of allocating the mantissa (for example with an `std::vector`). This would require rewriting large parts of Ratpack though.
> 
> Any thoughts on this?

Absolutely in support.  In general, you can tell we're trying to move the codebase to be standards-compliant which increases the portability of the engine.

danbelcher-MSFT:
@DeveloperPaul123 
> On a slightly unrelated note, is there any sort of documentation for the calculator engine?

Unfortunately, there's not much.  The codebase has been handed around Microsoft for 20+ years and each team has onboarded separately.  Now that the source is available publicly, there's opportunity from the community to both improve the standards and readability of the code as well as invest in documentation.  As @fwcd pointed out, there's a starter area where documentation can be added.

README.md
```diff
@@ -35,6 +37,14 @@ Prerequisites:
 - Open [src\Calculator.sln](/src/Calculator.sln) in Visual Studio to build and run the Calculator app.
 - For a general description of the Calculator project architecture see [ApplicationArchitecture.md](docs/ApplicationArchitecture.md).
 
+### On macOS/Linux
+Prerequisites:
+- CMake 3.8 or newer
```
janisozaur:
```suggestion
- CMake 3.8 or newer
- C++17-compatible compiler, e.g. GCC 7 or Clang 4
```

src/CalcManager/Header Files/CalcEngine.h
```diff
@@ -27,6 +32,8 @@
 #include "Rational.h"
 #include "RationalMath.h"
 
+#include <array>
```
janisozaur: Nice, this did not come up in #436.

src/CalcManager/pch.h
```diff
@@ -24,6 +22,22 @@
...
+#include <string_view>
+#include <future>
+
+#include "win_data_types_cross_platform.h"
+
+#if defined(_WIN32) && defined(_MSC_VER)
+
+#include <windows.h>
+#include <winerror.h>
 #include <ppltasks.h>
```
janisozaur: Is `ppltasks.h` still a required include with changes to `src/CalcManager/UnitConverter.{cpp,h}`?

src/CalcManager/winerror_cross_platform.h
janisozaur: `HRESULT` is already converted to a standard type: https://github.com/Microsoft/calculator/pull/212 / https://github.com/Microsoft/calculator/pull/212/commits/15e7e8d0aa628329588171cd9514ee6a689d0f3c, so yes.

fwcd:
@janisozaur The mantissa reimplementation using `std::vector` might be non-trivial, since there is a lot of code relying on manual memory management currently. I tried to work around it by replacing pointer references to `mant` with `vector.data()`, but that did not work out, since empty `vector`s internally uses null pointers.

DeveloperPaul123:
@danbelcher-MSFT If there is a desire for documentation from the community, would this come in the form of a github based wiki? Or did you have in mind documentation at the code level itself (i.e. doxygen style comments)?

danbelcher-MSFT:
General documentation describing the functionality of the engine seems like a good place to start.  Function documentation is also good, although I've seen it become overblown.  I personally prefer to follow along with the code.  I think the goal of adding documentation should be to make it easier for newcomers to enter the project.  Exactly how we do that should be a discussion we have elsewhere.

janisozaur:
@fwcd, ah fair point. I only checked if it compiles, as the test suite is Windows-only. That said, is it something that test suite covers?

src/CalcManager/pch.h
fwcd: Fixed by 10a9a9f

README.md
fwcd: Fixed by a05bb24

src/CalcManager/pch.h
fwcd: Fixed by 9a03400

src/CalcManager/pch.h
```diff
@@ -24,6 +22,26 @@
 #include <list>
 #include <regex>
 #include <unordered_map>
-#include <intsafe.h>
 #include <array>
+#include <string_view>
+
+#include "win_data_types_cross_platform.h"
```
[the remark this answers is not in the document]
fwcd: Fixed by 3916522

src/CalcManager/Ratpack/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+target_sources(CalcManager PUBLIC
```
janisozaur:
```
CMake Warning (dev) at src/CalcManager/Ratpack/CMakeLists.txt:1 (target_sources):
  Policy CMP0076 is not set: target_sources() command converts relative paths
  to absolute.  Run "cmake --help-policy CMP0076" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

  A private source from a directory other than that of target "CalcManager"
  has a relative path.
```

src/CalcManager/CalcManager.vcxproj
```diff
@@ -303,6 +303,10 @@
     <ClInclude Include="Ratpack\ratconst.h" />
     <ClInclude Include="Ratpack\ratpak.h" />
     <ClInclude Include="UnitConverter.h" />
```
janisozaur: `src/CalcManager/CalculatorManager.cpp` needs `<climits>` include. I can't comment on that file.

src/CalcManager/UnitConverter.cpp
```diff
@@ -5,7 +5,6 @@
 #include "Command.h"
 #include "UnitConverter.h"
```
janisozaur: Needs `<cmath>` include

fwcd: Fixed by b3694fa

src/CalcManager/CalcManager.vcxproj
fwcd: Fixed by b3694fa

src/CalcManager/Ratpack/CMakeLists.txt
fwcd: Fixed by a2a3489

fwcd:
The mantissa reimplementation is not quite finished yet. Once that is done, I would suggest merging this PR before #436, since both this branch and `master` still rely on pch's, thus the include changes could be applied on top pretty naturally.

fwcd:
@janisozaur Are there any concrete issues with `libstdc++` that could be fixed? I cannot reproduce this issue on macOS.

janisozaur:
https://github.com/Microsoft/calculator/pull/321

src/CalcManager/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_library(CalcManager
+	CalculatorHistory.cpp
+	CalculatorManager.cpp
+	ExpressionCommand.cpp
+	pch.cpp
```
janisozaur: I'm at odds when it comes to this file. It's not required for the `CalcManager` library at all, it does not provide any symbols, but then again it makes sense to test its compilability. It would probably be best to leave it be for now, though maybe with a comment explaining the purpose of this file being here. I will try to mimic #436 with cmake in the future and probably address this issue.

janisozaur:
I would also recommend cherry-picking https://github.com/Microsoft/calculator/commit/24ea8de23307353f4f81ae823a30af1d55310e6f to run the test suite on all supported architectures. Once the CIs are all green, the links to runs can be collected and the commit removed (or reverted, if the PR is to be squashed). See https://github.com/Microsoft/calculator/pull/399

janisozaur:
You can use this to address the `libc++` issue:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3ab801d..ef347a5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,4 +2,10 @@ cmake_minimum_required(VERSION 3.13)
 project(calculator CXX)
 set(CMAKE_CXX_STANDARD 17)
 
+if (CMAKE_CXX_COMPILER MATCHES ".*clang")
+    # Clang disagress with libstdc++ about constexpr-ness of wstring_view
+    # See https://github.com/Microsoft/calculator/pull/321
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
+endif ()
+
 add_subdirectory(src)
```

src/CalcManager/preprocessor_cross_platform.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#if !defined(_WIN32) || !defined(_MSC_VER)
+
+// Empty macro definitions for Windows-specific preprocessor directives
+
+#define __pragma(x)
```
janisozaur:
```suggestion
#ifndef __pragma
#define __pragma(x)
#endif
```

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -206,16 +209,18 @@ PNUMBER _createnum( _In_ uint32_t size )
     PNUMBER pnumret= nullptr;
     uint32_t cbAlloc;
```
janisozaur: Azure complains:
`src\calcmanager\ratpack\conv.cpp(212,0): Error C4700: uninitialized local variable 'cbAlloc' used`

fwcd: I know, working on it..

fwcd:
The currently revision of the application builds fine, but crashes instantly with a vague exception message:

`Unhandled exception at 0x00007FFAC714AFAC (ucrtbased.dll) in Calculator.exe: An invalid parameter was passed to a function that considers invalid parameters fatal.
`

Unfortunately, the call stack does not leave any hints on what the referenced function is:

image

The highest-source level stack frame is not very helpful either:

https://github.com/Microsoft/calculator/blob/af41a183a7984d505ce45bca2b255ac1488bc977/src/Calculator/App.xaml.cpp#L292

image

Any ideas on what the issue could be?

fwcd:
_Finally_ the calculator builds and runs on Windows as it is supposed to.

Unfortunately, the calculator suffers from some severe performance issues currently, which is especially noticeable when invoking a computationally expensive operation, such as the square root. The application freezes for two or three seconds with one processor on 100% until the result is displayed:

image

Most likely this is caused by the new implementation of the mantissa. Since `std::vector` manages a dynamically allocated array internally (instead of allocating the elements in-place as it was the case with flexible array members in the previous implementation), the application requires much more dynamic allocation even for basic operations.

This raises the question whether using `std::vector` is even a viable option in the long run. What might be possible too, is to introduce templates and determine the mantissa size statically, this would however require an even larger rewrite and make polymorphism trickier.

Instead, I would suggest using either a static allocator (such as [the one used in Chromium](https://chromium.googlesource.com/chromium/chromium/+/master/base/stack_container.h)) or something similar to [`boost::static_vector`](https://www.boost.org/doc/libs/1_55_0/doc/html/container/non_standard_containers.html#container.non_standard_containers.static_vector).

Both use a fixed-size stack-buffer that overflows onto the heap if the vector grows too large.

@danbelcher-MSFT What do you think?

fwcd:
A simpler solution might be to stack-allocate numbers directly and pass them by value. This is possible, because the structure has a fixed size with vectors dynamically allocating memory internally.

Not sure whether that would fit into the architecture though.

danbelcher-MSFT:
@fwcd, I'm looking over the change.  What concerns me the most right now is the noticeable performance decrease in debug builds.  I'll talk it over with our team to see what, if anything we can/should do, and whether we can take the change.

src/CalcManager/CMakeLists.txt
danbelcher-MSFT: @fwcd, @janisozaur, what are your thoughts on us distributing the CalcEngine separately?  CalcManager is very specific to the Windows Calculator app, as it manages the three engine instances needed for Standard, Scientific, and Programmer modes.  We could move CalcManager/UnitConverter out and leave CalcEngine/Ratpak as its own project.  Developers who only want to use CalcEngine would have one project to build in the repo.  We could also make this eaiser by publishing offical distributions of the CalcEngine lib at the same time we publish a release of Windows Calculator to the store.

fwcd:
Dynamic allocation might be a performance issue since not just the `NUMBER` struct, but also the `vector` is allocated on the heap. Since the struct has a fixed size now, many numbers could probably be allocated on the stack instead.

src/CalcManager/CMakeLists.txt
janisozaur: What do you mean by "distributing separately"? Does that imply moving it to a separate repository? Would you create another Visual Studio project for that library? Would there be any benefit on your side?

danbelcher-MSFT: By "distributing separately" I mean we would find a way to distribute the static lib, for example, through a native NuGet package.  Maybe simply tagging the commits is enough and consumers can use vcpkg to incorporate CalcEngine into their projects.  We don't have a concrete plan for this right now, we're just wondering if the interest is there in the community.  In general, it seems that some contributors are only interested in the CalcEngine part of the repo and we benefit if we make it easier for those contributors to build, consume, and improve the CalcEngine.

janisozaur: That sounds interesting, but I'm not convinced there's a demand for CalcEngine to be packaged as such a library. Maintaining such a library for external consumers would mean maintaining API and ABI compatibility, fixing the test suite (#340), improving documentation and offering something that GMP, MPFR, Boost.Multiprecision and others don't already offer. I'm not really sold on the idea even though the way to achieving it would address at least some of the issues I raised, but then again I'm not your target audience. Perhaps starting small like tagging commits that you mentioned would suffice for now and when this PR gets merged and there's either https://github.com/Microsoft/calculator/issues/395 implemented or it's equivalent for MSVC, separating out a CalcEngine library should be trivial for those who need it.

fwcd:
The last two commits replace every dynamically allocated instance of `NUMBER` with a stack-allocated value. Nullable types have been converted to `std::optional`s.

This change does not only improve performance, but also make the code safer to use and prevent potential memory leaks.

src/CalcManager/Ratpack/basex.cpp
```diff
@@ -36,29 +36,29 @@ void _mulnumx( PNUMBER *pa, PNUMBER b );
 //
 //----------------------------------------------------------------------------
 
-void __inline mulnumx( PNUMBER *pa, PNUMBER b )
+void __inline mulnumx( NUMBER *pa, NUMBER b )
 
 {
-    if ( b->cdigit > 1 || b->mant[0] != 1 || b->exp != 0 )
+    if ( b.cdigit > 1 || b.mant[0] != 1 || b.exp != 0 )
```
danbelcher-MSFT: Throws an exception for me in Debug x64 due to `b.mant[0]` but `b.size()` is 0.

fwcd: Immediately at startup? From which function was `mulnumx` called?

Debug x64 builds and runs fine on my machine.

danbelcher-MSFT:
@fwcd, after [89e454a](https://github.com/Microsoft/calculator/pull/211/commits/89e454a65b1893aa112dea237f662f08c165fa9e), I'm crashing in `conv.cpp:305` for the same reason as before, stemming from a call to `CCalcEngine::InitChopNumbers`.  You're saying Debug build runs correctly for you?

src/CalcManager/Ratpack/basex.cpp
danbelcher-MSFT: Let's table this and work off the most recent commit.

fwcd:
Migrating the numbers to value types seems to be a bit more work than I originally anticipated, in particular due to code fragments that rely on reference semantics.

danbelcher-MSFT:
Hey @fwcd, this PR has gotten quite large (68 commits so far) and the scope of the change has broadened from simply adding a CMake file to now including large changes to primary classes and functions in the engine.  I find a lot of value in the work you're doing to change the mantissa type to `vector` and to optimize the ratpack functions for both performance and safety.  We discussed this PR as a team yesterday and we're asking that you bring these optimizations into the codebase in a separate PR to be merged before this one.  This will make it easier for us to review the improvements and will refocus this PR to the smaller scope of simply adding CMake/Clang support.

danbelcher-MSFT:
Ping @fwcd.  We'd still love to take all of the improvements here, both to the engine's standard's compliance as well as cmake support.

fwcd:
@danbelcher-MSFT First of all, sorry for not responding, I totally support the idea of splitting the changes up into refactorings on the engine and CMake support.

Unfortunately, I have been quite busy recently since the summer term has begun. I hope to find some spare time to make the PR ready, but until then, there is not much I can do, I am afraid.

danbelcher-MSFT:
@fwcd, no problem, we definitely understand.  The effort to increase the portability of the Calc engine is driven by the community.  I hope you do find the time because I'd like you to get credit for the contributions you've made here.

fwcd:
Before splitting up this PR, I would try to fix the tests and get the CI build to pass, so the individual PRs can focus on the actual improvements (CMake support, Ratpack optimizations).

I have faced some issues due to localization, however. Since my machine is German, the default decimal format is `123.456.789,05` instead of `123,456,789.05` which causes many unit tests to fail (incorrectly), since they expect the American format. Is there a way to get around this?

mcooley:
@fwcd great, thanks for taking a look again! It sounds like to unblock you we should go ahead and get #272 fixed. (#289 has a fix but I think it was locked by mistake--@sanderl can you review that one when you get a chance?)

danbelcher-MSFT:
Hey @fwcd, you can adjust the format used for numbers even if your system remains in German.
Open the Run dialog (Win + R), and enter `intl.cpl`.  This should open the old Region settings panel.  Click on 'Additional Settings'.  You should be in the 'Customize Format' panel where you can adjust how numbers are displayed.

I confirmed the Unit Tests passed for standard US settings.  Then I changed my formatting preferences to your machine's and confirmed the UTs fail.  Following the reverse operation should get the UTs to pass for you.  Let us know if it doesn't work.  Thanks!

janisozaur:
https://github.com/microsoft/calculator/pull/289 has closed #272 and there were some changes to how the code is formatted

mcooley:
Given how stale this PR has become, it doesn't look like we'll be able to review and merge this as-is. Thanks for the contribution; we hope that some of these changes can make it in in smaller pieces in the future.

PR #213, opened 2019-03-09
Activate TextSelection on all interesting values/results

### LINKED ISSUE #223: The application should allow text selections on calculation results.
Ctrl+C is already supported in the app, but:
- not available on tablets/devices without keyboard.
- It's not really easy to guess that you can Ctrl+C without selecting the result.
- You can't copy an operation in the history
- You can't copy a part of a result (to ignore the fractional part of a number for example)
- If we allow users to select texts, the application will feel more like a "Desktop app", will be easier to use on a tablet and users will be more prompt to use Ctrl+C to use the result in another app.

Here is a prototype:
54065190-025b8280-41d2-11e9-8869-4029cf4532ea

rudyhuyn:
Fixes #223 

Ctrl+C is already supported in the app, but:
- not Tablet/Device without keyboard-friendly
- It's not really easy to guess that you can Ctrl+C without selecting the result.
- You can't copy an operation in the history
- You can't copy a part of a result (to ignore the fractional part of a number for example)

If we allow users to select texts, the application will feel more like a "Desktop app", will be easier to use on a tablet and users will be more prompt to use Ctrl+C to use the result in another app. 

2702B760-4B01-4F28-B444-05728DA30F76

## Note:
I didn't enabled TextSelection on `CalculationResult` when the control isn't "active" (text in bold), if we do,  the textblock can't catch the `Tap` events on the textblock, needed to allow users to activate the control.

### DIFF
--- a/src/Calculator/Views/DateCalculator.xaml
+++ b/src/Calculator/Views/DateCalculator.xaml
@@ -595,8 +595,7 @@
             <!-- Resulting Date -->
             <TextBlock x:Uid="DateLabel"
                        Grid.Row="8"
-                       Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"
-                       IsTextSelectionEnabled="True"/>
+                       Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"/>
 
             <TextBlock x:Name="DateResultLabel"
                        Grid.Row="9"
--- a/src/Calculator/Views/DateCalculator.xaml
+++ b/src/Calculator/Views/DateCalculator.xaml
@@ -459,11 +459,13 @@
                        AutomationProperties.LiveSetting="Polite"
                        AutomationProperties.Name="{Binding StrDateDiffResultAutomationName}"
                        ContextFlyout="{StaticResource ResultsContextMenu}"
+                       IsTextSelectionEnabled="True"
                        Text="{Binding StrDateDiffResult}"/>
             <TextBlock Grid.Row="10"
                        Style="{ThemeResource BaseTextBlockStyle}"
                        Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"
                        ContextFlyout="{StaticResource ResultsContextMenu}"
+                       IsTextSelectionEnabled="True"
                        Text="{Binding StrDateDiffResultInDays, Mode=OneWay}"
                        Visibility="{Binding IsDiffInDays, Converter={StaticResource BooleanToVisibilityNegationConverter}}"/>
         </Grid>
@@ -593,7 +595,8 @@
             <!-- Resulting Date -->
             <TextBlock x:Uid="DateLabel"
                        Grid.Row="8"
-                       Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"/>
+                       Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"
+                       IsTextSelectionEnabled="True"/>
 
             <TextBlock x:Name="DateResultLabel"
                        Grid.Row="9"
@@ -602,6 +605,7 @@
                        AutomationProperties.LiveSetting="Polite"
                        AutomationProperties.Name="{Binding StrDateResultAutomationName}"
                        ContextFlyout="{StaticResource ResultsContextMenu}"
+                       IsTextSelectionEnabled="True"
                        Text="{Binding StrDateResult}"/>
         </Grid>
     </Grid>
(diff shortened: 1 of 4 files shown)


HowardWolosky:
Rudy, again, you're _rocking_ these changes.  Please do keep them coming.  As I've been commenting on your other PR's just now, our team is standardizing on a flow where a _behavior change_ needs to have an associated Issue which will provide the opportunity for discussion and approval before we'll consider merging in the corresponding code change.

Again, _please_ keep the contributions coming.  I'm really looking forward to getting in all of these changes once we've had a bit more time to talk them through.

HowardWolosky:
Closing this PR until it's associated with an approved issue. Locking the PR so that conversation on the problem being addressed is discussed in the Issue, so that future PR discussion can be limited to the implementation of the fix.

src/Calculator/Views/MemoryListItem.xaml
```diff
@@ -59,6 +59,7 @@
                        Style="{ThemeResource TitleTextBlockStyle}"
                        FontWeight="SemiBold"
                        FlowDirection="LeftToRight"
+                       IsTextSelectionEnabled="True"
```
danbelcher-MSFT: As with the history items, recommending `HorizontalAlignment="Right"` here.

src/Calculator/Views/HistoryList.xaml
```diff
@@ -66,13 +66,15 @@
...
                                    AutomationProperties.Name="{x:Bind AccExpression}"
+                                   IsTextSelectionEnabled="True"
                                    Text="{x:Bind Expression}"
                                    TextAlignment="Right"
                                    TextWrapping="Wrap"/>
                         <TextBlock x:Name="resultTextBlock"
                                    Style="{ThemeResource TitleTextBlockStyle}"
                                    FontWeight="SemiBold"
                                    AutomationProperties.Name="{x:Bind AccResult}"
+                                   IsTextSelectionEnabled="True"
```
danbelcher-MSFT: The `exprTextBlock` and `resultTextBlock` both stretch across the width of the ListViewItem.  After text selection is enabled, this makes it extremely difficult to click on the item itself, as the only available hit target is the thin margin between the two textblocks.
image 

We should apply `HorizontalAlignment="Right"` to both textblocks here, so as to minimize their hit target while still enabling text selection when they are explicitly hovered.

src/Calculator/Views/DateCalculator.xaml
```diff
@@ -593,7 +595,8 @@
             <!-- Resulting Date -->
             <TextBlock x:Uid="DateLabel"
                        Grid.Row="8"
-                       Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"/>
+                       Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"
+                       IsTextSelectionEnabled="True"/>
```
danbelcher-MSFT: Was this a mistake?  None of the other labels have text selection enabled so this feels out-of-place.

src/Calculator/Views/MemoryListItem.xaml
rudyhuyn: Good idea!

src/Calculator/Views/DateCalculator.xaml
rudyhuyn: This one is not a standard label, it displays the result of the feature "Add or substract days"

danbelcher-MSFT: I'm not sure I understand why that makes this label special?
In this first image, the red shows selectable text and the blue shows non-selectable labels:
image

Compare with this second image:
image

I don't understand why the 'Date' label is selectable text.  In the first image, 'Difference' was not selectable although it was also displaying the result of the feature.  At least we can say that we're inconsistent between each feature.  Either the label for the result should be selectable in both or not selectable in both.

Second to that, I'm arguing that the 'Date' label should not be selectable.  The *only* action the user can take by selecting the label is to copy the label alone, meaning they can't grab `Date: Wednesday, May 19, 2021` as a single run.  What is special about the label then?  What is helped by allowing it to be selectable where it wasn't before?  Why does this label deserve to be selectable but other labels like 'From' and 'To' are not selectable?

grochocki: > I'm arguing that the 'Date' label should not be selectable. 

I agree with @danbelcher-MSFT. Field labels should not be selectable, only the values/output.

rudyhuyn:
Oh! I was sure this comment was about the another textblock, I totally
agree, this label should not be selectable, it was a mistake

src/Calculator/Views/DateCalculator.xaml
rudyhuyn: I was confused with `DateDiffAllUnitsResultLabel`, I agree, this label should not be selectable!

danbelcher-MSFT: 😅

PR #229, opened 2019-03-09
Optimised images losslessly. 

grischard:
Saved 301 KB out of 9.2 MB. 15.8% per file on average (up to 82.2%)

## Fixes #.

Saves 301 KB (although not all files are included in the build)

- For PNG files: used a combination of Pngcrush, OxiPNG, AdvPNG, PNGOUT and Zopfli
- For GIF files: used Gifsicle

- No changes to behaviour or source code.
- Previewed the changed files to make sure they still work


HowardWolosky:
Interesting, thanks for the submission.  We do have an automated process that @danbelcher-MSFT wrote which runs `TruePNG` on our product images (we may have missed running it on the repo-only images).  I remember that there were other tools that he had tried out and chose _not_ to use for various reasons, but I'm not sure if any were in the list that you used.

Daniel -- Had you tried any of these back when writing our Image Compression automation?

jlaanstra:
LGTM

HowardWolosky:
#273 has issued a similar change, but against fewer files.  It would be good to understand the difference in the optimization result between these two PR's.

grischard:
https://imgbot.net/docs/#compression says that it uses ImageMagick. In my experience, it's fast but doesn't squeeze images as much as other tools.

[ImageOptim](https://github.com/ImageOptim/ImageOptim), which got mentioned in the other thread, is indeed what I used. It compares the output of various lossless tools and picks the smallest. The `PNGOUT` and `Zopfli` tools in particular take a bit more CPU, but are very effective.

danbelcher-MSFT:
> Daniel -- Had you tried any of these back when writing our Image Compression automation?

Nope.  Different tools will optimize in different ways.  We aren't checking image compression any longer in our build pipelines.  If the community wants to crush the images further, that's fine with me.  We'll eventually end up with the most optimized images in the OSS community!

> #273 has issued a similar change, but against fewer files. It would be good to understand the difference in the optimization result between these two PR's.

Let's take this change and then have the author of #273 re-run their optimizations.  If it produces a diff, we can check that in as well.

danbelcher-MSFT:
Thanks, @grischard!

HowardWolosky:
@danbelcher-MSFT - this change will only impact the dev build, not our Store build (see my comment in #273). These files were previously optimized with your TruePNG tooling. If there are legit benefits seen here, we should look into updating our internal tooling.

HowardWolosky:
I edited the comment above, but I'm also posting that update below since there was some misunderstandng by some around the original comment and I want to be sure to make this super clear:

For clarity to anyone else reading the above comment, the Store version is built directly from the _code_ in this repo, however the _icons_ are replaced before the official build starts.  We use a different set of icons in the dev build vs the official build to help distinguish the unofficial vs official builds.  We also don't sign the dev builds.  So, optimizations made to the _icons_ in this repo have no impact to the Store version since these are non-shipping icons.  Instead of necessarily optimizing _these icons_, we should be updating the process that we use to internally to optimize the _shipping icons_, hence my interest more on the _process_ that was used in these two PR's as opposed to this specific _result_.

grischard:
It would be very cool to hear back once you've looked at the internal image pipeline. And I imagine Microsoft also has more software that includes images and gets downloaded often...

danbelcher-MSFT:
Hey @grishard, would you mind sharing what specific commands you ran for each of these tools, if any?  It would help when we want to add this as part of our internal process and I won't need to go learn each of the tools.

grischard:
Hey @danbelcher-MSFT, sorry I missed your question a year ago! I used [ImageOptim](https://github.com/ImageOptim/ImageOptim) which simply brute forces combinations of optimisation tools. It's Mac-only, but [the website](https://imageoptim.com/versions) lists similar tools for Windows.

PR #231, opened 2019-03-09
Fix wrong results after sqrt

### LINKED ISSUE #222: Wrong result for e.g. √(2.25) - 1.5
**Describe the bug**

**Steps To Reproduce**
2. Enter 2.25
3. Press √
4. Subtract 1.5
5. Result is wrong

**Expected behavior**

Result is 0

 - Architecture: X64

**Additional info**

uhliksk:
- Partial result rounding changed from integer to 32 digits after decimal point

- manual

### DIFF
--- a/src/CalcManager/Ratpack/exp.cpp
+++ b/src/CalcManager/Ratpack/exp.cpp
@@ -343,10 +343,31 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
         powratcomp(&originalResult, oneoveryDenom, radix, precision);
                 
         // ##################################
+        // Find number of digits after decimal point
+        // and prepare multiplier for rounding
+        // ##################################
+        PRAT resultMultiplier = nullptr;
+        PRAT resultDigits = nullptr;
+        PNUMBER numResult = RatToNumber(pxPow, radix, precision);
+        if (-numResult->exp < precision)
+        {
+            DUPRAT(resultDigits, longtorat(-numResult->exp / 2));
+        }
+        else
+        {
+            DUPRAT(resultDigits, rat_zero);
+        }
+        DUPRAT(resultMultiplier, rat_ten);
+        powratcomp(&resultMultiplier, resultDigits, radix, precision);
+        destroyrat(resultDigits);
+        destroynum(numResult);
+
+        // ##################################
         // Round the originalResult to roundedResult
         // ##################################
         PRAT roundedResult = nullptr;
         DUPRAT(roundedResult, originalResult);
+        mulrat(&roundedResult, resultMultiplier, precision);
         if (roundedResult->pp->sign == -1)
         {
             subrat(&roundedResult, rat_half, precision);
@@ -356,6 +377,8 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
             addrat(&roundedResult, rat_half, precision);
         }
         intrat(&roundedResult, radix, precision);
+        divrat(&roundedResult, resultMultiplier, precision);
+        destroyrat(resultMultiplier);
 
         // ##################################
         // Take the yDenom power of the roundedResult.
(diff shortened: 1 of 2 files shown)


HowardWolosky:
Thanks @uhliksk.

This PR includes a _behavior change_ to the app, and our team is standardizing on a flow where a _behavior change_ needs to have an associated Issue which will provide the opportunity for discussion and approval before we'll consider merging in the corresponding code change.

jlaanstra:
@uhliksk can you explain the math behind multiplying by 10^32? The comments also need to be updated to reflect the new approach and the 32 should probably be replaced by a constant value based on the mode.

danbelcher-MSFT:
@uhliksk, can you add a unit test that would have failed before the change, but succeeds afterwards?

GeeLaw:
I'm impressed by this attempt. Why is base-10 so special? What about `sqrt(1/9)-1/3`?

If you are serious, do the precise nth-root for the numerator and the denominator separately (when you're working on a fraction), and see if both are integral, in which case you have the actual value represented as a fraction. Otherwise, keep the finite-precision result.

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -345,8 +345,14 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
         // ##################################
         // Round the originalResult to roundedResult
         // ##################################
+        PRAT rat_digits = nullptr;
```
joshkoon: Please add a call to `destroyrat()` for `rat_digits`. Every `PRAT` which is assigned a value other than `nullptr` requires a `destroyrat()` call to avoid a memory leak.

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -345,8 +345,14 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
...
+        DUPRAT(rat_digits, rat_ten);
+        // Round to 32 digits after decimal point
+        powratcomp(&rat_digits, longtorat(32), radix, precision);
```
joshkoon: Please assign the call to `longtorat` to a variable and call `destroyrat()` on the variable after use to avoid a memory leak.

joshkoon:
> @uhliksk, can you add a unit test that would have failed before the change, but succeeds afterwards?

To add to this - we should accompany changes to the way the calculator performs core calculations with a set of unit tests that comprehensively cover the change. This also helps drive one of the areas of focus in the [Roadmap](https://github.com/Microsoft/calculator/blob/master/docs/Roadmap.md) to increase test coverage.

HowardWolosky:
Locking the PR so that conversation on the problem being addressed is discussed in the Issue, so that future PR discussion can be limited to the implementation of the fix.  Will unlock once the associated issue has been approved.

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -342,11 +342,32 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
...
+        // ##################################
+        // Find number of digits after decimal point
+        // and prepare multiplier for rounding
+        // ##################################
+        PRAT resultMultiplier = nullptr;
+        PRAT resultDigits = nullptr;
+        PNUMBER numResult = RatToNumber(pxPow, radix, precision);
+        if (-numResult->exp < precision)
+        {
+            DUPRAT(resultDigits, longtorat(-numResult->exp / 2));
```
joshkoon: `longtorat` returns a `PRAT`, which should be captured by a variable and then destroyed later to avoid a memory leak.

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -342,11 +342,32 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
         DUPRAT(originalResult, pxPow);
         powratcomp(&originalResult, oneoveryDenom, radix, precision);
                 
+        // ##################################
+        // Find number of digits after decimal point
```
jlaanstra: Let's update this comment to: Find the decimal place where we should do rounding.

src/CalcManager/Ratpack/exp.cpp
jlaanstra: -numResult->exp / 2  truncates. Can you add a test where the number of digits after the decimal point is uneven?

joshkoon: The 2 here only works for squares, not other powers. We've unlocked this PR for the purpose of being able to give feedback on the code itself, but let's keep discussing what the correct fix should be on the Issue.

And thanks, @uhliksk, for taking the time to dive into this issue!

HowardWolosky:
Replaced by #297

PR #236, opened 2019-03-09
Fix the project code style, as it is not consistent.

### LINKED ISSUE #202: clang-format spec
Is there a chance to have a `.clang-format` file added to the repository so that code can be automatically formatted?

seyfer:
## Fixes #202
This PR fixes code style for the project files.

### The Problem
Different files in the project use different code style. That is not consistent and leads to harder maintenance of the project.

- Have investigated and determined the most used code style across the given codebase
- ~~Have configured IDE and applied code style to all project files.~~
- Have crafted clang-formatter config.
see https://clang.llvm.org/docs/ClangFormat.html
https://clang.llvm.org/docs/ClangFormatStyleOptions.html
- Some cases were fixed manually

manual/ad-hoc testing, automated testing
- All tests pass as before because these are only code style changes.

### Additional ###
Please review, and let me know if I have any mistake in the code style. In case of any mistake, I will change the configuration and re-apply it to the project.

~~This resource can be used for better rules understanding
https://clangformat.com/~~
btw: outdated


janisozaur:
Can you encapsulate the style in `.clang-format` file, as per #202, and apply it project-wide?

seyfer:
@janisozaur looks interesting. I will investigate and add it.

seyfer:
@janisozaur done.

src/Calculator/Converters/BitFlipAutomationNameConverter.cpp
```diff
@@ -42,18 +42,19 @@ Object ^ BitFlipAutomationNameConverter::Convert(_In_ Object ^ value, TypeName t
         unsigned int binaryLength = 0;
 
         // remove all the characters except 0 and 1 from the array.
-        for
-            each(wchar_t bit in binaryDisplay)
+        // clang-format off
```
janisozaur: What was the problem here? I imagine that's some kind of foreach macro? If that's it, `clang-format` has `ForEachMacros` or it can be changed to C++'s ranged-for

seyfer: @janisozaur the problem was that `clang-format` moved `each` on the next line, as you can see from the diff. And it was not a valid construction. The build was failing here. 
If you see how to fix it better feel free to push commit in my PR. :)
I would check later `ForEachMacros`, thanks.

seyfer: @janisozaur probalby if I add this with space
```
ForEachMacros:   
  - for each
  - foreach
```
to the list, it should work. I will check.

HowardWolosky:
Locking the PR so that conversation on the problem being addressed is discussed in the Issue, so that future PR discussion can be limited to the implementation of the fix.  Will unlock once the associated issue has been approved.

Specific to this PR, we want to delay checking this in until we have a solution that can run clang-format as part of the build process so that we can prevent regressions on this (and this is something we want posthaste).

danbelcher-MSFT:
@seyfer, there's some conflicts to resolve here before we can take the changes.

danbelcher-MSFT:
@seyfer, I think this is a good change to make, in general.  One question I have for you is, how will this format be enforced?  For example, I don't have any Clang/llvm tooling on my machine.  Will the clang-format spec still work?  Here's a scenario:
1. We check this change in.
2. Someone with Visual Studio makes a code change that doesn't match the format specified here. That change is checked in.
3. You make a code change, with an unrelated side-effect of formatting the inconsistent code that was previously approved.  Now those users of Clang bear the responsibility of ensuring formatting and it will pollute their PRs.

Is there a way we can solve this for both MSVC and Clang developers?  For example, can we make this run as part of our CI or even auto format files when they are merged?  This would allow everyone to develop with their own tools and still keep the codebase consistently formatted.

In short, I like the idea for this change, but I think there are some considerations missing here that mean we are setting ourselves up to immediately deviate from the new formatting style.

src/.clang-format
```diff
@@ -0,0 +1,104 @@
...
+PenaltyBreakString: 1000
+PenaltyBreakTemplateDeclaration: 10
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
+ReflowComments: true
+SortIncludes: false
+SortUsingDeclarations: true
+SpaceAfterCStyleCast: false
+SpaceAfterTemplateKeyword: false
```
danbelcher-MSFT:
```suggestion
SpaceAfterTemplateKeyword: true
```

src/.clang-format
```diff
@@ -0,0 +1,104 @@
...
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeCpp11BracedList: true
```
danbelcher-MSFT:
```suggestion
SpaceBeforeCpp11BracedList: false
```

src/CalcManager/CEngine/Number.cpp
```diff
@@ -42,23 +33,14 @@ namespace CalcEngine
         return ret;
     }
 
-    int32_t const& Number::Sign() const
-    {
-        return m_sign;
-    }
+    int32_t const& Number::Sign() const { return m_sign; }
```
danbelcher-MSFT: What setting do we change so that this is reverted?  The previous style is the goal.

danbelcher-MSFT: > this is not really possible to keep the previous style at all.

Can't we use `AllowShortFunctionsOnASingleLine: None`?

danbelcher-MSFT:
@seyfer
> @danbelcher-MSFT if I set `AlwaysBreakAfterDefinitionReturnType: true` then it will be
>```C++
>int32_t const& 
>Number::Sign() const { 
>   return m_sign; 
>}
>```
>is it better?
> The same can be achieved with AlwaysBreakAfterReturnType

Sorry, I'm not sure which comment of mine you are replying to. We want the function signature on a single line.  The goal
```C++
int32_t const& Number::Sign() const
{
    return m_sign;
}
```

src/CalcManager/CEngine/Number.cpp
seyfer: @danbelcher-MSFT I will check it, but I think I have tried it already.

As I tried to explain, such precise adjustments are tough with clang-format. 
Also, I would try custom rules for braces. If it doesn't solve the problem - will mean that it is not possible. Let's see.

seyfer: @danbelcher-MSFT fixed.

danbelcher-MSFT:
@seyfer, I checked out your branch and I'm running into errors when formatting using VisualStudio.  For example, when I format `CalculatorProgrammerRadixOperators.xaml.h` using CTRL+K+D, I get:
image

According to the error, it expected a boolean for this option.  [clangformat.com](https://www.clangformat.com/) shows a boolean but the [Clang 9 documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) shows `Yes`, `No`, `MultiLine` string values.  Can you help me understand why they are different?

The maintainers for this repo use VisualStudio as our only IDE.  We can't accept a change that would break users of VS.  Maybe the format file needs to be updated to align with `clangformat.com`?

seyfer:
@danbelcher-MSFT the problem is that ruleset used for extension is outdated in comparison to the ruleset understandable by `clang-format` utility.
If you want me to make config compatible with VS extension then I need to remove about 5 rules. I did it before, but then returned back for some reason. Will do.

danbelcher-MSFT:
I'm not actually using an extension when I hit this error.  VisualStudio has native support for clang-format:
image

Separately, https://clang.llvm.org/docs/ClangFormat.html links to a VS ClangFormat extension published by LLVM.  I installed the extension, disabled VS native clang format, and attempted to format the same document.  The extension also throws an error:
image

So it seems even the suggested extension doesn't support the latest clang-format schema.  Either way, we'll need to change the format file to support VS.

seyfer:
@danbelcher-MSFT thank you for the descriptive feedback. There definitely was a problem with the ruleset compability. I have manually checked problematic rules and have found, as expected, about 5 rules to exclude.
Also, I checked this online builder - it is very outdated.

Now the ruleset fixed and works from everywhere: VS, extension, console.

janisozaur:
See https://github.com/Microsoft/calculator/pull/405 for discussion regarding `Objective-C` error.

The latest comment in there suggests:

> I recommend removing [CX extensions, e.g. [0]] and verifying everything works (it should).

[0]:

https://github.com/Microsoft/calculator/blob/f6a6aae6e6ce1c485b4b4b02413259649cf2b58f/src/Calculator/Converters/BitFlipAutomationNameConverter.h#L13

seyfer:
@janisozaur I have fixed it by removing `Language` rule, which was too strict. Everything works without errors now.
Proposed adjustments seem valid, but out of the scope of this PR.

danbelcher-MSFT:
@seyfer, thank you for the updates!  I'll give it a try again and leave feedback or approve.

> Proposed adjustments seem valid, but out of the scope of this PR.

Agree with @seyfer here.  While it's safe to remove those attributes, let's focus on formatting for this PR.  Here is an example of an attribute that shouldn't be removed, meaning that we need our formatting rules to properly handle these cases:
https://github.com/Microsoft/calculator/blob/091732aa944d4d7f0085513f40cf91c1007e85e6/src/GraphControl/Control/Grapher.h#L11-L12

danbelcher-MSFT:
@seyfer, I can confirm the updates fixed clang formatting with VisualStudio.  Thanks!

How do you want to handle more changes to the format file?  I have some rules I'd like changed.  Do you want to keep going through review passes on this PR?  Or else we can check in what you have now and make further changes in a separate PR.

seyfer:
@danbelcher-MSFT I think the best is to merge what we have now in this PR and then you can continue changing rules in a separate PR. 
Would be more productive.

src/clang-format-all.sh
```diff
@@ -0,0 +1,68 @@
...
+        return 1
+    fi
+    find-dominating-file "$(realpath "$1"/..)" "$2"
+    return $?
+}
+
+# Run clang-format -i on all of the things
+for dir in "$@"; do
+    pushd "${dir}" &>/dev/null
+    if ! find-dominating-file . .clang-format; then
```
danbelcher-MSFT: Can't we just use the `-style=file` option?
```
Use -style=file to load style configuration from .clang-format file located in one of the parent directories of the source file (or current directory for stdin).
```

src/clang-format-all.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
danbelcher-MSFT: Given that this is primarily a Windows product, we need to have solutions that work for Windows users.  For scripting, this would typically mean PowerShell.  We're asking that you move this script to the following location:
```
\Tools\Scripts\clang-format\clang-format-all.sh
```

After we merge your change, we will add a `clang-format-all.ps1` to the same location.  Then there should be a consistent location for any user that needs this functionality.    You'll have to update your script to correctly format files under the `/src/` directory.  It would also then make sense for the `.clang-format` file to move to the repo root.

src/clang-format-all.sh
seyfer: The idea here was to find the first top config and use it. In this way, it is possible to have different configs for different directories. Since it is not needed and below a static structure proposed - I will remove this.

seyfer: Actually, since `-style=file` is a default, it was just a check, that any style found.

src/clang-format-all.sh
seyfer: @danbelcher-MSFT all was done. :)

danbelcher-MSFT:
Thanks @seyfer.  There's some merge conflicts but I can easily fix those before merging.  I plan on getting this checked in tomorrow.

Issue #237, opened 2019-03-09
Default line ending for files in the project


HowardWolosky:
#225 introduced a change that made `crlf` the default line ending when Visual Studio saves `cpp` and `c` files in this project.

After that change was merged in, [@janisozaur brought up counter-arguments](https://github.com/Microsoft/calculator/pull/225#issuecomment-471222773) for that specific part of the PR.

I'm creating this issue to hear feedback from others in the community on the impact that specifying `crlf` (or any line ending for that matter) has for them.

I'm not tied to keeping the `crlf` change, but if a change is going to be made, I'd like to not have it keep changing because not enough discussion occurred first.

rudyhuyn:
CRLF and UTF-8 are the default values used by Visual Studio (and can't be modified without a .editorconfig file or add-ins) and by almost all the files of this project.

The `.editorconfig` file doesn't change how visual studio behaves by default, but counterbalances if a file has been modified by another app.

Moreover, the current source files already use BOM and CRLF, here is an example:(`EF BB BF` at the beginning and CRLF (`0D 0A`) used for new lines):
![image](https://user-images.githubusercontent.com/1226538/54078420-62a6fe80-427c-11e9-8095-5ea53e17786a.png)

HowardWolosky:
Fixed by #239

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

jlaanstra:
FYI that this project already had an .editorconfig in the root and now there are two that also conflict with each other.

rudyhuyn:
🤦‍♂️

PR #239, opened 2019-03-09
Remove end-of-line from .editorconfig

rudyhuyn:
## Fixes #.
As mentioned here, https://github.com/Microsoft/calculator/issues/237#issuecomment-471230876, `crlf` is already automatically modified by `.gitattribute`, by removing this file we will:
- not change the behavior on Windows (git will replace CR and LF by CRLF)
- prevent issues on Mac/Linux.

There is no rational reasons to keep it, sometimes, less is more. (the only reason I think about is if a file is modified by 2 file editors on the same computer, but the odds are high that they already uses the same EOL).

- remove the `end_of_line = crlf` from .editorconfig.


HowardWolosky:
Thanks Rudy.

Looks like you're not up-to-date.  As opposed to this showing a diff of removing one line, you're making the same 2 file changes again (minus that one line).

Please get your fork up-to-date with master and then push an update to here.

rudyhuyn:
Yes, my upstream was outdated. fixed

rudyhuyn:
So now we have a diff ready if https://github.com/Microsoft/calculator/issues/237 is greenlighted by contributors!

janisozaur:
Thanks

rudyhuyn:
Thank you for your feedback!

Issue #232, opened 2019-03-09
Wrong thousands separation in currency converter


uhliksk:
**Describe the bug**
Big numbers like 100000000000000 are using thousands separator as 100000 000 000 000 instead of 100 000 000 000 000.

**Steps To Reproduce**
1. Go to Currency Converter
2. Type 100000000000000

**Expected behavior**
Thousands separator should be applied on all thousands.

**Screenshots**
![image](https://user-images.githubusercontent.com/37454226/54074472-16b38400-4293-11e9-8583-af7ab48a12da.png)

**Device and Application Information (please complete the following information):**
 - OS Build: 10.0.17763.0
 - Architecture: X64
 - Application Version: 10.1812.10048.0

**Additional context**
Other convesion types are not affected.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

grochocki:
Thanks for the report! I really appreciate your thoroughness when filing this bug. I am not able to repro the issue though. I am on the same OS build with latest code from master. Are there more details you can provide to help us reproduce what you are seeing?

![image](https://user-images.githubusercontent.com/1414423/54077241-a133bd80-426a-11e9-8b99-3562eba2ae6e.png)

rudyhuyn:
I'm able to repro the issue. If everyone agrees that we should fix this issue, I can create a PR.

grochocki:
Approving, since there are multiple confirmed reports. Out of curiosity, which locales are you seeing this in?

rudyhuyn:
All locales using space as thousand delimiter (French for example)

grochocki:
Hmm, that is what I suspected, but when I tried changing my currency delimiter to space, it still seemed to work ok for me. Odd.

rudyhuyn:
Not the "normal" space in fact, the "non-breaking" one (`U+00A0`)

uhliksk:
> Approving, since there are multiple confirmed reports. Out of curiosity, which locales are you seeing this in?

My locale is sk-sk.

PR #242, opened 2019-03-10
Fix formatting issues with CurrencyConverter and some locales

### LINKED ISSUE #240: Currency converter displays an extra space with some locales
**Describe the bug**

An extra space is added at the end of the number with some locales and numbers:

image

**Steps To Reproduce**
- Change your OS language to French or Spanish
- Open the currency converter
- Type 1111111
- See that the 2 currencies display 2 spaces instead of one 

**Expected behavior**
- Display only one space

 - OS Build: 10.0.18348.0
 - Architecture: X64

rudyhuyn:
## Fixes #240 and #232 

The ViewModel wrongly assumed that non-breaking spaces were only used between the value and the symbol. It's not the case of all locales using non-breaking spaces as a thousand delimiter (French for example).

When it was the case, the function only replaced the first thousand delimiter found and kept the extra space at the end of the string, generating 2 issues:

- Extra space at the end: https://github.com/Microsoft/calculator/issues/240 
- Bad formatting of the number: https://github.com/Microsoft/calculator/issues/232

Replace `currencyResult.find(L'\u00a0')` by a regex only removing spaces at the end of the string.

### Screenshots
Before:

image

After:

image

### DIFF
--- a/src/CalcViewModel/UnitConverterViewModel.cpp
+++ b/src/CalcViewModel/UnitConverterViewModel.cpp
@@ -53,6 +53,7 @@ constexpr size_t SELECTED_TARGET_UNIT = 2;
 
 // x millisecond delay before we consider conversion to be final
 constexpr unsigned int CONVERSION_FINALIZED_DELAY_IN_MS = 1000;
+const wregex regexTrimSpacesStart = wregex(L"^\\s+");
 const wregex regexTrimSpacesEnd = wregex(L"\\s+$");
 
 namespace CalculatorApp::ViewModel
@@ -350,9 +351,13 @@ String^ UnitConverterViewModel::ConvertToLocalizedString(const std::wstring& str
             {
                 currencyResult.erase(pos, currencyCode.length());
                 std::wsmatch sm;
+                if (regex_search(currencyResult, sm, regexTrimSpacesStart))
+                {
+                    currencyResult.erase(sm.prefix().length(), sm.length());
+                }
+
                 if (regex_search(currencyResult, sm, regexTrimSpacesEnd))
                 {
-                    currencyResult.erase(pos, 1);
                     currencyResult.erase(sm.prefix().length(), sm.length());
                 }
             }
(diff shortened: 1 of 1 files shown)


src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -53,6 +53,8 @@ constexpr size_t SELECTED_TARGET_UNIT = 2;
 
 // x millisecond delay before we consider conversion to be final
 constexpr unsigned int CONVERSION_FINALIZED_DELAY_IN_MS = 1000;
+const wregex regexTrimSpacesStart = wregex(L"^\\s+");
```
jefgen: A few thoughts: I think what is considered a whitespace character (the `\s`) by the std::wregex library depends on the current locale setting. So there might be some odd cases where `\u00a0` isn't consider whitespace character by the locale, yet shows up in the currency result.

rudyhuyn: I had a doubt too, but it should be fine. If no flags are passed in the constructor, the wregex will use ECMA and based on the spec, `\s` will always match `\u00a0`.

Spec here:
http://ecma-international.org/ecma-262/5.1/#sec-7.2

jefgen: Nice! Thanks for the reply.

I suppose the problem might be if locales start using Narrow No-break Space (U+202F) at some point.
See the "Migration" section in CLDR v34 here: http://cldr.unicode.org/index/downloads/cldr-34

However, we currently override this in the CLDR data used by NLS in Windows, so that won't be an issue for now...

jefgen: Actually, upon second reading of the ECMAScript spec:

> ECMAScript implementations must recognise all of the white space characters defined in Unicode 3.0. Later editions of the Unicode Standard may define other white space characters. ECMAScript implementations may recognise white space characters from later editions of the Unicode Standard.

U+202F `NARROW NO-BREAK SPACE` was added in Unicode 3.0.0, and is a Space (Zs) category. So I guess this *should* work for U+202F as well.

jefgen:
Thanks @rudyhuyn :+1:

src/CalcViewModel/UnitConverterViewModel.cpp
HowardWolosky: Thanks for this thorough and thoughtful review, @jefgen.  Much appreciated.

HowardWolosky:
Closing and re-opening to try and fix the PR build.

HowardWolosky:
@mcooley - Something odd is going on with the CI pipeline here.  Any chance you can take a look?

PR #246, opened 2019-03-10
Fix auto-scaling of CalculationResult when the current locale displays currency symbols on the right

### LINKED ISSUE #245: The FontSize of CalculationResult doesn't auto-scale with some locales.
The `CalculationResult` control has a built-in feature to automatically modify the FontSize of the control to make sure the content is always fully visible.

This behavior works perfectly in English (US):
image

But not with some other languages (example with French):
75602968-ACE9-48C0-8CDB-5756EAA6AB18

**Steps To Reproduce**
- Change the system language
- Select one of the converter (bug visible with all of them)
- Type a long number.

**Expected behavior**

The control should modify the FontSize when the number is too large

 - OS Build: 10.0.18348.0
 - Architecture: X64

rudyhuyn:
In order to move the currency symbol to the right, the property `Grid.Column` of the textblock and CalculationResult were switched. The issue is that one of the column width in `Auto` while the other is `*`. CalculationResult were then no more able to correctly calculate the space available and modify the font-size to fit correctly.

Instead of switching columns **and** modifying the columns width, I simply added a third column and used it for the currency symbol without moving the `CalculationResult`.

Manually tested in english (symbol on the left), spanish (symbols on the right) and french (symbols on the right)


sanderl:
Hi Rudy,
Thanks for finding this issue and getting a PR out. I built your change to see what the diffence would be with your change. It looks like the currency symbol is now on the right edge of the grid at all times. I agree this fixes the issue where text was being cut off when the value was very large. However, now with smaller numbers, the currency symbol is so far to the right it feels disjointed from the value. 
RightSideCurrencySymbol

I've been thinking through how to get the currency symbol closer to the value, however I haven't come up with a great solution yet. Maybe we need a MaxWidth set on the value and adjusted it to fit the window size in the visual states. We could also potentially have 2 controls one for currency symbol on the left and one for currency symbol on the right. We could just toggle the visibility in the code behind.

rudyhuyn:
I'm checking that! (I'm in Building 33 by the way!)

rudyhuyn:
Fixed!

> how to get the currency symbol closer to the value
I changed that too:

**Before/After**
image

sanderl:
It looks great now! Thanks for making that update and fixing this bug.

PR #200, opened 2019-03-08
Improving code style : verbosity, indentation levels

garsaud:
Fixing some nested if() statements and reducing indentation levels.

Making some sections less verbose, e.g:
```cpp
if (a == 1)
{
    b = true;
}
else
{
    b = false;
}
```
↓
```cpp
b = (a == 1)
```

Changes are best viewed ignoring whitespaces:

PR #253, opened 2019-03-10
Propose code-cleanups#2

Maharramoff:
- Remove redundancy 
- Simplify if statements

- Manual

### DIFF
--- a/src/CalcManager/CalculatorHistory.cpp
+++ b/src/CalcManager/CalculatorHistory.cpp
@@ -41,7 +41,8 @@ unsigned int CalculatorHistory::AddItem(_In_ shared_ptr<HISTORYITEM> const &spHi
     }
 
     m_historyItems.push_back(spHistoryItem);
-    return static_cast<unsigned>(m_historyItems.size() - 1);
+unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
+return lastIndex;
 }
 
 bool CalculatorHistory::RemoveItem(_In_ unsigned int uIdx)
--- a/src/CalcManager/CalculatorHistory.cpp
+++ b/src/CalcManager/CalculatorHistory.cpp
@@ -35,16 +35,13 @@ unsigned int CalculatorHistory::AddToHistory(_In_ shared_ptr<CalculatorVector <p
 
 unsigned int CalculatorHistory::AddItem(_In_ shared_ptr<HISTORYITEM> const &spHistoryItem)
 {
-    int lastIndex;
-
     if (m_historyItems.size() >= m_maxHistorySize)
     {
         m_historyItems.erase(m_historyItems.begin());
     }
 
     m_historyItems.push_back(spHistoryItem);
-    lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
-    return lastIndex;
+    return static_cast<unsigned>(m_historyItems.size() - 1);
 }
 
 bool CalculatorHistory::RemoveItem(_In_ unsigned int uIdx)
--- a/src/CalcManager/CalculatorHistory.cpp
+++ b/src/CalcManager/CalculatorHistory.cpp
@@ -41,7 +41,8 @@ unsigned int CalculatorHistory::AddItem(_In_ shared_ptr<HISTORYITEM> const &spHi
     }
 
     m_historyItems.push_back(spHistoryItem);
-unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
+    unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
+    return lastIndex;
 return lastIndex;
 }
 
--- a/src/CalcManager/CalculatorHistory.cpp
+++ b/src/CalcManager/CalculatorHistory.cpp
@@ -43,7 +43,6 @@ unsigned int CalculatorHistory::AddItem(_In_ shared_ptr<HISTORYITEM> const &spHi
     m_historyItems.push_back(spHistoryItem);
     unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
     return lastIndex;
-return lastIndex;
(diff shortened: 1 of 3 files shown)


HowardWolosky:
Same unitconverter change can be found in #200

src/CalcManager/CalculatorHistory.cpp
```diff
@@ -35,16 +35,13 @@ unsigned int CalculatorHistory::AddToHistory(_In_ shared_ptr<CalculatorVector <p
...
-
     if (m_historyItems.size() >= m_maxHistorySize)
     {
         m_historyItems.erase(m_historyItems.begin());
     }
 
     m_historyItems.push_back(spHistoryItem);
-    lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
-    return lastIndex;
+    return static_cast<unsigned>(m_historyItems.size() - 1);
```
HowardWolosky: On this one, I don't fully agree with the change.  Having the named variable there gives "documentation through code" that the _intention_ is to be returning the last index.  The change that _would_ make sense here would be to combine the variable declaration and assignment together:

```
int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
return lastIndex;
```

janisozaur: @HowardWolosky you might want to check a feature of GitHub that would make your suggestion applicable with one click via webui: https://help.github.com/en/articles/incorporating-feedback-in-your-pull-request#applying-a-suggested-change

HowardWolosky: Thanks for the suggestion, @janisozaur  -- It looks like @Maharramoff hasn't configured this PR to accept edits from maintainers, unless I'm misunderstanding how the feature works.

Maharramoff: > Thanks for the suggestion, @janisozaur -- It looks like @Maharramoff hasn't configured this PR to accept edits from maintainers, unless I'm misunderstanding how the feature works.

@HowardWolosky  I already did it. 
G5_l7zM8Te_VFwGRV_KzAg

janisozaur: @HowardWolosky sorry, I linked to the receiving part of GitHub's review suggestions, I should've rather linked to this instead: 
[suggestion-block](https://help.github.com/en/articles/commenting-on-a-pull-request) and you can see a sample of how it would like below:
```suggestion
    int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
    return lastIndex;
```

Maharramoff: > On this one, I don't fully agree with the change. Having the named variable there gives "documentation through code" that the _intention_ is to be returning the last index. The change that _would_ make sense here would be to combine the variable declaration and assignment together:
> 
> ```
> int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
> return lastIndex;
> ```

Thanks, you are right. Then, I have one more suggestion before editing commit. If the return type of the function is **unsigned int**, shouldn't we declare the local variable as **unsigned int** also ?

```
unsigned int CalculatorHistory::AddItem(_In_ shared_ptr<HISTORYITEM> const &spHistoryItem)
{
    int lastIndex;
```

HowardWolosky: That indeed is a great catch.  `lastIndex` _should_ be an `unsigned int`.  Given that `AddItem` is always ensuring that there is something in `m_historyItems`, the value of `m_historyItems.size()` should always be `1` or greater, meaning that we're guaranteed that our index calculation will never go negative.  Given that, yes, it should be an `unsigned int`.

Using the new tip from @janisozaur:

```suggestion
unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
return lastIndex;
```

src/CalcManager/CalculatorHistory.cpp
```diff
@@ -35,16 +35,14 @@ unsigned int CalculatorHistory::AddToHistory(_In_ shared_ptr<CalculatorVector <p
...
-
     if (m_historyItems.size() >= m_maxHistorySize)
     {
         m_historyItems.erase(m_historyItems.begin());
     }
 
     m_historyItems.push_back(spHistoryItem);
-    lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
-    return lastIndex;
+unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
```
HowardWolosky: ok...still learning how that suggested change feature works.  Looks like the indentation got lost when you accepted the change.  Let's try this again:
```suggestion
    unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
    return lastIndex;
```

janisozaur: Sadly, a limitation of GitHub's UI is ability to only suggest a change to single line. You can suggest inserting multiple lines, but it will always remove just one. If your suggestion were to be merged, there would be two return statements. An alternative would be doing two suggestions, but they will create two commits and thus two CI runs

src/CalcManager/CalculatorHistory.cpp
```diff
@@ -41,7 +41,8 @@ unsigned int CalculatorHistory::AddItem(_In_ shared_ptr<HISTORYITEM> const &spHi
     }
 
     m_historyItems.push_back(spHistoryItem);
-unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
+    unsigned int lastIndex = static_cast<unsigned>(m_historyItems.size() - 1);
+    return lastIndex;
 return lastIndex;
```
HowardWolosky: As @janisozaur pointed out, the `suggestion` feature at the moment introduced _two_ return statements.
For now, do you think you can just fix this manually?

Maharramoff: Sure

PR #256, opened 2019-03-11
Currency rate: always display minimum 4 significant decimals or 2 trailing zeros if the number is integer

### LINKED ISSUE #252: Incorrect currency ratio equality when low valued unit converted to high valued
**Describe the bug**
Converter rounds/truncates rate to 4 decimal places, therefore currency ratio of low valued units such iranian rial or indonesian rupiah compared to highly valued units like US dollars displayed incorrectly.

**Steps To Reproduce**
1. Launch the calculator
2. Open the currency converter
3. Select Iranian rial as first unit and US dollar as second.
4. See today's exchange rate.

**Expected behavior**
Ideally ratio should be rounded appropriately, not too much and not too little. Basically converters round it to 6 decimals. E.g **1** IRR = 0.000024 USD

**Screenshots**
Снимок экрана (7)_LI

 - OS Build: 17763
 - Architecture: X64

**Additional context**
Possible solution in:
- https://github.com/Microsoft/calculator/blob/0f615faff93dc11101f6b730ce46c69f5390b94c/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp#L42
- https://github.com/Microsoft/calculator/blob/0f615faff93dc11101f6b730ce46c69f5390b94c/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp#L281

rudyhuyn:
## Fixes #252

_Converter rounds rate to 4 decimal places, therefore currency ratio of low valued units such iranian rial or indonesian rupiah compared to highly valued units like US dollars displayed incorrectly._

image
Left: before, Right: after.

Modify how we round the currency rate. Instead of always cropping the number after 4 decimals, we compute how many decimals are needed to display the currency rate with minimum 2 meaningful decimals.

Number of decimals to display:  ` Max(4, (int)-Log10(currencyRate) + 2)` 

Replace also `static_cast<int>(ratio * static_cast<int>(scale)) / scale;` to support bigger scales.
 
- Unit tests added
- Manual tests in English and French


jlaanstra:
Looks like we used to do 4 and now arbitrarily do two for small numbers. I think we should be consistent and either always use 4 or 2 or display all digits.

rudyhuyn:
> Looks like we used to do 4 and now arbitrarily do 2 for small numbers. I think we should be consistent and either always use 4 or 2 or display all digits.

Not really. 

This code always displays 4 decimals, but if the number looks like `0.0003` or ` 0.0000`, we add more digits to display minimum 2 meaningful digits: 

| Source  | Current (always 4 decimals) | Diff (from 4 to X, in order to always display minimum 2 meanful digits) |
| ------------- | ------------- |------------ |
| 12  | 12.0000  | 12.0000 |
| 12.232 | 12.2320 | 12.2320 |
| 12.232124| 12.2321 | 12.2321 |
| 0.1231 | 0.1231 | 0.1231 |
| 0.0231 | 0.0231 | 0.0231 |
| 0.005311 | 0.0053 | 0.0053 |
| 0.00014311 | **0.0001**  | **0.00014** |
| 0.000064314 | **0.0000** | **0.000064** |
| 0.0000000064314 | **0.0000** | **0.0000000064**|

jlaanstra:
@rudyhuyn my point would be that 2 meaningful digits is random. I would say that if we decide to round, we should do so consistently, and always use the same amount of meaningful digits.

rudyhuyn:
> @rudyhuyn my point would be that 2 meaningful digits is random. I would say that if we decide to round, we should do so consistently, and always use the same amount of meaningful digits.

Got it! It's only a constant to change, I will do it later!

rudyhuyn:
@jlaanstra If the value is exactly 0.00001, should we display `0.00001` or `0.00001000`?

jlaanstra:
@rudyhuyn if you decide to display 4 meaningful digits, then 0.00001000 would be correct.

rudyhuyn:
`12.23` isn't enough precise, especially for a currency rate.

Diff updated to manage 4 meaningful digits! Thanks for your feedback!

I'm not sure `0.0001000` will be helpful, but if it's something you want to do that, `DecimalFormatter` will need a refactoring.

grochocki:
> I'm not sure 0.0001000 will be helpful, but if it's something you want to do that, DecimalFormatter will need a refactoring.

We should not show trailing zeros. I believe precision with zeros matters more for scientific applications, but for currency, we should trim them.

rudyhuyn:
> > I'm not sure 0.0001000 will be helpful, but if it's something you want to do that, DecimalFormatter will need a refactoring.
> 
> We should not show trailing zeros. I believe precision with zeros matters more for scientific applications, but for currency, we should trim them.

the current diff would work in this case!

mdrejhon:
### I Am An Ex-Bank Employee here.  APPROVE. 

Consistency is not the right tool for this particular job:

1. The big numbers should be **Easy Info** -- e.g. "0.00" or the most accepted common number of digits (e.g. 2 digits for most currencies) -- whatever cashiers need to do to their job.  Using Calc on a Surface at a charity booth.  Cashiers don't need to know more digits than needed.

2. The tiny number should be **Details Info** with sufficient enough number of digits to display at least 2 significant digits, even if it requires a large number of digits after the decimal. e.g. "0.000024" -- basically it tells you 1 unit of a currency that it's worth something but less than a penny -- important context that is not quickly answered by "0.00" nor "0.000" since some users get confused by that.

I thereby agree with @rudyhuyn approach.  

Much more logical approach matching real-world "need to know" information use cases.   

Going back to consistency in this particular case, **is a regression**, full stop.  Even keeping this issue untouched is potentially a superior move (on average, user ecosystem wide).

The problem is that this original tweak away from consistency (for an excellent good reason) was incomplete because hardcoded-4 does not reveal enough digits in some currency conversions like this one.  To fix this, the @rudyhuyn suggestion solves this closer to the intent of "good reason" away from consistency.

Thus, easy info shall stay hardcoded-2 as **cash register standard** or **accounting standard** -- and details shall go "a few significant digits visible" as **currency infoboard standard** regardless of how many digits is needed.  

*Note: I've worked as an employee for a Canadian bank, working as an app developer. [LinkedIn Proof](https://www.linkedin.com/in/mark-rejhon). Mic drop.* 

### I Am An Ex-Bank Employee here.  APPROVE.

HowardWolosky:
@grochocki / @jlaanstra - @mdrejhon has a **super** compelling argument here, and has fully convinced me.  I think that if we're going to make a change here, let's do it once and do it right.  I love @mdrejhon's suggestion that we keep the big number at two-significant digits, and the detail number to be as many as necessary to have two significant digits (with a minimum of two digits).

rudyhuyn:
>  we keep the big number at two-significant digits, and the detail number to be as many as necessary to have two significant digits (with a minimum of two digits).

The current version of the application up to 4 decimals:
image

My initial PR was: display up to 4 decimals (with a minimum of two significant digits)

Following feedback, I changed it to: display minimum of 4 significant digits.

It's very easy to change the behavior of this code using these 2 contants:
```
static constexpr int FORMATTER_RATE_MIN_DIGIT_COUNT = 4;
static constexpr int FORMATTER_RATE_MIN_SIGNIFICANT_DIGITS = 4;
```

It looks like the behavior you are describing is:

```
static constexpr int FORMATTER_RATE_MIN_DIGIT_COUNT = 2;
static constexpr int FORMATTER_RATE_MIN_SIGNIFICANT_DIGITS = 2;
```
@HowardWolosky and @mdrejhon, is it what you want?

| value |  current version of the app | New version |
| ------------- | ------------- | ------------ |
| 12  | 12  | 12 |
| 12.232 | 12.232 | 12.23 |
| 12.232124 | 12.2321 | 12.23 |
| 0.1231 | 0.1231 | 0.12 |
| 0.0231 | 0.0231 | 0.023 |
| 0.005311 | 0.0053 | 0.0053 |
| 0.00014311 | 0.0001  | 0.00014 |
| 0.000064314 | 0.0000 | 0.000064 |
| 0.0000000064314 | 0.0000 | 0.0000000064 |

grochocki:
@rudyhuyn I don't think we want to lose precision for "big" numbers in currency converter in the same way that @mdrejhon suggests is ok for the cash register scenario. I would expect:

| Value |  Desired |
| ------------- | ------------ |
| 12  | 12 |
| 12.2321 | 12.2321 |
| 12.232 | 12.232 |
| 12.0000007132 | 12 |
| 12.232124 | 12.2321 |
| 0.1231 | 0.1231 |
| 0.0231 | 0.0231 |
| 0.005311 | 0.0053 |
| 0.00014311 | 0.00014 |
| 0.000064314 | 0.000064 |
| 0.0000000064314 | 0.0000000064 |

rudyhuyn:
> @rudyhuyn I don't think we want to lose precision for "big" numbers in currency converter in the same way that @mdrejhon suggests is ok for the cash register scenario. I would expect:
> 
> Value	Desired
> 12	12
> 12.2321	12.2321
> 12.232	12.232
> 12.0000007132	12
> 12.232124	12.2321
> 0.1231	0.1231
> 0.0231	0.0231
> 0.005311	0.0053
> 0.00014311	0.00014
> 0.000064314	0.000064
> 0.0000000064314	0.0000000064

OK! It will be easy, it was the behavior of my first commit, let me rollback the last change!

mdrejhon:
This is sometimes a contentious debate even inside banks, when I worked within banks -- but it ended up in a consensus similar to

* Select a good Exchange Quote profider
* External Internet datafeeds often had the correct number quotes
* External Internet datafeeds often had zeros pre-stripped
* Some datafeeds kept X number of significant digits
* Sometime datafeeds had weird effects like 12.3523000000008
* So it was necessary to choose an arbitrary number of significant digits, such as MAX("number of significant digits" versus "total number of digits to left of decimal") combined with a minimum number of decimal digits (such as 4).  This certainly varies from bank to bank as the consensus is not unamious.
* However, it is quite fairly unamious consent to strip things like 12.000000000072 to "12" 

So possible general guidelines sequence is as follows:

value	current version of the app	Big number	Small Number
12	12	12	12
12.2321	12.232	12.23	12.232
12.232	12.232	12.23	12.232
12.0000007132	12	12	12
12.232124	12.2321	12.23	12.2321
0.1231	0.1231	0.12	0.1231	
0.0231	0.0231	0.02	0.0231
0.005311	0.0053	0.00	0.005311
0.00014311	0.0001	0.00	0.0001431
0.000064314	0.0000	0.00	0.00006431
0.0000000064314	0.0000	0.00	0.000000006431

mdrejhon:
The details is sometimes a debate even within banks, investments, and sometimes (For the small number, currency infoboards) we just displayed exactly what the quote provider gave for a FX quote.   And reformatted to standard 2 digit with rounding for the transaction (accounting, cash register, bank statement, etc).  But sometimes data feeds had odd data like 12.00000000000000000713 which should be trunctated.   If your Quote provider guarantees that doesn't happen, just display the quote feed directly for the small details number.

Thusly,

* Preserve all digits to left of decimal
* Keep **MAX(**count of non-decimal digits **,**4**)** total digits
* For accounting format, strip/add trailing zeros to always show two decimal digits (for most currency regions; querying the OS will tell you the standard)

Entry | Claimed-Current | Small-Number | Big-Number
--|--|--|--
12|12|12.00|12.00
12.2321|12.232|12.23|12.232
12.232|12.232|12.23|12.232
12.0000007132|12|12.00|12.00
12.232124|12.2321|12.23|12.2321
0.1231|0.1231|0.12|0.1231|
0.0231|0.0231|0.02|0.0231
0.005311|0.0053|0.00|0.005311
0.00014311|0.0001|0.00|0.0001431
0.000064314|0.0000|0.00|0.00006431
0.0000000064314|0.0000|0.00|0.000000006431

HowardWolosky:
cc @grochocki 

Before we waste any more of @rudyhuyn's time here with changes, it seems that we need to have a bit more discussion on this topic to make sure that we're making the right change.  I'm going to _lock_ this PR for the time being so that we don't have the conversation split between the PR and the Issue.  Once we have what appears to be a mutually agreed upon desired outcome, I'll unlock this so that Rudy can pursue the appropriate fix.

grochocki:
Unlocking after approving #252. See details there for more information on direction for this.

rudyhuyn:
Modified to use the specified format. Ready for review

PR #259, opened 2019-03-11
Use the current locale to sort country names

### LINKED ISSUE #258: Countries in the currency converter aren't correctly sorted (because of accents and diacritics)
**Describe the bug**

Countries with an accent in their names aren't correctly sorted. (For example "États-Unis" (USA in French) is at the end of the list, making it harder to find.

97E3BE94-3DF2-4ADC-AE15-25D520430F4D

**Steps To Reproduce**
- Change the language of the OS to French
- Open the currency converter
- Try to find "États-Unis" in the list

**Expected behavior**
The application must take into account accents and diacritics when sorting countries

 - OS Build: 10.0.18348.0
 - Architecture: X64

rudyhuyn:
## Fixes #258

> Countries with an accent in their names aren't correctly sorted. (For example "États-Unis" (USA in French) is at the end of the list, making it harder to find.

(see also the gif available here: #258)

### Screenshot of the fix
image

`std::wstring::operator<` does a very basic comparison of strings based on the value (int) of the characters without taking into account accents, diacritics, etc...

To fix this issue, we will use std::collate with the current locale, it will not only ignore accents and diacritics but use locale rules to sort the string (for example `ch` in Czech is between ` h`  and `i`)

- tested in french, spanish and english.


src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -547,9 +548,15 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String^ rawJson, _Inout_ vector
     }
 
     // TODO - MSFT 8533667: this sort will be replaced by a WinRT call to sort localized strings
```
rudyhuyn: Should we keep that?

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -547,9 +548,15 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String^ rawJson, _Inout_ vector
...
-    sort(begin(staticData), end(staticData), [](CurrencyStaticData unit1, CurrencyStaticData unit2)
+
+    auto loc = std::locale("");  //Use the user-preferred locale to sort country names
```
jefgen: So, I'm not sure if this would pick on "alternate sorts" that the user might have selected in the Settings/Regional Control Panel.

rudyhuyn: If it was the case, it would be even better, no?

jefgen: No, it would mean that the user's selection in the Settings was being disregarded, which is a bad thing.

Though on second though if that was the case, then it feels like it should be bug on the library or CRT.

rudyhuyn: > No, it would mean that the user's selection in the Settings was being disregarded, which is a bad thing.

We are on the same page in fact, my answer was perhaps ambiguous, "if the applications uses the alternate sort" selected by the user, it would be even better" (Not tested but it's probably already the case.)

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
joseartrivera: I think you can go ahead and remove it.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -9,6 +9,7 @@
 #include "Common/LocalizationSettings.h"
 #include "Common/TraceLogger.h"
 #include "UnitConverterDataConstants.h"
+#include <locale>
```
joseartrivera: I believe the style of the project is to move standard headers into the pch file. +@danbelcher-MSFT to double check me here.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
joseartrivera: You can drop the std:: prefix here and in the rest of the file.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -547,9 +548,15 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String^ rawJson, _Inout_ vector
...
-    sort(begin(staticData), end(staticData), [](CurrencyStaticData unit1, CurrencyStaticData unit2)
+
+    auto loc = std::locale("");  //Use the user-preferred locale to sort country names
+    const std::collate<wchar_t>& coll = std::use_facet<std::collate<wchar_t> >(loc);
+    sort(begin(staticData), end(staticData), [&coll](CurrencyStaticData unit1, CurrencyStaticData unit2)
     {
-        return unit1.countryName < unit2.countryName;
+        auto country1 = unit1.countryName.data();
+        auto country2 = unit2.countryName.data();
+        return coll.compare(country1, country1 + unit1.countryName.length(),
```
joseartrivera: Any reason you can't check the .length() with country1.length()?

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -547,9 +548,15 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String^ rawJson, _Inout_ vector
     }
 
     // TODO - MSFT 8533667: this sort will be replaced by a WinRT call to sort localized strings
-    sort(begin(staticData), end(staticData), [](CurrencyStaticData unit1, CurrencyStaticData unit2)
+
+    auto loc = std::locale("");  //Use the user-preferred locale to sort country names
+    const std::collate<wchar_t>& coll = std::use_facet<std::collate<wchar_t> >(loc);
```
joseartrivera: nit: Remove blank space after the second bracket <wchar_t> >

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
jefgen: Taking a look at PR #289 , I wonder if this should possibly be changed to get the User Locale from the  `LocalizationService` instead.
This would start working towards only having a single place in the code that references locale, which would mean that mocking (for tests) in the future would be easier.

jefgen: Additionally, the `LocalizationService` could cache the locale and collator object for perf as well.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
danbelcher-MSFT: Yes, please move to pch

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
danbelcher-MSFT: Related to this, maybe `LocalizationService` should have a `Sort` function which does a localized sort of a sequence of strings.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
rudyhuyn: `wcslen` is O(n) while wstring::length is O(1)

src/CalcViewModel/Common/LocalizationService.h
```diff
@@ -39,6 +39,21 @@ namespace CalculatorApp { namespace Common
         Windows::UI::Text::FontWeight GetFontWeightOverride();
         double GetFontScaleFactorOverride(LanguageFontType fontType);
 
+        void Sort(std::vector<Platform::String^>& source);
+
+        template<typename T>
+        void Sort(std::vector<T>& source, std::function<Platform::String^(T)> func)
```
danbelcher-MSFT: Interested in making this more generic?  For example, you could accept a begin and end iterator, like STL algorithms, and then sort over that sequence.  We can probably also remove the dependency on Platform::String^, but that requires doing more work at the callsite.

src/CalcViewModel/Common/LocalizationService.h
```diff
@@ -39,6 +39,21 @@ namespace CalculatorApp { namespace Common
         Windows::UI::Text::FontWeight GetFontWeightOverride();
         double GetFontScaleFactorOverride(LanguageFontType fontType);
 
+        void Sort(std::vector<Platform::String^>& source);
```
danbelcher-MSFT: Unused?

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -547,17 +552,11 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String^ rawJson, _Inout_ vector
         };
     }
 
-    // TODO - MSFT 8533667: this sort will be replaced by a WinRT call to sort localized strings
+    auto sortCurrencyNames = [](UCM::CurrencyStaticData s) {
```
danbelcher-MSFT: `CurrencyStaticData` is a struct.  Let's capture by const-ref here.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -547,17 +552,11 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String^ rawJson, _Inout_ vector
...
+        return ref new Platform::String(s.countryName.c_str());
```
danbelcher-MSFT: Don't need fully qualified names in cpps

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -505,6 +504,12 @@ bool CurrencyDataLoader::TryParseWebResponses(
         && TryParseAllRatiosData(allRatiosJson, allRatiosData);
 }
 
+Platform::String ^ te(const UCM::CurrencyStaticData& s)
```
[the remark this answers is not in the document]
rudyhuyn: Oops, it's what happens when you write a diff at 2am :D

src/CalcViewModel/Common/LocalizationService.h
rudyhuyn: Yes, but it's probably better to provide a method to directly sorts strings without using one with a lambda (better performance). If it was an internal project, I would agree, but because it's an open-source project, it's probably better to provide some extra functions/helpers, especially in a class like `LocalizationService`, I see this class as a toolbox, it will probably help another developer in a future.

rudyhuyn:
I just solved conflicts with `Master`, so ready to merge if you are ok with this PR!

PR #263, opened 2019-03-11
Added two missing spaces

mikeshatch:
## Fixes #.
Missing spaces

- Added missing spaces that weren't visible due to manual line breaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ For information regarding Windows Calculator plans and release schedule, please 
 ## Data / Telemetry
 This project collects usage data and sends it to Microsoft to help improve our products and services.
 Read our [privacy statement](https://go.microsoft.com/fwlink/?LinkId=521839) to learn more.
-Telemetry is disabled in development builds by default, and can be enabled with the `SEND_TELEMETRY`
+Telemetry is disabled in development builds by default, and can be enabled with the `SEND_TELEMETRY` 
 build flag.
 
 ## Reporting Security Issues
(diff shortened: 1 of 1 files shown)


HowardWolosky:
Not seeing any problem with the file being rendered as-is.  There's also no consistency with why you chose _these_ two lines to add trailing spaces to, but not others.  If you can point to specific MarkDown documentation that indicates that split-line paragraphs should have trailing spaces on those split lines, then I will certainly reconsider, but I'm not aware of any [documentation](https://daringfireball.net/projects/markdown/syntax) that indicates that.

mikeshatch:
Hi Howard. Interestingly, it's not showing the missing space now for the first suggested correction. However, the second one still shows. Did you try saving it as a pdf? If you save it as a pdf you should see there is no space between "the" and "Security". Unless it's just buggy on my end.

HowardWolosky:
Saving it as a PDF from GitHub? Or another viewer?  For instance, MarkdownPad explicitly renders a space there.  On GitHub, I reviewed the rendered HTML, and there is no explicit space that it inserts, but it still renders fine.  And if this was indeed an issue, there are _other_ lines in the README (and the other MarkDown files in this repo) that would need this same fix.

Again, I'm not inherently against making this change, but it's not clear to me why this change is just fixing _these two specific instances_, and looking more broadly, if this is an issue with our documentation, or rather the MarkDown renderer that you're viewing the content in.

mikeshatch:
Saving it as a pdf via Google Docs. I didn't actually check the _other_ README files. Just this one. It's good this happened sooner rather than later because I think I'll avoid GitHub contributions of this nature in the future. Thank you.

HowardWolosky:
Thanks Mike.  As I've mentioned, we want these docs to be as accessible and correct as possible...I'm simply unable to reproduce the specific problem that you're having.  I'd encourage you to open up a bug if you can come up with some clear repro steps for the problem, as it would better help me to understand the scope and impact of the issue.

PR #269, opened 2019-03-12
CopyPasteManagerTest fails to test all values

rudyhuyn:
CopyPasteManager unit tests were not fully run, the first item of arrays were never tested. (Luckily, the not-tested values were ok).

`while(--size)`
the following code won't run the code in the loop with `size==0`.

Instead we should use:
`while(--size >= 0)`

**Before** (first value not tested):
image

**After** (the test fails as expected):
image

### DIFF
--- a/src/CalculatorUnitTests/CopyPasteManagerTest.cpp
+++ b/src/CalculatorUnitTests/CopyPasteManagerTest.cpp
@@ -25,30 +25,17 @@ namespace CalculatorUnitTests
 
 #define ASSERT_POSITIVE_TESTCASES(func, dataSet) \
 {\
-    int size = sizeof(dataSet)/sizeof(*dataSet);\
-    while(--size >= 0)\
+    for(auto data : dataSet)\
     {\
-        VERIFY_ARE_EQUAL(func(dataSet[size]), dataSet[size]);\
+        VERIFY_ARE_EQUAL(func(data), data);\
     }\
 }
 
 #define ASSERT_NEGATIVE_TESTCASES(func, dataSet) \
 {\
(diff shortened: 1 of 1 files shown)


rudyhuyn:
Do we need to open a Issue ticket when it doesn't concern the app itself?

rudyhuyn:
> Woah. Nice catch, Rudy.
> 
> I _detest_ pre-increment/decrement. In my experience, more times than not, their usage leads to bugs. Whenever I see pre-increment/decrement, Case in point. I have to slow down _significantly_ when reading the code to understand why that part of the code is so special.
> 
> Your fix is correct. I'd much prefer it though if you simply switched it to being a `for` loop with an inline-declared `int i` iterator (e.g. `for (int i = 0; i < size; i++)`). You'd have to update the access at `31`, `40` and `457`, but I think it would be totally worth it for just code clarity. Modifying size to use as an iterator is a _really_ bad pattern, as it gives the completely wrong impression on what is going on...especially at `457`. Think you can do one additional update to simply switch this to a `for` loop instead?
> 
> As for whether or not this needs an issue -- I'm working on getting the updated contributing guidance up by tomorrow. We've internally settled on keeping it consistent that there should be an approved issue for every PR that is reviewed. The bar for getting an issue approved will be lower for changes that don't impact what an end-user will see. PR's that are out before the updated contribution documentation is committed will be grandfathered in on a case-by-case basis. This one will be grandfathered in to not need the issue.

I can't more agree, this is the kind of "I try to optimize my code" but in fact, it adds bugs/makes the code harder to read, create an .exe less optimized because the compiler had more troubles to optimize the code".

HowardWolosky:
Closing (and will immediately re-open) in order to kick off the PR build again.

PR #273, opened 2019-03-12
Lossless optimization of images

kolappannathan:
- The images were optimised using ImgBot a bot that optimizes images losslessly.


HowardWolosky:
#229 has issued a similar change, but against more files.  It would be good to understand the difference in the optimization result between these two PR's.

JacobMGEvans:
I sometimes use ImageOptim, it is really hard to get a breakdown of saved space from git add/commit though there would usually be pretty big differences. Usually the lossless just remove superfluous metadata (if you didn't already know).   

I think the best way to compare the difference would probably check a couple of similar files that have been optimized between the two, make sure the #229 hasn't done any lossy compression and see which was more successful in reducing the sizes.

HowardWolosky:
To be clear, the _content_ of this specific PR isn't as interesting to me as is the _process_.  None of the image assets in the repo ship with the official Store app (you can see that all of the icons are "dev" versions of the icons to make it easier to differentiate the dev build from the official build).

We already have a process in place internally that removes all extraneous metadata from our shipped images using [`TrunePNG`](https://css-ig.net/articles/truepng.php), and that same process was applied on these assets as well (although the repo-only/documentation assets might have been overlooked).  GitHubBot (which you referenced in your PR description) won't help us for those, unfortunately, since those assets aren't hosted on GitHub.

I'll take some time later to do a comparison betwee the two PR's to see if there is a big difference between the two PR's, as well as if either PR provides any substantial difference with what's already checked-in.  Thanks again for the contribution.

JacobMGEvans:
Thank you for elaborating and taking the time to read my comment.

grischard:
https://imgbot.net/docs/#compression says that it uses ImageMagick. In my experience, it's fast but doesn't squeeze images as much as other tools.

[ImageOptim](https://github.com/ImageOptim/ImageOptim), which @JacobMGEvans mentioned, is indeed what I used in #229. It compares the output of various lossless tools and picks the smallest, so it does a bit more than stripping metadata. The `PNGOUT` and `Zopfli` tools in particular take a bit more CPU, but are very effective - but not always the most effective, which makes ImageOptim's approach so interesting.

JacobMGEvans:
@grischard I certainly oversimplified it, that is my bad. Definitely my favorite tool for the job.

HowardWolosky:
@kolappannathan, #229 has been merged in.  Can you sync up and re-run your toolset to see if there is still any difference?  If there still is an improvement in filesize, please add exact details on how to acquire, configure and run your toolset in #307.  Thanks.

kolappannathan:
@HowardWolosky Thank you for your time. I'm in travel right now. Will do it on Monday.

kolappannathan:
@HowardWolosky There is no improvements to the current images.

HowardWolosky:
@kolappannathan Thanks for checking.  Sorry _this_ change didn't go in.  Looking forward to seeing your next contribution.

PR #287, opened 2019-03-13
Fix separators for unit tests

### LINKED ISSUE #272: 45 unit tests don't pass with some locales
**Describe the bug**

45 unit tests fail when the computer is in French for example:
image

(Probably more with other languages)

**Steps To Reproduce**
- Change system language to French
- Run unit tests

**Expected behavior**
- All unit tests should pass whatever the language used on the computer running them. Non-english developers should not be forced to modify the language of their computers.

 - OS Build: 10.0.18348.0
 - Architecture: X64

uhliksk:
- Hardcoded separators for unit tests

- manual

### DIFF
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -108,6 +108,18 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
         m_nTempCom = (INT)wParam;
     }
 
+    // Hardcoded settings for unit tests
+    if (wParam == IDC_TESTS)
+    {
+        m_decimalSeparator = L'.';
+        SetDecimalSeparator(m_decimalSeparator);
+        m_groupSeparator = L',';
+        m_decGrouping = DigitGroupingStringToGroupingVector(L"3;0");
+        m_input.SetDecimalSymbol(m_decimalSeparator);
+        m_HistoryCollector.SetDecimalSymbol(m_decimalSeparator);
+        s_engineStrings[IDS_DECIMAL] = m_decimalSeparator;
+    }
+
     if (m_bError)
     {
         if (wParam == IDC_CLEAR)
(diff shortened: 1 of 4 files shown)


rudyhuyn:
I'm afraid it will only partially fix the issue but it's a good start! It will still use the wrong `CEngineStrings.resw`, not fix issues with currencies and `LocalizationService` will still return the wrong formats.

I'm working on a diff, I will CC you for review!

uhliksk:
> I'm afraid it will only partially fix the issue but it's a good start!

Yes, of course. This is only for numbers. There are some "unfixable" tests like DateCalcViewModelAddSubtractResultAutomationNameTest where I don't know any method to force OS to return en-us day of week name while user have other locale set.

uhliksk:
> I'm afraid it will only partially fix the issue but it's a good start!

I'll try to fix other things as you metioned in your original comment.

rudyhuyn:
> > I'm afraid it will only partially fix the issue but it's a good start!
> 
> I'll try to fix other things as you metioned in your original comment.

Can you take a look at this diff? https://github.com/Microsoft/calculator/pull/289

It uses a different approach, the idea was to force the locale at the lower level possible (`LocaleService`). Don't hesitate to share feedback!

uhliksk:
> Can you take a look at this diff? #289
> It uses a different approach, the idea was to force the locale at the lower level possible (LocaleService). Don't hesitate to share feedback!

Nice solution but actually I was thinking today about something else. I think better way will be to convert expected values to current locale because if we will test everything with en-US only there is a chance we will miss some bugs for international users. Then we can easily repeat all tests for all existing locales to be sure calculator is working expected way in all international formats.

HowardWolosky:
@uhliksk  - It sounds like this isn't the approach that you want to take forward (and I agree, this doesn't feel quite right).  I'm going to close out this PR.  Please let me know if I misunderstood.

PR #290, opened 2019-03-13
Remove .editorconfig from sln

rudyhuyn:
## Fixes #257

Clean the `.sln` to remove the reference to the extra `.editorconfig` already removed by another diff.


HowardWolosky:
Fixes #257

#225 accidentally added a duplicate .editorconfig to the repo as well as to the solution.
#257 removed the duplicate .editorconfig, but not its entry in the solution.
This update removes the extra .editorconfig reference to a version that #257 already removed from the repo.

PR #296, opened 2019-03-14
Fix wrong results after pow and sqrt

### LINKED ISSUE #222: Wrong result for e.g. √(2.25) - 1.5

uhliksk:
- Pow is calculated for Numerator and Denominator separately to avoid rounding difference
- gcdrat fixed to not accept result if optimization caused change in numeric value

- new automated tests added


HowardWolosky:
Replaced by #297

PR #297, opened 2019-03-14
Fix wrong results after pow and sqrt

### LINKED ISSUE #222: Wrong result for e.g. √(2.25) - 1.5

uhliksk:
- Pow is calculated for Numerator and Denominator separately to avoid rounding difference
- gcdrat fixed to not accept result if optimization caused change in numeric value

- new automated tests added


HowardWolosky:
Previous conversation on this can be seen in #231

HowardWolosky:
@joshkoon / @jlaanstra - Interested in your feedback on this one.  The issue hasn't been approved for fixing yet, but the change might be worth a look now anyway.  Thanks.

joshkoon:
@uhliksk, apologies for the delay on the review. This does appear to fix the issue. I think this is a good time, however, to start building out unit tests for the ratpack directly. Could you add a RatPackTests.cpp class to the unit tests project, and create some tests for the functions you're modifying? Namely, gcdrat(), and powrat()? I won't ask that they're comprehensively tested right now, but we should at least have enough test coverage there to be confident in these changes.

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -392,6 +392,31 @@ void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precis
     destroyrat(pxPow);
 }
 
+void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precision)
+{
+    // We need px as simple as possible
+    gcdrat(px, precision);
```
joshkoon: Looking at the code for the underlying gcd() function, there are two assumptions made about the inputs:
1. Inputs are in BASEX
2. Inputs are integers

Here, the powrat* functions can be called with a different radix than BASEX, however, which would cause gcdrat to fail. Debugging this, however, shows that CalcInput::ToRational always generates a BASEX Rational, so this ends up working out.

I think there are more issues to dig into around how radices are used and assumptions made around them, but for the purposes of this change, could you see if the fix still works if you:
1. Pass the radix to gcdrat as a parameter
2. Inside gcdrat, if radix != BASEX, convert `a` to BASEX by calling numtonRadixx() on a->pp and a->pq (this is what I haven't checked: does numtonRadixx work if you SAY radix is 10, when in fact the PNUMBER passed is in BASEX?)
3. Perform the gcd calculation
4. Convert output back into provided radix (if not BASEX) using nRadixxtonum and return (this will likely not actually work right now, since the passed in values will actually be BASEX due to CalcInput::ToRational)

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -297,7 +297,7 @@ void powrat(PRAT *px, PRAT y, uint32_t radix, int32_t precision)
     }
 }
 
-void powratNumeratorDenominator(PRAT *px, PRAT y, uint32_t radix, int32_t precision)
+void powratNumeratorDenominatorPartial(PRAT *px, PRAT y, uint32_t radix, int32_t precision)
```
joshkoon: Since this is using the Power of Powers rule, why don't we name it accordingly, e.g. `powratPowerOfPowers`?

joshkoon: Let's stay consistent with the rest of ratpak and declare this in ratpak.h

joshkoon: Please also add a comment description of the function similar to the one on powrat, describing what calculation it's performing. We should also add one for powratNumeratorDenominator below.

src/CalcManager/Ratpack/rat.cpp
```diff
@@ -54,9 +54,14 @@ void gcdrat( PRAT *pa, int32_t precision)
...
 
     destroynum( pgcd );
-    *pa=a;
 
-    RENORMALIZE(*pa);
+    RENORMALIZE(a);
+
+    if (rat_equ(*pa, a, precision))
+    {
+        DUPRAT(*pa, a);
```
joshkoon: This check works, but it seems like it's just a catch-all for errors caused by bad inputs. It would be preferable to do a better job of validating the inputs instead, as noted in the comment above.

If it isn't possible to just validate our inputs due to assumptions made elsewhere in code, let's have a conversation then about what the right fix will look like.

HowardWolosky:
@uhliksk - Looks like you have some unaddressed feedback from @joshkoon.  Do you think you have time to provide an update that to address that feedback?  Thanks!

uhliksk:
> @uhliksk - Looks like you have some unaddressed feedback from @joshkoon.  Do you think you have time to provide an update that to address that feedback?  Thanks!

I'm sorry. I have lot of work with accounting because I had to submit annual tax report for our local Financial Administration office before March 31st. I'll respond to all feedback later this week. Thank you for your understanding.

HowardWolosky:
@uhliksk - Checking-in to see if you'll be getting back to this PR.  Thanks!

uhliksk:
Hello, sorry for delay in response. My second child has been born last week. I'll come back to this PR much later than I expected initially.

sanderl:
> Hello, sorry for delay in response. My second child has been born last week. I'll come back to this PR much later than I expected initially.

Congratulations!

Rahul-Dixit:
@uhliksk This pull request has been pending for a long time. Do you think you have the time to get back to it?

uhliksk:
> @uhliksk This pull request has been pending for a long time. Do you think you have the time to get back to it?

@Rahul-Dixit Sorry for very long inactivity. I'll have some spare time next week. I can review all comments and complete this PR finally.

src/CalcManager/Ratpack/exp.cpp
uhliksk: I renamed as you suggested.

uhliksk: I added declaration in ratpak.h and added a comment description as you suggested.

src/CalcManager/Ratpack/rat.cpp
```diff
@@ -44,7 +44,7 @@ void gcdrat( PRAT *pa, int32_t precision)
     PNUMBER pgcd= nullptr;
     PRAT a= nullptr;
 
-    a=*pa;
+    DUPRAT(a,*pa);
     pgcd = gcd( a->pp, a->pq );
 
     if ( !zernum( pgcd ) )
```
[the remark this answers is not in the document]
uhliksk: I cleaned up formatting.

uhliksk:
I made some progress finally. Thank you for your patience. I hope I'll be able to review remaining comments in next few days.

sanderl:
We're converting Calculator to C# with a PR for the conversion coming in the next few days, so we're going to close this PR. If you want to revisit the issue, please open a new PR after the conversion has completed.

PR #298, opened 2019-03-14
Fix invalid check of m_precedenceOpCount

### LINKED ISSUE #122: Invalid check of m_precedenceOpCount
https://github.com/Microsoft/calculator/blob/1eb717f336bae8b35281bbb27571748d458d862e/src/CalcManager/CEngine/scicomm.cpp#L515

https://github.com/Microsoft/calculator/blob/1eb717f336bae8b35281bbb27571748d458d862e/src/CalcManager/Header%20Files/CalcEngine.h#L115

`m_precedenceOpCount` is of unsigned type `size_t`, yet it is tested for being larger than 0.

danbelcher-MSFT:
- The conditional `m_precedenceOpCount >= 0` was always true because `m_precendenceOpCount` is an unsigned type. Update the conditional to simply be true and rely on a break statement in the loop. Although this member variable used to be a signed type, in practice, the value was never less than 0.

- Manual.  Unit tests pass locally.

### DIFF
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -514,7 +514,7 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
             m_HistoryCollector.PopLastOpndStart();
 
             m_bNoPrevEqu = true;
-        } while (m_precedenceOpCount >= 0);
+        } while (true);
 
         if (!m_bError)
         {
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -761,11 +761,12 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
 // Helper function to resolve one item on the precedence stack.
 void CCalcEngine::ResolveHighestPrecedenceOperation()
 {
-    if (m_nOpCode) /* Is there a valid operation around?        */
+    // Is there a valid operation around?
+    if (m_nOpCode)
     {
-        /* If this is the first EQU in a string, set m_holdVal=m_currentVal */
-        /* Otherwise let m_currentVal=m_holdVal.  This keeps m_currentVal constant */
-        /* through all EQUs in a row.                     */
+        // If this is the first EQU in a string, set m_holdVal=m_currentVal 
+        // Otherwise let m_currentVal=m_holdVal.  This keeps m_currentVal constant 
+        // through all EQUs in a row.
         if (m_bNoPrevEqu)
         {
             m_holdVal = m_currentVal;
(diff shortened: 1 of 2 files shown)


danbelcher-MSFT:
@janisozaur

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -514,7 +514,7 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
             m_HistoryCollector.PopLastOpndStart();
 
             m_bNoPrevEqu = true;
-        } while (m_precedenceOpCount >= 0);
+        } while (true);
```
joshkoon: This would be the simple fix, but I think we could clean this up a bit to make things more readable:

1. Move the if/else from lines 468-499 into a new function, e.g. `void ResolveCurrentPrecedenceOperation()`
2. Convert the remaining code in the do/while loop into a regular while loop, using the condition at line 501: `while (m_precedenceOpCount > 0 && m_fPrecedence)`
3. Call the function from (1) at the bottom of the while loop.

```c++
ResolveCurrentPrecedenceOperation();

// Evaluate the precedence stack 
while (m_precedenceOpCount > 0 && m_fPrecedence)
{
    --m_precedenceOpCount;
    m_nOpCode = m_nPrecOp[--m_precedenceOpCount];
    m_lastVal = m_precedenceVals[m_precedenceOpCount];

    // Precedence Inversion check
    ni = NPrecedenceOfOp(m_nPrevOpCode);
    nx = NPrecedenceOfOp(m_nOpCode);

    if (ni <= nx)
    {
        m_HistoryCollector.EnclosePrecInversionBrackets();
    }
    m_HistoryCollector.PopLastOpndStart();

    m_bNoPrevEqu = true;

    ResolveCurrentPrecedenceOperation();
}
```

It isn't perfect, but I don't like having a `while(true)` with the condition buried in the center of the loop.

danbelcher-MSFT: Sure thing!

joshkoon: Edited my code block above since I realized I had a double-decrement in the while loop.

danbelcher-MSFT:
@HowardWolosky, Josh can't be the official reviewer because he doesn't have write access.  Someone from our team needs to approve.

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -789,6 +758,45 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
...
+
+        // Do the current or last operation.
+        m_currentVal = DoOperation(m_nOpCode, m_currentVal, m_lastVal);
+        m_nPrevOpCode = m_nOpCode;
+        m_lastVal = m_currentVal;
+
+        /* Check for errors.  If this wasn't done, DisplayNum */
+        /* would immediately overwrite any error message.     */
+        if (!m_bError)
+            DisplayNum();
```
HowardWolosky: I realize that you tried to just move the code verbatim to this helper function, but it would be great to clean it up a little bit since we're touching it.  Here's a good opportunity to wrap the conditional in brackets, since it's generally asking for trouble to leave single-line conditionals like this.

Optional: simplify a number of the comments from being `/*   */` with lots of unnecessary spaces trying to (unsuccessfully) keep alignment to just using the `//` syntax.

PR #306, opened 2019-03-14
ViewModelProperties namespaces converted to static class properties.

### LINKED ISSUE #305: Expose ViewModel property names as static class members
The CalcViewModels currently follow a paradigm of declaring a namespace to store strings that are property names of the matching class, for example:
https://github.com/Microsoft/calculator/blob/ac5292cf88e958c15c34c6a6236766fe3e820126/src/CalcViewModel/StandardCalculatorViewModel.h#L34-L39

These strings are then used by the View layer in PropertyChanged event handlers, to identify if a specific property changed, e.g.
https://github.com/Microsoft/calculator/blob/ac5292cf88e958c15c34c6a6236766fe3e820126/src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp#L69-L75

## Problem
The problem with this approach is that the namespaces storing these strings are only related to the ViewModel class by the namespace name.  The strings within otherwise have no relation to the ViewModel they are supposed to be associated with.  Furthermore, namespaces can be reopened and added to at-will, meaning the ViewModel has no guarantee about what content goes in the namespace.  Correctness is left for the programmer to manually verify.

## Solution
CX classes with DependencyProperties follow a common convention for implementation of the properties.  Per property, there will be a `static property DependencyProperty^ *PropertyName*Property` and a non-static `property *Type* *PropertyName*`.  Similarly, for properties that need to be referenced by name, we should define `static property Platform::String^ *PropertyName*PropertyName` and `property *Type* *PropertyName*`.

The benefit of this change is that the property name is now a static member of the ViewModel class it is associated with.  This makes sense because the property name _is_ a part of the ViewModel as much as the property itself is also a member.  Additionally, the ViewModel has full control over what properties it names and no other parts of the program can add to the ViewModels interface.

danbelcher-MSFT:
- Implement the `OBSERVABLE_NAMED_PROPERTY` class of macros. This macro will define the requested property as a member property and will additionally define a static class property that provides the name of the property as a string.
- Remove unused propertynames that were never used.
- For places where the source file still need the property name, e.g. to raise a property changed event or when internally checking for property change events, moved the property names to an anonymous namespace for the file.

- Manual. Unit tests pass locally.
- Smoke tests.  Verify properties that were checked via code-behind, e.g. IsMemoryEmpty, still correctly functioned after the change.


HowardWolosky:
@danbelcher-MSFT - Looks like recent merges have introduced conflicts.  Can you sync/merge/push an update?

src/CalcViewModel/Common/Utils.h
```diff
@@ -42,10 +42,16 @@
...
     } private: t m_##n; public:
 
-#define NAMED_OBSERVABLE_PROPERTY_RW(t, n)\
+#define OBSERVABLE_NAMED_PROPERTY_R(t, n)\
+    OBSERVABLE_PROPERTY_R(t, n)\
+    internal: static property Platform::String^ n##PropertyName {\
+    Platform::String^ get() { return Platform::StringReference(L#n); }\
+    } public:
+
+#define OBSERVABLE_NAMED_PROPERTY_RW(t, n)\
```
sanderl: Does this need a set()?

danbelcher-MSFT: No, and that's intentional.  If it had a setter, then outside clients could change the value of the string returned for the named property.  The property name should be read-only data.

PR #241, opened 2019-03-10
Propose minor code-cleanups

Maharramoff:
### Description of the changes:
- Remove unnecessary 'else', 'continue' statements

### How changes were validated:
- Manual.

PR #314, opened 2019-03-15
Remove redundant if condition.

PawanKartikS:
NumbersAndOperatorsEnum::None != op check was already done in the parent "if" (line 972). Hence this redundant check can be dropped safely.


HowardWolosky:
Thanks for this submission, but this change is already a part of #241.  Sorry that this one can't go in, but I look forward to your next contribution.

PR #304, opened 2019-03-14
Code analysis

ariccio:
## Fixes:
- Inconsistent annotations causing warnings
- Missing annotations causing warnings about uninitialized _Out_ parameters
- Incorrect annotations in `_destroynum` and `_destroyrat` causing spurious warnings.
- Analyzer crashes on `CalcViewModel` with 32 bit toolchain.

### Description of the changes:
- I had to fix the annotations on a number of functions so that code analysis would build with a reasonable number of warnings. Most of them were just lacking the annotations on the definition, so I could've used `_Use_decl_annotations_`, but it looks more consistent this way.
- Also, analyzer on CalcViewModel uses a ton of memory, and so it won't complete without crashing when run as a 32 bit process. So I enabled the x64 build for it.

Code analysis reveals a couple codepaths where there could be nullptr dereferences. For example, this looks like a legitimate warning:
pnumret
...where `stripzeroesnum` immediately dereferences the pointer.

### How changes were validated:

- No actual functional changes to code. None needed?

There are a whole bunch of other warnings, like uninitialized member variables and also some arithmetic overflow warnings, but it's less obvious if they're legit.

PR #315, opened 2019-03-15
Code analysis

ariccio:
(this is a redo of #304)

## Fixes:
- Inconsistent annotations causing warnings
- Missing annotations causing warnings about uninitialized _Out_ parameters
- Incorrect annotations in `_destroynum` and `_destroyrat` causing spurious warnings.
- Analyzer crashes on `CalcViewModel` with 32 bit toolchain.

- I had to fix the annotations on a number of functions so that code analysis would build with a reasonable number of warnings. Most of them were just lacking the annotations on the definition, so I could've used `_Use_decl_annotations_`, but it looks more consistent this way.
- Also, analyzer on CalcViewModel uses a ton of memory, and so it won't complete without crashing when run as a 32 bit process. So I enabled the x64 build for it.

Code analysis reveals a couple codepaths where there could be nullptr dereferences. For example, this looks like a legitimate warning:
pnumret
...where `stripzeroesnum` immediately dereferences the pointer.

- No actual functional changes to code. None needed?

There are a whole bunch of other warnings, like uninitialized member variables and also some arithmetic overflow warnings, but it's less obvious if they're legit.


ariccio:
Goddarnit, it keeps getting conflicted!

janisozaur:
Sorry, that'd be #212. Once you rebase, I'll check with GCC and Clang.

ariccio:
Yeah, that's [generally](https://github.com/Microsoft/calculator/pull/212/files#r269811274) a good change though, using platform/system independent types, so nbd!

PR #320, opened 2019-03-16
Update the year in the Copyright string

### LINKED ISSUE #319: The About page still displays Copyright 2018 even if the last public build were made in 2019
**Describe the bug**

Even if the last public build were created in February, the About page still displays 2018 in the About dialog.

image

In order to simplify the process and not having to localize new strings every year (or manually update resw files), a better solution would be to automatize the process:
- templatize the copyright string
- use the date of the compilation to retrieve the year (using preprocessor directive)

**Steps To Reproduce**
- Open the app
- Click on About
- The dialog displays "Copyright 2018"

**Expected behavior**
- The dialog should display "Copyright 2019"

 - OS Build: 10.0.18348.0
 - Architecture: X64

rudyhuyn:
## Fixes #319

Templatize the copyright string and use a preprocessor definition to set the year, so the build server will be able to update it.

- tested with English and French and with different dates.

### DIFF
--- a/src/Calculator/pch.h
+++ b/src/Calculator/pch.h
@@ -21,6 +21,7 @@
 #include <sstream>
 #include <concrt.h>
 #include <regex>
+#include <string>
 
 // C++\WinRT Headers
 #include "winrt/base.h"
@@ -44,5 +45,3 @@ namespace CustomPeers = CalculatorApp::Common::Automation;
 
 // Project Headers
 #include "App.xaml.h"
-
-#include <string>
--- a/src/Calculator/pch.h
+++ b/src/Calculator/pch.h
@@ -44,3 +44,5 @@ namespace CustomPeers = CalculatorApp::Common::Automation;
 
 // Project Headers
 #include "App.xaml.h"
+
+#include <string>
--- a/src/Calculator/AboutFlyout.xaml
+++ b/src/Calculator/AboutFlyout.xaml
@@ -38,7 +38,7 @@
                 <Paragraph>
                     <Run x:Name="AboutFlyoutVersion"/>
                     <LineBreak/>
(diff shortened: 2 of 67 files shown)


janisozaur:
I'm against this in current form. Using macros like `__DATE__` and `__TIME__` introduces a number of issues:

* It makes reproducible builds harder
* It invalidates compiler cache

GCC has even added `-Wdate-time` to help reduce usage of such macros.

Instead, I would suggest injecting the value from the build system or keeping it as a verbatim in source code in a way that doesn't depend on the [builder's environment](https://github.com/Microsoft/calculator/issues/272). This should be made easy to inspect what string exactly was used. This PR replaces most (all?) occurrences of the copyright year with a single source of truth, which should be easy to maintain.

rudyhuyn:
Interesting feedback, especially the reproducible build part, I will get
rid of the __DATE__ macro tonight.

Rudy

src/Calculator/pch.h
```diff
@@ -44,3 +44,5 @@ namespace CustomPeers = CalculatorApp::Common::Automation;
 
 // Project Headers
 #include "App.xaml.h"
+
+#include <string>
```
HowardWolosky: I'm sure @janisozaur will love seeing _another_ heading being added to PCH.... ;)

rudyhuyn: Sorry! 

(I will move this one to the top, just before WinRT Headers)

rudyhuyn:
> @rudyhuyn -- Given the change from `x:Uid` to `x:Name`, can you confirm that you've also tested this change with [Narrator](https://support.microsoft.com/en-us/help/22798/windows-10-complete-guide-to-narrator) so that we can be sure there's no accessibility regression?

Tested, it still works well.

PR #321, opened 2019-03-16
Use operator""sv to ensure string_view are initialized correctly

janisozaur:
Addresses following warning raised by clang:
```
src/CalcManager/CEngine/scidisp.cpp:23:24: error: constexpr variable 'c_decPreSepStr' must be initialized by a constant expression
constexpr wstring_view c_decPreSepStr = L"[+-]?(\\d*)[";
                       ^                ~~~~~~~~~~~~~~~
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/8.2.1/../../../../include/c++/8.2.1/bits/char_traits.h:431:11: note: non-constexpr function 'wcslen' cannot be used in a constant expression
          return wcslen(__s);
                 ^
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/8.2.1/../../../../include/c++/8.2.1/string_view:100:39: note: in call to 'length(&L"[+-]?(\\d*)["[0])'
      : _M_len{__str == nullptr ? 0 : traits_type::length(__str)},
                                      ^
```

- Addresses a warning raised by clang. I'm not 100% positive it's not a bug in the STL implementation used by my compiler, but doesn't cause other issues.

### DIFF
--- a/src/CalcManager/CEngine/calc.cpp
+++ b/src/CalcManager/CEngine/calc.cpp
@@ -19,8 +19,8 @@ static constexpr long DEFAULT_RADIX = 10;
 
 static constexpr wchar_t DEFAULT_DEC_SEPARATOR = L'.';
 static constexpr wchar_t DEFAULT_GRP_SEPARATOR = L',';
-static constexpr wstring_view DEFAULT_GRP_STR = L"3;0";
-static constexpr wstring_view DEFAULT_NUMBER_STR = L"0";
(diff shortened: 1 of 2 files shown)


janisozaur:
By using libc++ instead of libstdc++, I've confirmed this to be a bug with the latter. I'll check what's the status in libstdc++ 9 and file a bug if needed.

PR #323, opened 2019-03-17
Add explicit [[fallthrough]] attribute in Ratpack

janisozaur:
This is the only one I found in Ratpack.

### DIFF
--- a/src/CalcManager/Ratpack/conv.cpp
+++ b/src/CalcManager/Ratpack/conv.cpp
@@ -612,6 +612,7 @@ PNUMBER StringToNumber(wstring_view numberString, uint32_t radix, int32_t precis
                 break;
             }
             // Drop through in the 'e'-as-a-digit case
+            [[fallthrough]];
         default:
             state = machine[state][NZ];
             break;
@@ -646,7 +647,7 @@ PNUMBER StringToNumber(wstring_view numberString, uint32_t radix, int32_t precis
             break;
         case LD:
             pnumret->exp++;
-            // Fall through
+            [[fallthrough]];
         case DD:
             {
                 curChar = NormalizeCharDigit(curChar, radix);


janisozaur:
I have added one more `[[fallthrough]]` attribute: https://github.com/Microsoft/calculator/blob/78aab3767c1cff88fb7edb61a19118aa95d9f809/src/CalcManager/Ratpack/conv.cpp#L650 and just to be sure, I also rechecked `CEngine`, finding nothing to fix in there.

While I see your arguments about the empty cases, it would be a unique case, I know no other project that would elect to use such form.

To support my case, I quote this [snippet from the C++ Standard Draft](https://github.com/cplusplus/draft/blob/2d94f2f59be31ebdae9533adb69f34f27ef97139/source/declarations.tex#L9117-L9130):
```cpp
void f(int n) {
  void g(), h(), i();
  switch (n) {
  case 1:
  case 2:
    g();
    [[fallthrough]];
  case 3:                       // warning on fallthrough discouraged
    h();
  case 4:                       // implementation may warn on fallthrough
    i();
    [[fallthrough]];            // ill-formed
  }
}
```

janisozaur:
The CI job is red with following error:

> ##[Error 1]
The agent: essential000004 lost communication with the server. Verify the machine is running and has a healthy network connection.

https://dev.azure.com/ms/calculator/_build/results?buildId=6533

HowardWolosky:
Thanks -- re-running that leg.

HowardWolosky:
Your CR feedback response was thorough and convincing.  Thanks for updating the other instance where there had been the `// fallthrough` comment.  I reviewed the [guidelines that we try to follow](http://isocpp.github.io/CppCoreGuidelines), and [they concur](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-break) that empty cases don't need the `[[fallthrough]]` annotation. but all non-empty ones either need `break` or `[[fallthrough]]`.

PR #329, opened 2019-03-18
Remove redundant type qualifiers on function return type

janisozaur:
Remove a redundant CV-qualifier on function return type


janisozaur:
GCC

PR #330, opened 2019-03-18
Remove unused member field from CalculatorHistory

janisozaur:
`m_mode` field was assigned to, but never read from.

### DIFF
--- a/src/CalcManager/CalculatorHistory.cpp
+++ b/src/CalcManager/CalculatorHistory.cpp
@@ -7,8 +7,7 @@
 using namespace std;
 using namespace CalculationManager;
 
-CalculatorHistory::CalculatorHistory(CALCULATOR_MODE eMode, size_t maxSize) :
-    m_mode(eMode),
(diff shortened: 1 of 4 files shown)


src/CalcManager/CalculatorHistory.cpp
```diff
@@ -7,8 +7,7 @@
 using namespace std;
 using namespace CalculationManager;
 
-CalculatorHistory::CalculatorHistory(CALCULATOR_MODE eMode, size_t maxSize) :
-    m_mode(eMode),
+CalculatorHistory::CalculatorHistory(size_t maxSize) :
```
HowardWolosky: @danbelcher-MSFT - Can you check the archived history to see if the fact that this is still here is the bug, or the fact that the code referencing it is absent is the bug?

danbelcher-MSFT: The member variable exists for this class as far back as we have archived history.  It was never used from the earliest record.  Seems safe enough to remove.  I believe CalculatorManager does the logic of checking if the current mode supports history or not.

PR #331, opened 2019-03-18
Make CalculatorManager final rather than sealed

janisozaur:
I have no idea if it is required to be `sealed`, I have seen no `^`
operator which makes me think it could be a regular C++ code, barring
the concurrency stuff.


HowardWolosky:
@danbelcher-MSFT  / @joshkoon  / @jlaanstra - what say you?

danbelcher-MSFT:
Interesting that `sealed` works here...  `final` is appropriate and seems safe.

PR #332, opened 2019-03-18
Fix order of initialization list in CalculatorManager ctor

janisozaur:
This ensures the initialization order matches the layout of member
fields in class declaration

This will conflict with https://github.com/Microsoft/calculator/pull/330, but I will rebase the loser


HowardWolosky:
@janisozaur  - Looks like #330 did go in first after all.  Can you resolve the conflict? Then we can get this merged in.  Thanks!

janisozaur:
Updated

PR #334, opened 2019-03-18
Replace fallthrough comment with C++17's attribute

janisozaur:
Sorry for making 3 PRs for the same kind of issue, I was not able to verify this file earlier.

This time, however, I grepped the whole project for similar comments and this is the final one.

### DIFF
--- a/src/CalcManager/CalculatorManager.cpp
+++ b/src/CalcManager/CalculatorManager.cpp
@@ -292,7 +292,7 @@ namespace CalculationManager
             break;
         case Command::CommandFE:
             m_isExponentialFormat = !m_isExponentialFormat;
-            // fall through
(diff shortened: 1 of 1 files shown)


janisozaur:
As of f3cd220:
```
$ git grep -in fall.\*thr
src/CalcManager/CEngine/scifunc.cpp:172:            // This case falls through to IDC_DMS case because in the old Win32 Calc, 
src/CalcManager/CEngine/scifunc.cpp:176:            [[fallthrough]];
src/CalcManager/CalculatorManager.cpp:295:            [[fallthrough]];
src/CalcManager/Ratpack/conv.cpp:615:            [[fallthrough]];
src/CalcManager/Ratpack/conv.cpp:650:            [[fallthrough]];
```

PR #342, opened 2019-03-20
Minor corrections to the german locale.

vonunwerth:
## Fixes #.
Minor corrections to the german locale.

Strings in Resources.resw for de-DE edited. Despriction of the changes: 'Arbeitsspeicher' --> 'Speicher', there is just the memory. A split in memory and random access memory (Arbeitsspeicher) is confusing. 'Gespeicherten Wert speichern' --> 'Angezeigten Wert speichern', there is no memory saved till i hit the set (MS) button. 'Tagens' in line 918 --> 'Tangens', just a spelling mistake. 'Modus-Dropdownfeld' and 'Kategorie-Dropdownfeld' --> '-Auswahlfeld', 'Auswahl' is more common than Dropdown. 'L' --> 'l', cause the official abbreviation for liter (ISO, IEC) is l. Also edited this for milliliter. 'wo' --> 'w', with respect to ISO8601. Line 1597 changed to 'Acre' insted of 'Morgen'. 'cd' --> 'Kt' as common abbreviation for Carat. 'Hand' --> 'Hände', because there are just the plural forms used.

Official reference work like Duden and my native language experience.

### DIFF
--- a/src/Calculator/Resources/de-DE/Resources.resw
+++ b/src/Calculator/Resources/de-DE/Resources.resw
@@ -422,7 +422,7 @@
     <comment>This is the automation name and label for the memory button when the memory flyout is open.</comment>
   </data>
   <data name="MemoryButton.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
-    <value>Arbeitsspeicher</value>
+    <value>Speicher</value>
     <comment>This is the tool tip automation name for the memory button.</comment>
   </data>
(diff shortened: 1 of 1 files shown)


HowardWolosky:
Unfortunaely, our Localizaton system is not yet able to directly accept localization changes directly through this repo.  I've worked with our Localization team to identify a temporary workaround to make sure that this contribution isn't lost.  I have filed an issue in our internal tracking system (`20912928`) which tracks all of your suggested changes, and the Localization team will use that internal issue to review and apply your changes via their system.

Your participation is very much appreciated, and your work _will_ get applied.  Unfortunately, I cannot directly merge this change in at this time. 

Thanks again.

PR #366, opened 2019-03-24
Smart pointers

### LINKED ISSUE #61: Update Calculator engine’s use of smart pointers

vlogozzo:
-moved the Rational class over to smart pointers(spawned in `ToSmartRat()`)

-passed manual tests
-lots of watching memory allocate and de-allocate

### DIFF
--- a/src/CalcManager/CEngine/Rational.cpp
+++ b/src/CalcManager/CEngine/Rational.cpp
@@ -31,22 +31,18 @@ namespace CalcEngine
 
     Rational::Rational(int32_t i)
     {
-        PRAT pr = longtorat(static_cast<long>(i));
+        unique_ptr<RAT, decltype(&_destroyrat)> pr(Ulongtorat(static_cast<unsigned long>(i)), &_destroyrat);
 
         m_p = Number{ pr->pp };
         m_q = Number{ pr->pq };
-
-        destroyrat(pr);
     }
 
     Rational::Rational(uint32_t ui)
     {
-        PRAT pr = Ulongtorat(static_cast<unsigned long>(ui));
+        unique_ptr<RAT, decltype(&_destroyrat)> pr(Ulongtorat(static_cast<unsigned long>(ui)), &_destroyrat);
 
         m_p = Number{ pr->pp };
         m_q = Number{ pr->pq };
-
-        destroyrat(pr);
     }
 
     Rational::Rational(uint64_t ui)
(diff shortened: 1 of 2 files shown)


danbelcher-MSFT:
Hi Vincent.  First, I want to say thank you for your work on this issue.


After looking at the change, I think that it doesn't really address the issue that is referenced and so further changes are needed.  The core task in this issue is to eliminate the engine's reliance on `new`, `zmalloc`, and `delete`.  There are a number of ways to accomplish this, for example, using smart pointers.  This change does eliminate a few places where `destroyrat` is called, but there are still plenty of other cases throughout the engine.  I would expect a comprehensive fix for this issue to, in some way, remove all cases of `destroyrat` and `_destroyrat`.

Similarly, `createrat` and `_createrat` still exist.  Smart pointers are fundamentally about expressing _ownership_ of the memory.  Today's implementation of `_createrat` allocates memory and returns a pointer, that is left to anyone else to delete later.  One way I could see us avoid `createrat` is:
```C++
struct RAT
{
    std::unique_ptr<NUMBER> pp;
    std::unique_ptr<NUMBER> pq;
};
```

Now it's clear that a `RAT` owns the memory for its member data and there is no need for a custom delete functionality.  Then we can delete `createrat` entirely, because places that want a `PRAT` can instead use `auto prat = make_unique<RAT>()`.  Something like `Rational::ToPRAT` could then be updated like:
```C++
unique_ptr<RAT> Rational::ToPRAT() const
{
    auto ret = make_unique<RAT>();

    ret->pp = this->P().ToPNUMBER();
    ret->pq = this->Q().ToPNUMBER();

    return ret;
}
```

This is the _right_ change to make, because we are now using smart pointer types to _clearly_ indicate to the caller that they must own this memory and figure out how to control its lifetime.  This decision is enforced by the use of `unique_ptr`.  Using `make_unique` means we can delete `createrat`; its functionality is provided by `make_unique` in addition to the ownership semantics that we gain.


Deleting `createrat`, deleting `destroyrat`, updating `ToPRAT` to return a `unique_ptr`, these changes will cause a ripple of compilation errors throughout the project.  That's a _good_ thing, because they point out places where we now need to make decisions about the lifetime of these objects, whereas previously, there was no enforcement and it was easy for a programmer to accidentally not delete some memory.  By now I hope you can see that there is much more to this issue than is addressed by this change.


I do think we, the repository owners, did drop the ball in that we weren't clear on the direction we wanted to take the engine.  `PRAT`, `createrat`, `destroyrat`, et al. are remnants from the original calc engine which was C code.  There were no classes and destructors, so functions were written to create and destroy memory.  The desire we have is to eliminate PRAT and PNUMBER entirely, replacing them with Rational and Number.  You can see that, from what we have so far, we try to use Rational and Number as much as possible but the true functionality of the engine is still implemented with these old types.  Our hope for the engine is that the functionality can be moved entirely into the Rational and Number classes and that we can delete `PRAT` and `PNUMBER` from the codebase entirely.  A side-effect of that deletion would be that this issue of memory management is solved, as the classes will naturally have their memory managed by stack allocation and destruction.


Regarding this PR, given that it doesn't fix the issue and the desired change involves much more work, I'm going to politely close the PR for now.  I encourage you to ask questions on the issue and we can work together to make the right change for the codebase.  I think I'll need to do a followup on the issue to more clearly capture our roadmap for evolving the engine.  The referenced issue regarding smart pointers is too vague and doesn't capture the true intent.

PR #368, opened 2019-03-24
Improve the support of Narrator with parenthesis

### LINKED ISSUE #367: The narrator doesn't convey the number of open parenthesis when the user uses a keyboard
**Describe the bug**
- The narrator only conveys the number of open parenthesis when the button is pressed (via tab+space or using mouse/touch screen) but not when users type '('  or ')' via a keyboard.
- Minor issue: Each time the button is pressed, the narrator will say again the name of the button (and grammatically incorrectly): "Left parenthesis - open parenthesis count 5 - button"

**Steps To Reproduce**
- Activate the Narrator
- Launch the Calculator in Scientific or Programmer mode
- Type '('
--> the number of parenthesis won't be convey by the narrator.

**Expected behavior**
- The narrator should communicate the number of parenthesis when the button is pressed
- The Narrator should say _"Left parenthesis button - open parenthesis count 5"_ instead of _"Left parenthesis - open parenthesis count 5 - button"_ when the focus is set on the button
- The Narrator should not say the name of the button again when pressed, but only "open parenthesis count 5"
 

 - OS Build: 10.0.18361.0
 - Architecture: X64

rudyhuyn:
- Modify how the opening parenthesis button manages the Narrator to only convey the name of the `(` or `)` button when they get the focus
- Narrator will now say how many parenthesis are still open when users type '(' or ')'
- Modify and simplify how the right parenthesis button supports Narrator, we should limit registrations to `ViewModel::PropertyChanged` when possible, ideally, for performance reasons, we should only use bindings or specialized events. 
- Remove `Format_OpenParenthesisAutomationNamePrefix` to unify the behavior with the right parenthesis button and not repeat the name of the button each time the user clicks on it.

- tested manually

### DIFF
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -632,7 +632,7 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
         // Set the "(=xx" indicator.
         if (nullptr != m_pCalcDisplay)
         {
-            m_pCalcDisplay->SetParenthesisNumber(m_openParenCount >= 0 ? (unsigned int)m_openParenCount : 0);
+            m_pCalcDisplay->SetParenthesisNumber(m_openParenCount >= 0 ? static_cast<unsigned int>(m_openParenCount) : 0);
         }
 
         if (!m_bError)
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -83,7 +83,7 @@ void CCalcEngine::ClearTemporaryValues()
     m_bError = false;
 }
 
-void CCalcEngine::ProcessCommand(WPARAM wParam)
+void CCalcEngine::ProcessCommand(WPARAM wParam, bool useNarrator)
 {
     if (wParam == IDC_SET_RESULT)
     {
@@ -91,10 +91,10 @@ void CCalcEngine::ProcessCommand(WPARAM wParam)
         m_bSetCalcState = true;
     }
 
-    ProcessCommandWorker(wParam);
+    ProcessCommandWorker(wParam, useNarrator);
 }
 
-void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
+void CCalcEngine::ProcessCommandWorker(WPARAM wParam, bool useNarrator)
 {
     INT            nx, ni;
 
@@ -397,7 +397,7 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
         cleared for CENTR */
         if (nullptr != m_pCalcDisplay)
         {
-            m_pCalcDisplay->SetParenDisplayText(L"");
+            m_pCalcDisplay->SetParenDisplayText(0, false);
             m_pCalcDisplay->SetExpressionDisplay(make_shared<CalculatorVector<pair<wstring, int>>>(), make_shared<CalculatorVector<shared_ptr<IExpressionCommand>>>());
         }
 
@@ -440,7 +440,7 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
             }
             // automatic closing of all the parenthesis to get a meaningful result as well as ensure data integrity
             m_nTempCom = m_nLastCom; // Put back this last saved command to the prev state so ) can be handled properly
-            ProcessCommand(IDC_CLOSEP);
+            ProcessCommand(IDC_CLOSEP, useNarrator);
             m_nLastCom = m_nTempCom; // Actually this is IDC_CLOSEP
             m_nTempCom = (INT)wParam; // put back in the state where last op seen was IDC_CLOSEP, and current op is IDC_EQU
         }
@@ -632,7 +632,7 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
         // Set the "(=xx" indicator.
         if (nullptr != m_pCalcDisplay)
         {
-            m_pCalcDisplay->SetParenDisplayText(m_openParenCount ? to_wstring(m_openParenCount) : L"");
+            m_pCalcDisplay->SetParenDisplayText(m_openParenCount >= 0 ? (unsigned int)m_openParenCount : 0, useNarrator);
         }
 
         if (!m_bError)
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -83,7 +83,7 @@ void CCalcEngine::ClearTemporaryValues()
     m_bError = false;
 }
 
-void CCalcEngine::ProcessCommand(WPARAM wParam, bool useNarrator)
+void CCalcEngine::ProcessCommand(WPARAM wParam)
 {
     if (wParam == IDC_SET_RESULT)
     {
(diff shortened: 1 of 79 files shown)


src/Calculator/Views/CalculatorScientificOperators.xaml.cpp
```diff
@@ -107,10 +99,11 @@ void CalculatorScientificOperators::SetOperatorRowVisibility()
     InvRow2->Visibility = invRowVis;
 }
 
-void CalculatorScientificOperators::OnViewModelPropertyChanged(Object^ sender, PropertyChangedEventArgs^ e)
+void CalculatorScientificOperators::OpenParenthesisButton_GotFocus(Platform::Object^ sender, Windows::UI::Xaml::RoutedEventArgs^ e)
```
joseartrivera: Use RoutedEventArgs instead of Windows::UI::Xaml::RoutedEventArgs in cpp files.

src/Calculator/Views/CalculatorScientificOperators.xaml.cpp
```diff
@@ -107,10 +99,11 @@ void CalculatorScientificOperators::SetOperatorRowVisibility()
...
 {
-    if (e->PropertyName == StandardCalculatorViewModel::OpenParenthesisCountPropertyName && closeParenthesisButton->FocusState != ::FocusState::Unfocused)
-    {
-        Model->SetOpenParenthesisCountNarratorAnnouncement();
-    }
+    Model->SetOpenParenthesisCountNarratorAnnouncement();
+}
+
+String^ CalculatorScientificOperators::ParenthesisCountToString(unsigned int count) {
+    return count == 0 ? ref new String(L"") : ref new String(to_wstring(count).data());
```
joseartrivera: Is there a constant we could use here instead for empty string?

joseartrivera: I think it is more readable if you surround (count == 0) in parenthesis.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -100,10 +98,12 @@ void CalculatorProgrammerRadixOperators::IsErrorVisualState::set(bool value)
     }
 }
 
-void CalculatorProgrammerRadixOperators::OnViewModelPropertyChanged(Object^ sender, PropertyChangedEventArgs^ e)
+String^ CalculatorProgrammerRadixOperators::ParenthesisCountToString(unsigned int count) {
+    return count == 0 ? ref new String(L"") : ref new String(to_wstring(count).data());
+}
+
+
+void CalculatorProgrammerRadixOperators::CalculatorProgrammerRadixOperators::OpenParenthesisButton_GotFocus(Platform::Object^ sender, Windows::UI::Xaml::RoutedEventArgs^ e)
```
joseartrivera: Here and in the rest of the cpp files, you can just use Object^ instead of the full Platform::Object^

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -100,10 +98,12 @@ void CalculatorProgrammerRadixOperators::IsErrorVisualState::set(bool value)
     }
 }
 
-void CalculatorProgrammerRadixOperators::OnViewModelPropertyChanged(Object^ sender, PropertyChangedEventArgs^ e)
+String^ CalculatorProgrammerRadixOperators::ParenthesisCountToString(unsigned int count) {
+    return count == 0 ? ref new String(L"") : ref new String(to_wstring(count).data());
```
joseartrivera: Same feedback as the same line in CalculatorScientificOperators.xaml.cpp

src/Calculator/Resources/am-ET/Resources.resw
```diff
@@ -901,10 +901,7 @@
     <value>የግራ ቅንፍ</value>
     <comment>Screen reader prompt for the Calculator "(" button on the scientific operator keypad</comment>
   </data>
-  <data name="Format_OpenParenthesisAutomationNamePrefix" xml:space="preserve">
```
joseartrivera: You shouldn't need to remove this for every language, just the en-us resw, the rest should be removed automatically by our tooling if I am not mistaken. @mcooley can you double check me here?

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -216,19 +214,21 @@ void StandardCalculatorViewModel::DisplayPasteError()
     m_standardCalculatorManager.DisplayPasteError();
 }
 
-void StandardCalculatorViewModel::SetParenthesisCount(_In_ const wstring& parenthesisCount)
+void StandardCalculatorViewModel::SetParenthesisCount(_In_ unsigned int parenthesisCount)
 {
+    if (m_OpenParenthesisCount == parenthesisCount)
```
joseartrivera: Add curly braces around single line if statements

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -632,7 +632,7 @@ void CCalcEngine::ProcessCommandWorker(WPARAM wParam)
         // Set the "(=xx" indicator.
         if (nullptr != m_pCalcDisplay)
         {
-            m_pCalcDisplay->SetParenDisplayText(m_openParenCount ? to_wstring(m_openParenCount) : L"");
+            m_pCalcDisplay->SetParenthesisNumber(m_openParenCount >= 0 ? (unsigned int)m_openParenCount : 0);
```
joseartrivera: Use static_cast instead

src/Calculator/Resources/am-ET/Resources.resw
mcooley: @joseartrivera yep, any changes in the non-English resw files will just get overwritten by the localization tools. But there's no real harm in modifying the files if you need to validate changes.

src/Calculator/Views/CalculatorScientificOperators.xaml.cpp
rudyhuyn: Not that I know of. But we can simplify and use `ref new String()`

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
rudyhuyn: cpp files are full of `Windows::UI::Xaml::RoutedEventArgs` and `Platform::Object` (because of the XAML editor), I will do a full cleaning and remove all of them

src/Calculator/Resources/am-ET/Resources.resw
rudyhuyn: I prefer to remove all of them (via a regex, it takes only 10 seconds) so I can test with other languages (in fact, I randomly dev with the app in French, Spanish or one of the RtL language) to be sure to be sure the app is fully localized and support all types of region settings

src/CalcViewModel/StandardCalculatorViewModel.cpp
rudyhuyn: 👍 I should really add a rule in VS, working on many projects with different code formatting rules is confusing :D Thank you for checking!

rudyhuyn:
@joseartrivera let me know if you need other changes!

joseartrivera:
Thanks Rudy, this change is a great improvement!

PR #369, opened 2019-03-24
Keep input when users refresh currency rates

### LINKED ISSUE #260: Inputs are lost after clicking the "Update rates" button.
**Describe the bug**
Converter doesn't keep inputs after clicking the "Update rates" button.

**Steps To Reproduce**
1. Launch the calculator
2. Open the currency converter
3. Select currency pair and enter any value for base unit.
4. Click on "Update rates"

**Expected behavior**
It would be nice if the converter keeps the inputs when you click the "Update rates" button or recalculate result automatically (based on from unit) if ratio have changed. This will prevent re-entering the same amount again.

**Screenshots**
kLkmVLrInx

 - OS Build: 17763
 - Architecture: X64

rudyhuyn:
- prevent `UnitConverterViewModel` to reset values when users click on `update rates`.
- recompute UnitConverter's caches (`m_ratioMap` and `m_categoryToUnits`) once rates are updated (but check first if the user did/didn't change the category)

- Manually tested with fake currency rates (HTTP responses modified on the fly via FiddlerCore)
- Verified that it works no matter the selected field (`From` or `To`)
- Verified that the currencies selected are kept after a refresh


src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -152,11 +152,14 @@ UnitConverterViewModel::UnitConverterViewModel(const shared_ptr<UCM::IUnitConver
     PopulateData();
 }
 
-void UnitConverterViewModel::ResetView()
+void UnitConverterViewModel::ResetView(bool resetValues)
```
danbelcher-MSFT: I'm not really a fan of this change because it leads to awkward code like `Model->RefreshCurrencyRatios(false);`  Are we refreshing the currenty ratios or not?  Now the casual reader needs to know what `true`/`false` mean in the context of this function and overall clarity of the code is reduced.

My suggestion would be to write `void UnitConverterViewModel::ResetValues()` and call it where appropriate.

danbelcher-MSFT: Same opinion for `UpdateCategory`.

danbelcher-MSFT: Ping @rudyhuyn

rudyhuyn: I totally forgot this PR, I will update it tonight!

rudyhuyn: I totally refactored this PR without the boolean, I agree, it's better this way!

danbelcher-MSFT:
@rudyhuyn, looks like UTs are failing.

src/CalcManager/UnitConverter.h
```diff
@@ -228,7 +230,7 @@ namespace UnitConversionManager
         bool CheckLoad();
         double Convert(double value, ConversionData conversionData);
         std::vector<std::tuple<std::wstring, Unit>> CalculateSuggested();
-        void Reset();
+        void ResetCategoriesAndRatios();
```
danbelcher-MSFT: `IUnitConverter` declares all of its pure virtual functions as public.  In `UnitConverter`, some of those overriden functions are public and some are private.  While not technically an error according to the standard, it's at least strange.  There is an area above where most of the `IUnitConverter` overrides are publicly declared and grouped.  I'd appreciate if you moved the private declarations to that group and added the `override` specifier.

PR #381, opened 2019-03-25
[Build] Don't run component governance detection in postbuild steps

mcooley:
See also #353.

In the internal build environment, there's an auto-injected component governance task which needs to run once during the build. This task doesn't need to run during the unit test, package, and internal release jobs.

### DIFF
--- a/build/pipelines/templates/package-appxbundle.yaml
+++ b/build/pipelines/templates/package-appxbundle.yaml
@@ -20,6 +20,8 @@ jobs:
     vmImage: vs2017-win2016
   workspace:
     clean: outputs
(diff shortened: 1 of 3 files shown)


HowardWolosky:
@mcooley - with this change, in which pipelines would Component Governance still run?

mcooley:
@HowardWolosky The release pipeline (internally, "Calculator-Release") contains multiple jobs. The CG task was getting injected into every job--so it would run 9x during a typical pipeline run. After this change it will still run in Calculator-Release, but only in the _build_ jobs (so, once per architecture).

PR #390, opened 2019-03-26
Updates to copypastemanager

### LINKED ISSUE #55: Calculator should use constexpr when defining static const strings
Many strings in Calculator should not be modified and so we declare them 'static const'.  This is fine for ensuring compiler-enforced const-ness, but a downside to this approach is that the strings aren't created until runtime.  We can improve this by marking the strings with 'constexpr'.  This will tell the compiler to bake the strings into the binary itself, which results in less overhead.

Nicholas-Baron:
- Added `constexpr` to formerly `static const` or `#define` variables
- Applied C++ Core Guideline NR.2 
- Added `auto` and `const` in appropriate places

- Used the provided unit tests

### DIFF
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -418,8 +418,7 @@ bool CopyPasteManager::TryOperandToULL(const wstring& operand, int numberBase, u
         return false;
     }
 
-    // Default to base10
-    int intBase = 10;
+    int intBase;
     switch (numberBase)
     {
     case HexBase:
@@ -431,7 +430,8 @@ bool CopyPasteManager::TryOperandToULL(const wstring& operand, int numberBase, u
     case BinBase:
         intBase = 2;
         break;
-    case DecBase: // Unneeded as the default is already 10?
+    default:
+    case DecBase:
         intBase = 10;
         break;
     }
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -429,7 +429,7 @@ bool CopyPasteManager::TryOperandToULL(const wstring& operand, int numberBase, u
     case BinBase:
         intBase = 2;
         break;
-    case DecBase:
+    case DecBase: // Unneeded as the default is already 10?
         intBase = 10;
         break;
     }
@@ -445,6 +445,10 @@ bool CopyPasteManager::TryOperandToULL(const wstring& operand, int numberBase, u
         // Do nothing
     }
     catch (out_of_range)
+    {
+        // Do nothing
+    }
+    catch(...)
     {
         // Do nothing
     }
@@ -472,13 +476,13 @@ size_t CopyPasteManager::OperandLength(wstring operand, ViewMode mode, CategoryG
 
 size_t CopyPasteManager::StandardScientificOperandLength(wstring operand)
 {
+    // Loop until a decimal is found or if the end is reached
     bool hasDecimal = false;
-    for (size_t i = 0; i < operand.length(); i++)
+    for (size_t i = 0; i < operand.length() && !hasDecimal; i++)
     {
         if (operand[i] == L'.')
         {
             hasDecimal = true;
-            break;
         }
     }
 
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -340,14 +340,16 @@ bool CopyPasteManager::ExpressionRegExMatch(vector<wstring> operands, ViewMode m
 
 pair<size_t, uint64_t> CopyPasteManager::GetMaxOperandLengthAndValue(ViewMode mode, CategoryGroupType modeType, int programmerNumberBase, int bitLengthType)
 {
+    constexpr size_t defaultMaxOperandLength = 0;
+    constexpr uint64_t defaultMaxValue = 0;
     
     if (mode == ViewMode::Standard)
     {
-        return make_pair(MaxStandardOperandLength, 0);
+        return make_pair(MaxStandardOperandLength, defaultMaxValue);
     }
     else if (mode == ViewMode::Scientific)
     {
-        return make_pair(MaxScientificOperandLength, 0);
+        return make_pair(MaxScientificOperandLength, defaultMaxValue);
     }
     else if (mode == ViewMode::Programmer)
     {
@@ -394,10 +396,10 @@ pair<size_t, uint64_t> CopyPasteManager::GetMaxOperandLengthAndValue(ViewMode mo
     }
     else if (modeType == CategoryGroupType::Converter)
     {
-        return make_pair(MaxConverterInputLength, 0);
+        return make_pair(MaxConverterInputLength, defaultMaxValue);
     }
 
-    return make_pair(0, 0);
+    return make_pair(defaultMaxOperandLength, defaultMaxValue);
 }
 
 wstring CopyPasteManager::SanitizeOperand(const wstring& operand)
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -15,8 +15,6 @@ using namespace Windows::Foundation;
 using namespace Windows::System;
 using namespace Windows::ApplicationModel::DataTransfer;
 
-unsigned long long maxOperandNumber;
-
 String^ CopyPasteManager::supportedFormats[] =
 {
     StandardDataFormats::Text
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -106,19 +106,16 @@ task<String^> CopyPasteManager::GetStringToPaste(ViewMode mode, CategoryGroupTyp
 
 int CopyPasteManager::ClipboardTextFormat()
 {
-    int result = -1;
-
-    auto dataPackageView = Clipboard::GetContent();
+    const auto dataPackageView = Clipboard::GetContent();
 
     for (int i = 0; i < RTL_NUMBER_OF(supportedFormats); i++)
     {
         if (dataPackageView->Contains(supportedFormats[i]))
         {
-            result = i;
-            break;
+            return i;
         }
     }
-    return result;
+    return -1;
 }
 
 String^ CopyPasteManager::ValidatePasteExpression(String^ pastedText, ViewMode mode, int programmerNumberBase, int bitLengthType)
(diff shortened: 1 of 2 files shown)


src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -15,28 +15,29 @@ using namespace Windows::Foundation;
...
+static const auto c_signedDecFloat = L"[-+]?\\d*(\\d|[.])\\d*";
 
 // Programmer Mode Integer patterns
 // Support digit separators ` (WinDbg/MASM), ' (C++), and _ (C# and other languages)
-static const wstring c_hexProgrammerChars = L"([a-f]|[A-F]|\\d)+((_|'|`)([a-f]|[A-F]|\\d)+)*";
-static const wstring c_decProgrammerChars = L"\\d+((_|'|`)\\d+)*";
-static const wstring c_octProgrammerChars = L"[0-7]+((_|'|`)[0-7]+)*";
-static const wstring c_binProgrammerChars = L"[0-1]+((_|'|`)[0-1]+)*";
-static const wstring c_uIntSuffixes = L"[uU]?[lL]{0,2}";
+static constexpr auto c_hexProgrammerChars = L"([a-f]|[A-F]|\\d)+((_|'|`)([a-f]|[A-F]|\\d)+)*";
```
rudyhuyn: Why not `const` like previous lines?

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -15,28 +15,29 @@ using namespace Windows::Foundation;
...
 // [\s\x85] means white-space characters
 static const wstring c_wspc = L"[\\s\\x85]*";
-static const wstring c_wspcLParens = c_wspc + L"[(]*" + c_wspc;
-static const wstring c_wspcLParenSigned = c_wspc + L"([-+]?[(])*" + c_wspc;
-static const wstring c_wspcRParens = c_wspc + L"[)]*" + c_wspc;
-static const wstring c_signedDecFloat = L"[-+]?\\d*(\\d|[.])\\d*";
+static const auto c_wspcLParens = c_wspc + L"[(]*" + c_wspc;
+static const auto c_wspcLParenSigned = c_wspc + L"([-+]?[(])*" + c_wspc;
+static const auto c_wspcRParens = c_wspc + L"[)]*" + c_wspc;
+static const auto c_signedDecFloat = L"[-+]?\\d*(\\d|[.])\\d*";
```
rudyhuyn: It's probably better to not use `auto` for this kind of variables, it can create some confusions and can introduce some issues.

A good example is these 4 lines, `c_wspcLParens`, `c_wspcLParenSigned`, `c_wspcRParens` are `wstring` while `c_signedDecFloat` is now a `const wchar_t[]`

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -449,33 +448,37 @@ bool CopyPasteManager::TryOperandToULL(const wstring& operand, int numberBase, u
...
 
 size_t CopyPasteManager::OperandLength(wstring operand, ViewMode mode, CategoryGroupType modeType, int programmerNumberBase)
 {
-    size_t len = 0;
     if (mode == ViewMode::Standard || mode == ViewMode::Scientific)
     {
-        len = StandardScientificOperandLength(operand);
+        return StandardScientificOperandLength(operand);
     }
     else if (mode == ViewMode::Programmer)
```
rudyhuyn: We should probably use a `switch` to improve the readability of the code

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -449,33 +448,37 @@ bool CopyPasteManager::TryOperandToULL(const wstring& operand, int numberBase, u
...
-    return len;
+    return 0;
 }
 
 size_t CopyPasteManager::StandardScientificOperandLength(wstring operand)
 {
+    // Loop until a decimal is found or if the end is reached
     bool hasDecimal = false;
-    for (size_t i = 0; i < operand.length(); i++)
+    for (size_t i = 0; i < operand.length() && !hasDecimal; i++)
```
rudyhuyn: 👏🏻👏🏻👏🏻👏🏻👏🏻

rudyhuyn: `hasDecimal = operand.find('.') !=std::string::npos;` ?

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -341,16 +340,14 @@ bool CopyPasteManager::ExpressionRegExMatch(vector<wstring> operands, ViewMode m
...
 pair<size_t, uint64_t> CopyPasteManager::GetMaxOperandLengthAndValue(ViewMode mode, CategoryGroupType modeType, int programmerNumberBase, int bitLengthType)
 {
-    size_t maxLength = 0;
-    uint64_t maxValue = 0;
-
+    
     if (mode == ViewMode::Standard)
     {
-        maxLength = MaxStandardOperandLength;
+        return make_pair(MaxStandardOperandLength, 0);
```
rudyhuyn: 👏🏻👏🏻👏🏻👏🏻

src/CalcViewModel/Common/CopyPasteManager.cpp
Nicholas-Baron: Latest commit on this branch accomplishes this.

src/CalcViewModel/Common/CopyPasteManager.cpp
Nicholas-Baron: Change applied in latest commit to this branch.

Nicholas-Baron:
@rudyhuyn The changes that were suggested are now in the code.

src/CalcViewModel/Common/CopyPasteManager.h
```diff
@@ -13,15 +13,15 @@ namespace CalculatorUnitTests
...
-#define QwordType 1
-#define DwordType 2
-#define WordType 3
-#define ByteType 4
-#define HexBase 5
-#define DecBase 6
-#define OctBase 7
-#define BinBase 8
+    // TODO: Could be made into an enum : unsigned ?
+    constexpr auto QwordType = 1;
```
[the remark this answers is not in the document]
Nicholas-Baron: Could these instead be `static`? C++17 `static constexpr` implies `inline`. Source: [cppreference](https://en.cppreference.com/w/cpp/language/constexpr)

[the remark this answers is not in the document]
danbelcher-MSFT: Are you referring to:
> A constexpr specifier used in a function or static member variable (since C++17) declaration implies inline.

?  The use here doesn't fit either case.  We need to look at the documentation for [storage duration](https://en.cppreference.com/w/cpp/language/storage_duration).  Here are the relevant parts, shortened for brevity:
> **internal linkage.** The name can be referred to from all scopes in the current translation unit.
> Any of the following names declared at namespace scope have internal linkage:
> - variables, functions, or function templates declared static;
> - non-volatile non-inline (since C++17) const-qualified variables (including constexpr) that aren't declared extern and aren't previously declared to have external linkage;

The second bullet point describes the proposed change (`constexpr auto QwordType = 1;`).  This variable has internal linkage, meaning each translation unit has its own copy of the definition.  The first bullet point describes the addition you suggested (`static constexpr auto QwordType = 1;`).  The variable already has internal linkage so there is no difference between the two.

The goal for the codebase is that there is 1 definition for these expressions and so the appropriate keyword is `inline`.  This will force all references to the variable across translation units to actually refer to the same definition.  We can test all of this with a simple program that prints the address of these expressions in two different translation units.  I left out some of the headers just to keep the example simple.
```C++
// Common.h
constexpr auto var_constexpr = 1;
static constexpr auto var_static_constexpr = 2;
inline constexpr auto var_inline_constexpr = 3;

// Source1.cpp
void Source1::print()
{
    cout << "Source1::print\n";
    cout << "\tconstexpr: " << &var_constexpr << "\n";
    cout << "\tinline constexpr: " << &var_inline_constexpr << "\n";
    cout << "\tstatic constexpr: " << &var_static_constexpr << "\n";
    cout << endl;
}

// Source2.cpp
// same as above

// Main.cpp
int main()
{
    Source1::print();
    Source2::print();
}
```
```
Source1::print
        constexpr: 00329B30
        static constexpr: 00329B34
        inline constexpr: 00329B3C

Source2::print
        constexpr: 00329B98
        static constexpr: 00329B9C
        inline constexpr: 00329B3C
```
As discussed, only `inline constexpr` forced the single definition across the program.  This explanation was rather long, but the topic comes up often and I have a sample program I keep on hand :)  Linkage is a common source of confusion in C++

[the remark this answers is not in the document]
Nicholas-Baron: Yeah, I have heard that linkers are not as frequently updated as other parts of a compiler.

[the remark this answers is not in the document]
danbelcher-MSFT: A member variable that is `static constexpr` is implicitly `inline`, therefore, there would be no difference.  `inline` means that "multiple definitions are permitted".  Imagine that it were _not_ the case that `static constexpr` implies `inline` for a member variable.  Then consider:
```C++
// Class.h
class Foo {
    static constexpr bar = 1;
};

// Source1.cpp
#include "Class.h"

// Source2.cpp
#include "Class.h"
```
Each cpp is its own translation unit and, therefore, each provides a definition for `bar`.  Without `inline`, implied or explicit, then the duplicate definition would be a violation of the One Definition Rule.

According to [`inline`](https://en.cppreference.com/w/cpp/language/inline), "A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.".  This case is describing:
```C++
class Foo2 {
    void print() { std::cout << "hello!"; }
};
```

As with the scenario above, if the member function definition was included into multiple translation units, there would be an ODR-violation, therefore member function definitions are implicitly `inline` as well.

src/CalcViewModel/Common/CopyPasteManager.h
```diff
@@ -13,27 +13,26 @@ namespace CalculatorUnitTests
...
+    inline constexpr auto OctBase = 7;
+    inline constexpr auto BinBase = 8;
 
     class CopyPasteManager
     {
     public:
         static void CopyToClipboard(Platform::String^ stringToCopy);
         static concurrency::task<Platform::String^> GetStringToPaste(CalculatorApp::Common::ViewMode mode, CalculatorApp::Common::CategoryGroupType modeType, int programmerNumberBase = -1, int bitLengthType = -1);
-        static bool HasStringToPaste()
+        inline static bool HasStringToPaste()
```
danbelcher-MSFT:
```suggestion
        static bool HasStringToPaste()
```
According to [`inline`](https://en.cppreference.com/w/cpp/language/inline):
> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

src/CalcViewModel/Common/CopyPasteManager.h
```diff
@@ -13,27 +13,26 @@ namespace CalculatorUnitTests
...
         {
             return ClipboardTextFormat() >= 0;
         }
 
-        static constexpr auto PasteErrorString = L"NoOp";
+        inline static constexpr auto PasteErrorString = L"NoOp";
```
danbelcher-MSFT:
```suggestion
        static constexpr auto PasteErrorString = L"NoOp";
```
According to [`inline`](https://en.cppreference.com/w/cpp/language/inline):
> A static member variable (but not a namespace-scope variable) declared constexpr is implicitly an inline variable.

The exception clause here ("but not a namespace-scope variable) is why we needed to explicitly add `inline` for `QwordType` above.  They are namespace-scope variables so they are not implicitly `inline`.

PR #400, opened 2019-03-27
Code analysis 3

ariccio:
Retry of #315


janisozaur:
@ariccio why do you keep closing and submitting new PRs?

src/CalcManager/Ratpack/num.cpp
```diff
@@ -192,9 +192,9 @@ void _addnum( PNUMBER *pa, PNUMBER b, uint32_t radix)
 //
 //----------------------------------------------------------------------------
 
-void _mulnum( PNUMBER *pa, PNUMBER b, uint32_t radix);
+void _mulnum( _Inout_ PNUMBER *pa, _In_ _Const_ PNUMBER b, uint32_t radix);
```
janisozaur: What's the purpose of `_Const_` here? What benefit does it bring over `const`?

src/CalcManager/Ratpack/rat.cpp
```diff
@@ -99,8 +99,8 @@ void fracrat( PRAT *pa , uint32_t radix, int32_t precision)
 //
 //-----------------------------------------------------------------------------
 
-void mulrat( PRAT *pa, PRAT b, int32_t precision)
-
+void mulrat( _Inout_ PRAT *pa, _In_ PRAT b, int32_t precision)
+
```
janisozaur:
```suggestion
```
Please don't introduce unnecessary whitespace

src/CalcManager/Ratpack/rat.cpp
```diff
@@ -263,8 +263,8 @@ void addrat( PRAT *pa, PRAT b, int32_t precision)
 //
 //-----------------------------------------------------------------------------
 
-void rootrat( PRAT *py, PRAT n, uint32_t radix, int32_t precision)
-{
+void rootrat( _Inout_ PRAT *py, _In_ PRAT n, uint32_t radix, int32_t precision)
+{
```
janisozaur:
```suggestion
{
```

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -373,7 +373,7 @@ PNUMBER nRadixxtonum( _In_ PNUMBER a, uint32_t radix, int32_t precision)
 //
 //-----------------------------------------------------------------------------
 
-PNUMBER numtonRadixx(_In_ PNUMBER a, uint32_t radix)
+PNUMBER numtonRadixx(PNUMBER a, uint32_t radix)
```
janisozaur: I'm surprised by this change, was this intended?

src/CalcManager/Ratpack/num.cpp
ariccio: I don't want to change the interface of _mulnum, and (IIRC) it does actually modify it a few times within the function, before returning it to the original state.

src/CalcManager/Ratpack/conv.cpp
ariccio: Honestly, I'm not 100% sure, but I think it was because `a` is actually modified.

janisozaur: Wouldn't that make it `_Inout_`?

src/CalcManager/Ratpack/num.cpp
janisozaur: That makes no sense. Either it is const, in which case it should be marked so with `const` or it isn't and you shouldn't try faking it.

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -226,8 +226,8 @@ void lograt( PRAT *px, int32_t precision)
     destroyrat(offset);
     destroyrat(pwr);
 }
-
-void log10rat( PRAT *px, int32_t precision)
+
```
janisozaur:
```suggestion

```
Can't you just strip all the modified files?

src/CalcManager/Ratpack/num.cpp
ariccio: I think you're correct here.

src/CalcManager/Ratpack/conv.cpp
ariccio: In this codebase, I think they intend to treat `PNUMBER` and such as conceptually `_In_` arguments, since they usually pass parameters that are conceptually modified as pointer-to-pointer (`PNUMBER*`) or reference (`PNUMBER&`). I'm going to recompile with this as an `_In_` argument and see if it complains.

ariccio: Doesn't seem to complain, will return to original.

src/CalcViewModel/CalcViewModel.vcxproj
```diff
@@ -40,6 +40,7 @@
     <RootNamespace>CalcViewModel</RootNamespace>
     <DefaultLanguage>en-US</DefaultLanguage>
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+    <PreferredToolArchitecture>x64</PreferredToolArchitecture>
```
danbelcher-MSFT: According to the [docs](https://docs.microsoft.com/en-us/cpp/build/walkthrough-using-msbuild-to-create-a-visual-cpp-project), this is supposed to appear _after_ the import of `Microsoft.Cpp.default.props` and here you have it before.  You can take a look at how CalcManager sets this up:
https://github.com/Microsoft/calculator/blob/ad25feda6bfac4b4779b92500b01d341931dcbb0/src/CalcManager/CalcManager.vcxproj#L51-L55

src/CalcViewModel/Common/Utils.h
```diff
@@ -284,8 +284,8 @@ namespace Utils
 
     void IFTPlatformException(HRESULT hr);
     Platform::String^ GetStringValue(Platform::String^ input);
-    bool IsLastCharacterTarget(std::wstring const &input, wchar_t target);
-    std::wstring RemoveUnwantedCharsFromWstring(std::wstring inputString, wchar_t* unwantedChars, unsigned int size);
+    bool IsLastCharacterTarget(_In_ std::wstring const &input, wchar_t target);
```
danbelcher-MSFT:
```suggestion
    bool IsLastCharacterTarget(std::wstring const &input, wchar_t target);
```

danbelcher-MSFT:
Ping @ariccio

ariccio:
Sorry, I can't really put a whole lot of time into this for the next week or so - I'm quite busy -

danbelcher-MSFT:
No problem!  This improvement can come in when you're ready.

mcooley:
Many of the improvements here made it into #689.

I'm closing out this PR since it's been inactive for a while, but I know the code here will be a useful reference if you or others want to enable/fix more code analysis warnings in the future. We're tracking that work as #693 now. Thanks for contributing to Calculator!

PR #403, opened 2019-03-28
Decrease CPU usage of OverflowTextBlock + optimization

### LINKED ISSUE #402: Clicking negate more than ten times progressively slows UI responsiveness
I am using Calculator version 10.1812.10048.0, the latest version at the time of this writing.

**Describe the bug**
When I click on the negate button (on the bottom left hand corner) more than 10 times, the UI of the calculator becomes slower and slower to use. The more I click on it, the slower and slower the Calculator gets, until it gets so slow it is unusable.

**Steps To Reproduce**
1. Open Calculator
2. Switch to Standard mode
3. Click the negate icon (the +-) thirty or more times
4. Hover over another digit on the calculator and click it. The digit will not appear in the calculator window until a few seconds later

**Expected behavior**
Clicking another digit should be almost instantaneous after clicking negate thirty or more times.

**Screenshots**

 - OS Build: 10.0.17134.0
 - Architecture: X64

**Additional context**

rudyhuyn:
## Fixes #402 and #414 

Divide by 4 the CPU usage of `OverflowTextBlock` when buttons are pressed very quickly.

Snag_e26b41

**Xaml-side:**
- `OverflowTextBlock` has some performance issues: 
    - double scrollviewer: the listview was in a scrollviewer, while the control already containing one -> it breaks the virtualization of the listview and impacts on UI performance.
    - The listview used a `StackPanel`, this panel doesn't support virtualization of ListViewItems contrary to  `ItemsStackPanel`
    - No ListView-specific features were used, an `ItemsControl` is more efficient and lighter.
- refactor how we manage the visibility of the left/right buttons in OverflowTextBlock, the new version is more reactive and will not display the right arrow when not necessary (see GIF below).
- remove the ItemContainerSelector `ExpressionItemContainerStyle`, not really used by `OverflowTextBlock`
- remove UI glitches generated by ChangeView when users type fast (control partially hidden and scrolling issues, see the GIF below). 
- only modify the accessibility view when it's necessary

**ViewModel-side:**
- stop fully refreshing `ExpressionTokens` in `StandardCalculatorViewModel` when a new command were sent, instead, use a IObservableVector to only send new tokens to the UI (in average only 1 or 2 UI items are refreshed while the full expression was refreshed before)

Demo:
A0D05EB6-A342-4AD9-BE6C-D4FD51E97733

Manually


src/Calculator/Views/Calculator.xaml
```diff
@@ -19,20 +18,20 @@
             <TextBlock Margin="2,0,0,0"
                        Foreground="{ThemeResource SystemControlPageTextBaseMediumBrush}"
                        IsTextScaleFactorEnabled="False"
-                       Text="{Binding Token, Mode=OneWay}"/>
+                       Text="{x:Bind Token}"/>
```
danbelcher-MSFT: These bindings were updated to be `OneTime` (default mode for `x:Bind`).  There are places in the codebase where the `Token` property is updated and those changes wouldn't show in the UI after this change.  Are you sure this is correct?

rudyhuyn: The only place I found where Token is modified is a dead code (the UWP Calculator doesn't allowing users to edit an operation contrary to the Win7 version) but while this dead code is still part of the repository, you are right, we should should keep `Mode=OneWay`.

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -1422,29 +1431,29 @@ void StandardCalculatorViewModel::SaveEditedCommand(_In_ unsigned int tokenPosit
...
-        case static_cast<int>(Command::CommandATANH) :
-            updatedToken = CCalcEngine::OpCodeToUnaryString(static_cast<int>(Command::CommandTANH), true, angleType);
-            break;
-        case static_cast<int>(Command::CommandPOWE) :
-            updatedToken = CCalcEngine::OpCodeToUnaryString(static_cast<int>(Command::CommandLN), true, angleType);
-            break;
-        default:
-            updatedToken = CCalcEngine::OpCodeToUnaryString(nOpCode, false, angleType);
+            case static_cast<int>(Command::CommandASIN) :
+                updatedToken = CCalcEngine::OpCodeToUnaryString(static_cast<int>(Command::CommandSIN), true, angleType);
```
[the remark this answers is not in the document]
rudyhuyn: 👍 Visual Studio is very annoying with that...

src/Calculator/Views/Calculator.xaml
danbelcher-MSFT: We could also remove the dead code 👍  Then this can be kept as a `OneTime`

src/Calculator/Controls/OverflowTextBlock.cpp
```diff
@@ -34,7 +34,7 @@ void OverflowTextBlock::OnApplyTemplate()
 
     m_expressionContainer = safe_cast<ScrollViewer^>(GetTemplateChild("expressionContainer"));
     m_expressionContainer->ChangeView(m_expressionContainer->ExtentWidth - m_expressionContainer->ViewportWidth, nullptr, nullptr);
-
+    m_expressionContainer->ViewChanged += ref new Windows::Foundation::EventHandler<Windows::UI::Xaml::Controls::ScrollViewerViewChangedEventArgs ^>(this, &CalculatorApp::Controls::OverflowTextBlock::OnViewChanged);
```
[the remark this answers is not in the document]
rudyhuyn: Yes. I will also unregister the other events (`m_scrollRight->Click` and `m_scrollLeft->Click`) in case  `OnApplyTemplate` is called more than 1 time (it's not the case right now, but better to manage this scenario too)

src/Calculator/Controls/OverflowTextBlock.cpp
```diff
@@ -30,25 +30,38 @@ DEPENDENCY_PROPERTY_INITIALIZATION(OverflowTextBlock, TokensUpdated);
 
 void OverflowTextBlock::OnApplyTemplate()
 {
-    assert(((m_scrollLeft == nullptr) && (m_scrollRight == nullptr)) || ((m_scrollLeft != nullptr) && (m_scrollRight != nullptr)));
+    UnregisterEventHandlers();
```
rudyhuyn: slight refactoring to make this control more customizable (in case developers want to use it in their own apps):
- can apply a template than 1 time
- make controls optional

PR #405, opened 2019-03-28
Perform clang-format on Azure

janisozaur:
## Fixes (part of) #202.

#236 discusses the contents of `.clang-format`, this PR makes CI capable of enforcing it.

A problem I ran into was `clang-format` mistaking some of the sources for Objective-C, due to non-C++ constructs like https://github.com/Microsoft/calculator/blob/f6a6aae6e6ce1c485b4b4b02413259649cf2b58f/src/Calculator/Converters/BitFlipAutomationNameConverter.h#L13
I have no idea what that is, so I took the easy way out and only apply formatting files which name ends with `.cpp`.

I provided comments in the script itself, but the gist of it is:
* Using Hosted Ubuntu 16.04
* Downloads static build of `clang-format` from https://github.com/angular/clang-format/blob/master/bin/linux_x64/clang-format, a long standing repo with occasional updates to the tool
* Performs `clang-format` on the sources (files ending with `.cpp`)
* Reports the diff, if any, and
  - Fails the build, if diff present
  - Reports success otherwise.

You can see a failing build here: https://dev.azure.com/ms/calculator/_build/results?buildId=8367

And a passing one (with sources formatted) here: https://dev.azure.com/ms/calculator/_build/results?buildId=8372, as per #406 

This PR does not include `.clang-format`, I hope @seyfer will update https://github.com/Microsoft/calculator/pull/236 and it will get merged.

### DIFF
--- a/build/pipelines/azure-pipelines.ci.yaml
+++ b/build/pipelines/azure-pipelines.ci.yaml
@@ -18,6 +18,10 @@ jobs:
   parameters:
     platform: x64
 
+- template: ./templates/clang-format.yaml
+  parameters:
+    platform: x64
+
 - template: ./templates/build-app-public.yaml
   parameters:
     platform: x86
--- a/build/pipelines/templates/clang-format.yaml
+++ b/build/pipelines/templates/clang-format.yaml
@@ -1,0 +1,37 @@
+# Job for running clang-format on sources.
+
+parameters:
+  platform: ''
+  condition: ''
+
+jobs:
+- job: Format${{ parameters.platform }}
+  displayName: clang-format
+  pool:
+    vmImage: 'Ubuntu-16.04'
+
+  steps:
+  - script: |
+      # Download a recent, static build of clang-format. Be quiet, but don't hide errors.
+      wget --no-verbose https://github.com/angular/clang-format/raw/master/bin/linux_x64/clang-format
+      chmod a+x clang-format
+      # Make sure it identifies itself
+      ./clang-format --version
+      # Parse full lines
+      export IFS=$'\n'
+      # For each file in repository with name ending with ".cpp" ...
+      for file in $(git ls-files | grep \\.cpp$); do
+          # ... print its name, in case clang-format throws some errors ...
+          echo Formatting "$file"
+          # ... and format it in place, so git can pick it up.
+          ./clang-format -style=file -i "$file"
+      done
+      # Just some visual separation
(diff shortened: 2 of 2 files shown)


janisozaur:
There are tools like `git-clang-format` that make `clang-format` only test relevant git commits, but it only takes 11 seconds to download the utility, perform formatting of all the sources and report it back (23s if you count whole process of getting the VM, starting agent, etc.), so I deemed it an unnecessary complication.

rudyhuyn:
I would suggest to use the `clang-format.exe` already shipped with Visual Studio 15.7 and + for consistency

danbelcher-MSFT:
Can we set this up as a git hook so that source is automatically formatted?  We want our tools to work with us and help us take contributions faster.

janisozaur:
Setting up a git hook is a client-side action.

danbelcher-MSFT:
I see.  I haven't looked into this topic and I'm new to OSS development, so this isn't an area I'm very familiar with.  If this is typical process for other projects then we should follow convention.

janisozaur:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

> It’s important to note that client-side hooks are **not** copied when you clone a repository.

seyfer:
@janisozaur great. I will fix conflicts and review in my PR tomorrow and hope it will be merged together soon.

ChrisGuzak:
buildId=8372, as per #406
> I have no idea what that is

This is a C++ CX attribute. It is not strictly needed in this case as the generated .winmd file is not exposed to javascript client code (the case where this attribute means something) so it could be removed.

I'm surprised the other CX language elements `ref` and the `^`s don't cause problems but I'm glad clang-format accepts those.

janisozaur:
Can it be done with a regular C++ attribute?

ChrisGuzak:
Unfortunatly no. CX was invented to enable features like this that are non-standard. I recommend removing it and verifying everything works (it should).

mcooley:
This PR has been assigned to me for almost a year, and sadly I haven't had time to dig into this and figure out the right thing to do. I'm closing this PR (and the related #406) because unfortunately I don't think that's going to change anytime soon. I'm sorry it's taken so long to come to a decision.

Having lived with .clang-format settings in the repo for a while now, the impact on development has been mixed. For the C++/CX code--which unfortunately is a lot of the code in the repo, including the code which changes most frequently--the auto-formatted results can be pretty ugly. So I think we can reconsider this change only if we reduce the amount of C++/CX code we have.

PR #412, opened 2019-03-30
Modify how modulo is calculated in Normal and Scientific mode.

### LINKED ISSUE #111: Modulo operator should work like other calculators out there
The modulo operator on this calculator gives the result that is different to the most used calculators.

This problem appears when you start using negative numbers.

For example in the Bing and Google calculator, Maxima, Wolfram Alpha and Microsoft Excel the modulo operator gives the next results:
-25 mod 4 = 3
-25 mod -4 = -1
25 mod -4 = -3

on the Windows calculator the results came in like this:
-25 mod 4 = -1
-25 mod -4 = -1
25 mod -4 = 1

As you can see the calculation is performed different. Is it wrong? Well, that's is difficult to answer. But given this [article](https://en.wikipedia.org/wiki/Modulo_operation) in the Wikipedia, a lot of the programming languages take que second approach, meanwhile the math applications use the first approach.

Given that the calculator is mainly used to do math and not programming (although it has a programmer mode) I suggest that this change is needed and make it behave like a calculator.

I also sent in the past some [feedback]() to the feedback-hub, but I did not put a lot of information.

**Expected behavior**
We should align scientific and programmatic applications of mod in their respective modes.

>Note: A team member edited this comment for clarity on expected behavior

rudyhuyn:
## Fixes #111

> The modulo operator on this calculator gives the result that is different to the most used calculators.

The current `modrate` function is the equivalent of rem(...)/remainder(...), not mod(...)/modulo(...) available in some popular Math apps. 

- rename `modrate` in `remrate` to be more accurate.
- add `modrate`, calculating modulo similarly to Matlab, Bing, Google calculator, Maxima, Wolfram Alpha and Microsoft Excel 
- Add `RationalMath::Mod` using `modrate` as an alternative to `Rational::operator%` using `remrate`
- Add a helper `SIGN` to retrieve the sign of a `Rational`.
- modify `CalcEngine` to use `modrate` in Normal and Scientific mode and `remrate` in Programmer mode.

- manually and unit tests added

### DIFF
--- a/src/CalcManager/CEngine/Rational.cpp
+++ b/src/CalcManager/CEngine/Rational.cpp
@@ -189,7 +189,7 @@ namespace CalcEngine
 
         try
         {
-            modrat(&lhsRat, rhsRat);
+            remrat(&lhsRat, rhsRat);
             destroyrat(rhsRat);
         }
         catch (uint32_t error)
--- a/src/CalcManager/CEngine/Rational.cpp
+++ b/src/CalcManager/CEngine/Rational.cpp
@@ -186,7 +186,8 @@ namespace CalcEngine
     /// Calculate the remainder after division, the sign of a result will match the sign of the current object.
     /// </summary>
     /// <remarks>
-    /// This function has the same behavior as the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
+    /// This function has the same behavior as the standard C/C++ operator '%'
+    /// to calculate the modulus after division instead, use <see cref="RationalMath::Mod"/> instead.
     /// </remarks>
     Rational& Rational::operator%=(Rational const& rhs)
     {
--- a/src/CalcManager/CEngine/Rational.cpp
+++ b/src/CalcManager/CEngine/Rational.cpp
@@ -183,10 +183,10 @@ namespace CalcEngine
     }
 
     /// <summary>
-    /// Calculate the remainder after division, the sign of the result will match the sign of the current object.
+    /// Calculate the remainder after division, the sign of a result will match the sign of the current object.
     /// </summary>
     /// <remarks>
-    /// This function has the same behavior than the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
+    /// This function has the same behavior as the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
     /// </remarks>
     Rational& Rational::operator%=(Rational const& rhs)
     {
@@ -349,10 +349,10 @@ namespace CalcEngine
     }
 
     /// <summary>
-    /// Calculate the remainder after division, the sign of the result will match the sign of a.
+    /// Calculate the remainder after division, the sign of a result will match the sign of lhs.
     /// </summary>
     /// <remarks>
-    /// This function has the same behavior than the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
+    /// This function has the same behavior as the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
     /// </remarks>
     Rational operator%(Rational lhs, Rational const& rhs)
     {
--- a/src/CalcManager/CEngine/Rational.cpp
+++ b/src/CalcManager/CEngine/Rational.cpp
@@ -182,6 +182,12 @@ namespace CalcEngine
         return *this;
     }
 
+    /// <summary>
+    /// Calculate the remainder after division, the sign of the result will match the sign of the current object.
+    /// </summary>
+    /// <remarks>
+    /// This function has the same behavior than the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
+    /// </remarks>
     Rational& Rational::operator%=(Rational const& rhs)
     {
         PRAT lhsRat = this->ToPRAT();
@@ -342,6 +348,12 @@ namespace CalcEngine
         return lhs;
     }
 
+    /// <summary>
+    /// Calculate the remainder after division, the sign of the result will match the sign of a.
+    /// </summary>
+    /// <remarks>
+    /// This function has the same behavior than the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
+    /// </remarks>
     Rational operator%(Rational lhs, Rational const& rhs)
     {
         lhs %= rhs;
--- a/src/CalcManager/CEngine/RationalMath.cpp
+++ b/src/CalcManager/CEngine/RationalMath.cpp
@@ -387,3 +387,26 @@ Rational RationalMath::ATanh(Rational const& rat)
 
     return result;
 }
+
+
+Rational RationalMath::Mod(Rational const& base, Rational const& n)
+{
+    PRAT prat = base.ToPRAT();
+    PRAT pn = n.ToPRAT();
+
+    try
+    {
+        modrat(&prat, pn);
+        destroyrat(pn);
+    }
+    catch (uint32_t error)
+    {
+        destroyrat(prat);
+        destroyrat(pn);
+        throw(error);
+    }
+
+    auto res = Rational{ prat };
+    destroyrat(prat);
+    return res;
+}
--- a/src/CalcManager/CEngine/RationalMath.cpp
+++ b/src/CalcManager/CEngine/RationalMath.cpp
@@ -388,11 +388,16 @@ Rational RationalMath::ATanh(Rational const& rat)
     return result;
 }
 
-
-Rational RationalMath::Mod(Rational const& base, Rational const& n)
+/// <summary>
+/// Calculate the modulus after division, the sign of the result will match the sign of b.
+/// </summary>
+/// <remarks>
+/// When one of the operand is negative, the result will differ from the C/C++ operator '%', use <see cref="Rational::operator%"/> instead to calculate the remainder after division.
+/// </remarks>
+Rational RationalMath::Mod(Rational const& a, Rational const& b)
 {
-    PRAT prat = base.ToPRAT();
-    PRAT pn = n.ToPRAT();
+    PRAT prat = a.ToPRAT();
+    PRAT pn = b.ToPRAT();
 
     try
     {
--- a/src/CalcManager/CEngine/RationalMath.cpp
+++ b/src/CalcManager/CEngine/RationalMath.cpp
@@ -392,7 +392,9 @@ Rational RationalMath::ATanh(Rational const& rat)
 /// Calculate the modulus after division, the sign of the result will match the sign of b.
 /// </summary>
 /// <remarks>
-/// When one of the operand is negative, the result will differ from the C/C++ operator '%', use <see cref="Rational::operator%"/> instead to calculate the remainder after division.
+/// When one of the operand is negative
+/// the result will differ from the C/C++ operator '%'
+/// use <see cref="Rational::operator%"/> instead to calculate the remainder after division.
 /// </remarks>
 Rational RationalMath::Mod(Rational const& a, Rational const& b)
 {
--- a/src/CalcManager/CEngine/scioper.cpp
+++ b/src/CalcManager/CEngine/scioper.cpp
@@ -112,8 +112,17 @@ CalcEngine::Rational CCalcEngine::DoOperation(int operation, CalcEngine::Rationa
             }
             else
             {
-                iFinalSign = iNumeratorSign;
-                result %= temp;
+                if (m_fIntegerMode)
+                {
+                    // Programmer mode
+                    iFinalSign = iNumeratorSign;
+                    result %= temp;
+                }
+                else
+                {
+                    iFinalSign = iDenominatorSign;
+                    result = Mod(result, temp);
+                }
             }
 
             if (m_fIntegerMode && iFinalSign == -1)
--- a/src/CalcManager/CEngine/scioper.cpp
+++ b/src/CalcManager/CEngine/scioper.cpp
@@ -114,12 +114,13 @@ CalcEngine::Rational CCalcEngine::DoOperation(int operation, CalcEngine::Rationa
             {
                 if (m_fIntegerMode)
                 {
-                    // Programmer mode
+                    // Programmer mode, use remrat (remainder after division)
                     iFinalSign = iNumeratorSign;
                     result %= temp;
                 }
                 else
                 {
+                    //other modes, use modrat (modulus after division)
                     iFinalSign = iDenominatorSign;
                     result = Mod(result, temp);
                 }
--- a/src/CalcManager/CEngine/scioper.cpp
+++ b/src/CalcManager/CEngine/scioper.cpp
@@ -78,7 +78,7 @@ CalcEngine::Rational CCalcEngine::DoOperation(int operation, CalcEngine::Rationa
         case IDC_DIV:
         case IDC_MOD:
         {
-            int iNumeratorSign = 1, iDenominatorSign = 1, iFinalSign = 1;
+            int iNumeratorSign = 1, iDenominatorSign = 1;
             auto temp = result;
             result = rhs;
 
@@ -107,30 +107,30 @@ CalcEngine::Rational CCalcEngine::DoOperation(int operation, CalcEngine::Rationa
 
             if (operation == IDC_DIV)
             {
-                iFinalSign = iNumeratorSign * iDenominatorSign;
                 result /= temp;
+                if (m_fIntegerMode && (iNumeratorSign * iDenominatorSign) == -1)
+                {
+                    result = -(Integer(result));
+                }
             }
             else
             {
                 if (m_fIntegerMode)
                 {
                     // Programmer mode, use remrat (remainder after division)
-                    iFinalSign = iNumeratorSign;
                     result %= temp;
+
+                    if (iNumeratorSign == -1)
+                    {
+                        result = -(Integer(result));
+                    }
                 }
                 else
                 {
                     //other modes, use modrat (modulus after division)
-                    iFinalSign = iDenominatorSign;
                     result = Mod(result, temp);
                 }
             }
-
-            if (m_fIntegerMode && iFinalSign == -1)
-            {
-                result = -(Integer(result));
-            }
-
             break;
         }
 
--- a/src/CalcManager/Ratpack/itrans.cpp
+++ b/src/CalcManager/Ratpack/itrans.cpp
@@ -92,11 +92,9 @@ void asinanglerat( _Inout_ PRAT *pa, ANGLE_TYPE angletype, uint32_t radix, int32
 void asinrat( PRAT *px, uint32_t radix, int32_t precision)
 
 {
-    int32_t sgn;
     PRAT pret= nullptr;
     PRAT phack= nullptr;
-
-    sgn = SIGN(*px);
+    int32_t sgn = SIGN(*px);
 
     (*px)->pp->sign = 1;
     (*px)->pq->sign = 1;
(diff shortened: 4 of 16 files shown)


janisozaur:
@rudyhuyn I'd like to ask you to follow the common guidelines for writing commit messages: https://chris.beams.io/posts/git-commit/ https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53 and more. Basically, keep the subject line a subject.

rudyhuyn:
@janisozaur: I disagree and agree at the same time. 

I agree for Master/public branches.
But all commits from a PR are squashed once merged in Master, the subject of the commit becoming the title of the PR. So we have, when you look at the history you have:
- commit messages: short title (in fact the title of the PR associated)
- a preview of the detailed description of PR associated: details.

It's perfect, you can quickly read the history and get more details if you want.

This is not how commits work in a PR, their roles are very different.

As a developer, the only field you have access to describe the changes of a commit is the commit subject, you don't have another option, so it's better to be verbose and give the more information possible to reviewers to make it easier it to understand what you did since the last time they reviewed your code.

So I agree for Master/public branches, but not inside a PR. I saw yesterday a comment asking a developer to Squash its commits in a PR (It was a giant diff with more than 30 commits), it doesn't make sense, they will already be squashed once approved and merged, it's better to keep the history of all changes of the PR (in case the developer needs to roll back a change) than squashing for no real reason.

janisozaur:
Ah, I was not aware they squash the PRs. I just checked and it indeed is the case. I'll get to that in just a moment.

> This is not how commits work in a PR, their roles are very different.

I disagree with your arguments, I don't see how providing an actually usable subject line in a commit depends on place where it is used. You also seem to rely on all the PRs being squashed, which should not be the case. A commit in a PR is no different outside a PR – a set of changes that are coherent and can be explained with a short subject line.

Regarding squashing of PRs: I'm surprised that happens, but I have never checked. It has a number of issues, to name a few:
* It loses track of authors. It might be fine if you're a single author of all the commits in given PR, but it's often not the case
* It loses the granularity of changes. Each commit should be a self-contained, logically coherent set of changes, see https://www.git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

Have you read https://github.com/Microsoft/calculator/blob/master/CONTRIBUTING.md ?

https://github.com/Microsoft/calculator/blob/8a75dcd09d6630ac24e049af9ebb1164fdfac9bf/CONTRIBUTING.md#L52-L53

https://github.com/Microsoft/calculator/blob/8a75dcd09d6630ac24e049af9ebb1164fdfac9bf/CONTRIBUTING.md#L57-L62

If you follow the first link to https://guides.github.com/introduction/flow/ you can also see a helpful diagram explaining the process and describing how commits are _merged_ into master:
image

janisozaur:
https://github.com/Microsoft/calculator/blob/8a75dcd09d6630ac24e049af9ebb1164fdfac9bf/CONTRIBUTING.md#L61-L62

Ugh, I just re-read my comment and noticed that part. Still, my other arguments are unaffected.

rudyhuyn:
I agree for issues due to the squashing (I'm not marked as the author of 4-5 commits in other Microsoft git repository because of this system while I wrote 100% of the diff...).

But it also makes sense and you should really assume it works this way. You owe our own forked repository, when you create a PR, you don't request to merge your branches (your forked repo will not even know) but to apply a patch based on your branch. It's a different logic and probably a better and safer one. 
- better: shorter history (without thousands of commits named "take feedback in account", "fix typo", "add unit tests"
- safer: 100% of the code visible in the history is approved and stamped by the owner of the repo (Microsoft), they become the official owner and are responsible in case of copyright/pattern infringement, so better to squash everything to to take the ownership of the "sub-commits" not reviewed

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace CalculatorManagerTest
+{
+    TEST_CLASS(RationalTest)
+    {
+    public:
+        TEST_CLASS_INITIALIZE(CommonSetup)
+        {
+            ChangeConstants(10, 128);
+        }
+        TEST_METHOD(ModuloTest)
```
joshkoon: Additional tests for both functions:

- Floating point values, positive and negative e.g. 1.5 mod 0.5, 1.5 mod 0.4.
- 0 as numerator
- Numerator and Denominator are the same
- Numerator and Denominator are equivalent but not the same e.g. pa = Rational(Number(27), Number(12)) and b = Rational(Number(9),Number(4))

src/CalcManager/Ratpack/itrans.cpp
```diff
@@ -206,7 +206,7 @@ void acosrat( PRAT *px, uint32_t radix, int32_t precision)
 {
     int32_t sgn;
 
-    sgn = (*px)->pp->sign*(*px)->pq->sign;
+    sgn = SIGN(*px);
```
joshkoon: Combine the declaration at line 207 with this initialization?

joshkoon: Here and elsewhere that you're using SIGN(): if it's appropriate to declare-and-assign at the same time, we should. No need to change all the surrounding variables to follow this right now, but since you're updating the `sgn` variables, let's clean them up a bit.

src/CalcManager/CEngine/scioper.cpp
```diff
@@ -112,8 +112,17 @@ CalcEngine::Rational CCalcEngine::DoOperation(int operation, CalcEngine::Rationa
...
+                if (m_fIntegerMode)
+                {
+                    // Programmer mode
+                    iFinalSign = iNumeratorSign;
+                    result %= temp;
+                }
+                else
+                {
+                    iFinalSign = iDenominatorSign;
+                    result = Mod(result, temp);
```
joshkoon: Let's throw a comment on here as well that this is the arithmetic mod. The more clear we can make the code distinguishing our use of %/remrat and Mod/modrat and noting that the decision to use one or the other is deliberate, the better.

src/CalcManager/CEngine/scioper.cpp
```diff
@@ -112,8 +112,17 @@ CalcEngine::Rational CCalcEngine::DoOperation(int operation, CalcEngine::Rationa
...
-                result %= temp;
+                if (m_fIntegerMode)
+                {
+                    // Programmer mode
+                    iFinalSign = iNumeratorSign;
+                    result %= temp;
+                }
+                else
+                {
+                    iFinalSign = iDenominatorSign;
```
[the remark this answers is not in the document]
rudyhuyn: There was some discussions about having 2 buttons instead of 1 (`mod` and `rem`), even if not used, I prefer to still indicate that the sign is not calculated the same way.

src/CalcManager/CEngine/RationalMath.cpp
```diff
@@ -387,3 +387,26 @@ Rational RationalMath::ATanh(Rational const& rat)
 
     return result;
 }
+
+
```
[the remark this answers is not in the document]
rudyhuyn: I will!

+ a note explaining that Rational::operator% is consistent with the float::operator% (based on remainder)

src/CalcManager/Ratpack/logic.cpp
```diff
@@ -185,34 +185,77 @@ void boolnum( PNUMBER *pa, PNUMBER b, int func )
...
+
+//-----------------------------------------------------------------------------
+//
+//    FUNCTION: modrat
+//
+//    ARGUMENTS: pointer to a rational a second rational.
+//
+//    RETURN: None, changes pointer.
+//
+//    DESCRIPTION: Calculate the remainder of *pa / b, equivalent of 'pa modulo b' in arithmetic
```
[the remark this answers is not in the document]
rudyhuyn: I will add more comments, but I learned yesterday that Mod and Rem are both Euclidean modulo (I was sure it was only the case of the first one), the naive definition of modulo doesn't take into account negative numbers, creating even more confusion

src/CalcManager/Ratpack/logic.cpp
```diff
@@ -185,34 +185,77 @@ void boolnum( PNUMBER *pa, PNUMBER b, int func )
...
+//    NOTE: produces a result that is either zero or has the same sign as the divisor.
+//
+//-----------------------------------------------------------------------------
+
+void modrat(PRAT *pa, PRAT b)
+{
+    //contrary to remrat, modrat(a, 0) must return a
```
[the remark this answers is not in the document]
rudyhuyn: no, it returns the value of 'a' :D

src/CalcManager/CEngine/scioper.cpp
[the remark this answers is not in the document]
joshkoon: Understood - given that it isn't currently used, however, I'd recommend commenting it out and adding a short comment noting that it's unused but represents the logic if arithmetic mod is added to programmer mode.

src/CalcManager/CEngine/Rational.cpp
```diff
@@ -182,14 +182,20 @@ namespace CalcEngine
         return *this;
     }
 
+    /// <summary>
+    /// Calculate the remainder after division, the sign of the result will match the sign of the current object.
+    /// </summary>
+    /// <remarks>
+    /// This function has the same behavior than the standard C/C++ operator '%', to calculate the modulus after division instead, use <see cref="Rational::operator%"/> instead.
```
danbelcher-MSFT:
```suggestion
    /// This function has the same behavior as the standard C/C++ operator '%.
    /// To calculate the modulus after division instead, use <see cref="Rational::operator%"/>.
```

src/CalcManager/CEngine/Rational.cpp
```diff
@@ -342,6 +348,12 @@ namespace CalcEngine
         return lhs;
     }
 
+    /// <summary>
+    /// Calculate the remainder after division, the sign of the result will match the sign of a.
```
danbelcher-MSFT: "sign of a...."

src/CalcManager/Ratpack/logic.cpp
```diff
@@ -172,47 +172,93 @@ void boolnum( PNUMBER *pa, PNUMBER b, int func )
...
 //
-//    DESCRIPTION: Does the rational equivalent of frac(*pa);
+//    DESCRIPTION: Calculate the remainder of *pa / b,
+//                 equivalent of 'pa % b' in C/C++ and produces a result
+//                 that is either zero or has the same sign as the dividend.
 //
 //-----------------------------------------------------------------------------
 
-void modrat( PRAT *pa, PRAT b )
+void remrat(PRAT *pa, PRAT b)
```
danbelcher-MSFT: Why does this need to be a `PRAT*`?  Can't it be a `PRAT`?

src/CalcManager/CEngine/Rational.cpp
danbelcher-MSFT: The remarks for both comments added in this file confuse me.  They both redirect to `operator%` but they are _describing_ `operator%`!  (circular reference).  Did you intend to direct me somewhere else?

src/CalcManager/Ratpack/logic.cpp
```diff
@@ -172,47 +172,93 @@ void boolnum( PNUMBER *pa, PNUMBER b, int func )
...
+//    RETURN: None, changes pointer.
+//
+//    DESCRIPTION: Calculate the remainder of *pa / b, with the sign of the result
+//                 either zero or has the same sign as the divisor. 
+//    NOTE: When *pa or b are negative, the result won't be the same as
+//          the C/C++ operator %, use remrat if it's the behavior you expect. 
+//
+//-----------------------------------------------------------------------------
+
+void modrat(PRAT *pa, PRAT b)
```
danbelcher-MSFT: Same comment about `PRAT*` -> `PRAT`

src/CalcManager/Ratpack/ratpak.h
```diff
@@ -148,6 +148,9 @@ extern PRAT rat_min_i32;
 #define LOGNUM2(pnum) ((pnum)->cdigit+(pnum)->exp)
 #define LOGRAT2(prat) (LOGNUM2((prat)->pp)-LOGNUM2((prat)->pq))
 
+// SIGN returns the sign of the rational
+#define SIGN(prat) ((prat)->pp->sign*(prat)->pq->sign)
```
danbelcher-MSFT: Hmm, I realize this is the style of this file, but maybe we should still update this to be a free function?

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,189 @@
...
+        {
+            Rational rat25(25);
+            Rational ratminus25(-25);
+            Rational rat4(4);
+            Rational ratminus4(-4);
+            Rational res = Mod(rat25, rat4);
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 128), L"1");
+            res = Mod(rat25, ratminus4);
```
danbelcher-MSFT: `Rational` types can be implicitly constructed from `int` values, meaning we can write simpler and more readable code such as:
```suggestion
            res = Mod(25, -4);
```
Assuming you update these test to take advantage of the implicit conversion, is there a need for the named `Rational` objects you declared above? (`rat25`, `ratminus25`, etc.).

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,189 @@
...
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 128), L"-3");
```
danbelcher-MSFT: Do we need to be comparing against strings here?  Why not use:
```suggestion
            VERIFY_ARE_EQUAL(res, -3);
```

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,189 @@
...
+namespace CalculatorEngineTests
+{
+    TEST_CLASS(RationalTest)
+    {
+    public:
+        TEST_CLASS_INITIALIZE(CommonSetup)
+        {
+            ChangeConstants(10, 128);
+        }
+        TEST_METHOD(ModuloTest)
```
danbelcher-MSFT: @sanderl, thoughts on whether bulk tests like this are fine or if we should be breaking into smaller components?

src/CalcManager/Ratpack/logic.cpp
rudyhuyn: because the result of pa % b is stored in `pa`

src/CalcManager/Ratpack/ratpak.h
rudyhuyn: I'm not a big fan of macros, I would prefer `Sign()` to be a function of (P)RAT, but we should keep the code consistent. (Future PR to refactor RatPack?)

src/CalculatorUnitTests/RationalTest.cpp
rudyhuyn: I missed the implicit conversion, I will update the tests

rudyhuyn: It was part of the first version of the unit tests, before adding more. But I think we should keep them and use them to check that Mod and Rem don't change the value of operands.

src/CalcManager/CEngine/Rational.cpp
rudyhuyn: You are right! Reference fixed!

src/CalcManager/Ratpack/logic.cpp
joshkoon: While PRAT is just a typedef'd RAT* and can therefore be modified as a reference type, the RatPack uses PRAT as though it were a value-type, passing PRAT* to functions that will modify the value. For consistency, I think sticking with that precedent makes sense.

src/CalculatorUnitTests/RationalTest.cpp
sanderl: A lot of our tests have been grouping buckets of tests together has Rudy has done. I would probably break these up a little more so that we can get more coverage if one of the tests fails. The test stops execution once it fails, so the later tests within the test method would not execute. So I would probably make each section its own test. So there would be 1 test for verify the operands are not modified, 1 test for Modulo with integers, 1 for Modulo with zeros, etc.

rudyhuyn: @sanderl I will modify that!

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,226 @@
...
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"0.33333333");
+            res = Mod(Rational(Number(1, 0, { 1000 }), Number(1, 0, { 3 })), Rational(-10));
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"-6.6666667");
+            res = Mod(Rational(834345), Rational(Number(1, 0, { 103 }), Number(1, 0, { 100 })));
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"0.71");
+            res = Mod(Rational(834345), Rational(Number(-1, 0, { 103 }), Number(1, 0, { 100 })));
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"-0.32");
+        }
+
+        TEST_METHOD(TestRemainerOperandsNotModified)
```
joshkoon: nit: Typo: `TestRemainderOperandsNotModified`

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,226 @@
...
+            VERIFY_ARE_EQUAL(res, -1);
+            VERIFY_ARE_EQUAL(ratminus25, -25);
+            VERIFY_ARE_EQUAL(ratminus4, -4);
+            res = ratminus25 % rat4;
+            VERIFY_ARE_EQUAL(res, -1);
+            VERIFY_ARE_EQUAL(ratminus25, -25);
+            VERIFY_ARE_EQUAL(rat4, 4);
+        }
+
+        TEST_METHOD(TestRemainerInteger)
```
joshkoon: nit: Typo: `TestRemainderInteger`. Looks like it applies to other functions with `Remainder` in the name

src/CalculatorUnitTests/RationalTest.cpp
```diff
@@ -0,0 +1,226 @@
...
+            res = Rational(Number(-1, 0, { 12250 }), Number(1, 0, { 100 })) % Rational(10);
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"-2.5");
+            res = Rational(Number(-1, 0, { 12250 }), Number(1, 0, { 100 })) % Rational(-10);
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"-2.5");
+            res = Rational(Number(1, 0, { 12250 }), Number(1, 0, { 100 })) % Rational(-10);
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"2.5");
+            res = Rational(Number(1, 0, { 1000 }), Number(1, 0, { 3 })) % Rational(1);
+            VERIFY_ARE_EQUAL(res.ToString(10, FMT_FLOAT, 8), L"0.33333333");
+            res = Rational(Number(1, 0, { 1000 }), Number(1, 0, { 3 })) % Rational(-10);
+            auto sdsdas = res.ToString(10, FMT_FLOAT, 8);
```
joshkoon: Doesn't look like sdsdas is used? Artifact from debugging something?

rudyhuyn: :(

src/CalculatorUnitTests/RationalTest.cpp
rudyhuyn: Tip: Never code while watching news about Brexit!

PR #422, opened 2019-04-03
Updating version of release builds to 1904

mcooley:


HowardWolosky:
We should add a tag to master to denote what sha represents 1903.

mcooley:
@HowardWolosky internally we have a document which tracks the sha for each release. If you think we need tracking here, let's create a servicing/1903 branch based on that commit instead of using tags.

PR #430, opened 2019-04-04
Keep spaces when users copy an error message

### LINKED ISSUE #420: Copying "Result is undefined" to the clipboard pastes as "Resultisundefined'
**Describe the bug**

When dividing `0/0` and the "Result is undefined" text appears, copying the result to the clipboard and pasting it results in a string without spaces.

**Steps To Reproduce**

1. Open Calculator
2. Type `0/0`
3. Right-click on `Result is undefined` and select Copy
4. Open a text edit and paste
5. The pasted text will be `Resultisundefined`

**Expected behavior**

The pasted text should be `Result is undefined`

**Screenshots**

 - OS Build: 10.0.17763.0
 - Architecture: X64
 - Application Version: 10.1902.42.0

**Additional context**

This also occurs when copying the `Invalid input` text (it pastes as `Invalidinput`).

rudyhuyn:
- verify if the `StandardCalculatorViewModel::DisplayValue` isn't an error message before removing group separators (including space character).

- Tested in english and french
- Tested with "Cannot divide by zero" and "Result is undefined"


src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -1326,13 +1326,18 @@ void StandardCalculatorViewModel::SetCalculatorType(ViewMode targetState)
...
 
-Platform::String^ StandardCalculatorViewModel::GetRawDisplayValue()
+String^ StandardCalculatorViewModel::GetRawDisplayValue()
 {
-    wstring rawValue;
-
-    LocalizationSettings::GetInstance().RemoveGroupSeparators(DisplayValue->Data(), DisplayValue->Length(), &rawValue);
-
-    return ref new Platform::String(rawValue.c_str());
+    if (m_IsInError)
```
HowardWolosky: While this is technically right, it's leaking abstraction.  Nowhere else in the code do we directly reference `m_IsInError`, but instead it gets accessed through its property `IsInError`.  Is there a specific reason that I'm missing for why you're directly accessing this template-created member var?

```suggestion
    if (IsInError)
```

rudyhuyn: I prefer to use fields instead of properties when they are accessible, especially when the getter is not inline (we should probably modify the macro `OBSERVABLE_PROPERTY_RW` in my opinion). 

It looks like I'm not the only developer thinking this way, except `IsStandard`, `IsScientific` and `IsProgrammer`, almost all properties are accessed via the field associated instead of the getter in this class.

HowardWolosky: @danbelcher-MSFT - Do you have a preferred pattern here that you'd like to enforce within the codebase?  My pereference in projects has always been to always access through a property when one exists, regardless of being inside or outside of the defining class:

   1. It enables easy consistency when trying to find all references to a value (as opposed to having to remember to check for references to both the field _and_ the property)... In this case, it took me an extra minute to even see what was going on, because the field's definition couldn't even be found (since it was generated through templating).  Being able to reliably identify all consumers is _huge_ for me.
   2. By centralizing access patterns, it makes it easy to adjust the value's logic across all consumers, if it becomes necessary.  This is certainly an infrequent occurrence once the value is already defined, but does come up on occasion.

I didn't catch that there's such wide inconsistency already in how this is handled within the project.  We should definitely define the expected access pattern for future updates.

danbelcher-MSFT: Sorry for the late reply.  I don't think I have a personal preference here.  My thought is that we're dealing with the implementation of a class and so it is fully aware of its own data members.  I do understand the argument for consistency and finding all uses of a property is an important scenario.  There isn't a significant performance impact from changing to accessors everywhere, so  it seems like the best approach for uniformity.

PR #432, opened 2019-04-04
Make History and Memory list items context menu accessible using keyboard

### LINKED ISSUE #311: "History" list item context menu is not accessible using keyboard
**Repro Steps:**
* Launch Calculator app.
* Navigate to History list item and try to open the context menu
* Observe the issue.

**Actual Result**
History list item context menu is not accessible using keyboard. It is accessible using mouse.

**Expected Result**
History list item context menu should be accessible using keyboard.

App Version: 10.1902.16.0

rudyhuyn:
- Move History and Memory context menus to ListViewItemContainer so users can open them with keyboards
- Modify how Memory manages context menus to use a similar approach than History (a lot simplier)

- Manually test using keyboard, mouse and touch.


src/Calculator/Views/HistoryList.xaml.cpp
```diff
@@ -67,9 +67,13 @@ void HistoryList::ListView_ItemClick(_In_ Object^ sender, _In_ ItemClickEventArg
 
 void HistoryList::OnDeleteMenuItemClicked(_In_ Object^ sender, _In_ RoutedEventArgs^ e)
 {
-    auto clickedItem = safe_cast<HistoryItemViewModel^>(safe_cast<FrameworkElement^>(sender)->DataContext);
+    auto listViewItem = HistoryContextMenu->Target;
+    auto itemViewModel = safe_cast<HistoryItemViewModel^>(HistoryListView->ItemFromContainer(listViewItem));
```
joseartrivera: safe_cast would throw here if the cast didn't work, so the below check for nullptr would never occur. I think you want dynamic_cast instead

src/Calculator/Views/HistoryList.xaml
```diff
@@ -51,17 +51,15 @@
...
                                 Click="OnDeleteMenuItemClicked"
                                 Icon="Delete"/>
             </MenuFlyout>
 
             <DataTemplate x:Key="HistoryItemTemplate" x:DataType="model:HistoryItemViewModel">
                 <muxc:SwipeControl RightItems="{StaticResource HistorySwipeItems}">
-                    <StackPanel Margin="0,6,4,6"
-                                Background="Transparent"
-                                ContextFlyout="{StaticResource HistoryContextMenu}">
+                    <StackPanel Margin="0,6,4,6"Background="Transparent">
```
joseartrivera: Add a space before Background

src/Calculator/Views/Memory.xaml.cpp
```diff
@@ -135,7 +112,6 @@ void Memory::MemoryList_Unloaded(_In_ Object^ sender, _In_ RoutedEventArgs^ e)
 
 MemoryItemViewModel^ Memory::GetMemoryItemForCurrentFlyout()
 {
-    auto listViewItem = m_memoryItemFlyout->Target;
-
+    auto listViewItem = MemoryContextMenu->Target;
     return safe_cast<MemoryItemViewModel^>(MemoryListView->ItemFromContainer(listViewItem));
```
joseartrivera: Same feedback here safe_cast would throw so the checks for nullptr in the calling code would not occur.

src/Calculator/Views/HistoryList.xaml.cpp
rudyhuyn: Sorry for the delay, I was very busy these past few days, I will modify that tonight!

joseartrivera:
@rudyhuyn this looks good, could you resolve the merge conflicts and then I can merge it in?

PR #436, opened 2019-04-04
Compile CalcManager project with or without precompiled headers

### LINKED ISSUE #324: Consider reworking how PCHs are used
As discussed in https://github.com/Microsoft/calculator/pull/208#discussion_r263955391:

PCHs in current form are problematic to use due to varying support of build systems. While the most common setup I imagine would be with the MSVS project files included in the repository, #211 adds CMake support and when preparing #212 I manually invoked the compiler, which causes inclusion of many unnecessary header files, which leads to prolonged compilation times.

I can suggest a simple solution to that issue: make the leaf `.cpp` files `#include` the minimal set of what they truly need, leave `pch.{h,cpp}` files intact, then upon building:
1. Compile the PCH, e.g. 
```
g++ -std=c++17 -x c++-header src/CalcManager/pch.cpp -c -o src/CalcManager/pch.h.gch
```
2. When compiling each of leaf `.cpp` files, inject the header via compiler, e.g. 
```
g++ -std=c++17 -I src/CalcManager -I src/CalcManager/Header\ Files -c  src/CalcManager/CEngine/calc.cpp -include pch.h
```

This way you can have regular PCHs without forcing half of STL into each translation unit even when they don't need it.

I'm sure there are other ways of solving that issue as well.

danbelcher-MSFT:
In an effort to support other compilers (#109), this change reworks how precompiled headers are handled.  For toolchains where precompiled headers are not used, there is unnecessary compilation cost because each source file explicity includes the pch, meaning all system headers in the pch were recompiled for each translation unit. This change modifies the project's files so that each translation unit includes a minimal set of dependent headers.  For MSVC users, the precompiled headers option is still enabled and the precompiled header is added to each translation unit using the compiler's Forced Includes option.  The end result is that MSVC users still see the same build times, but other toolchains are free to use or not use precompiled headers.

### Risks introduced
Given that our CI build uses MSVC, this change introduces the risk that a system header is added to the pch and the CalcManager project builds correctly, but builds could be broken for other toolsets that don't use pch.  We know we want to add support for Clang in our CI build ([#211](https://github.com/Microsoft/calculator/pull/211#issuecomment-472186512)).  It seems reasonable to also compile without precompiled headers there so that we can regression test this setup.

1. Rebuild CalcManager project. Compile time: `~4.5s`.
2. Disable precompiled headers, keeping explicit include for pch in each source file. Compile time: `~13s`.
3. Remove explicit pch inclusion and add the appropriate headers to each translation unit to allow the project to compile.  Compile time: `~8s`.
4. Re-enable pch and include it using the Forced Includes compiler option. MSVC compile time: `~4.5s`.

#### Minor changes
- Delete 'targetver.h'.  I found this while looking around for system headers in the project.  It's unused and unreferenced so let's remove it.


CONTRIBUTING.md
```diff
@@ -44,6 +44,12 @@ the code. Please attempt to match the style of surrounding code as much as possi
 components, prefer the patterns described in the [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)
 and the [modern C++/WinRT language projections](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/).
 
+The projects in this codebase typically make use of precompiled headers (pch.h). When it is necessary to include a
```
mcooley: If someone doesn't understand this information the consequences are not very severe. Either they'll add a header only to the pch, in which case we'll catch it when we have automated Clang builds set up. Or they'll add it only to the specific file, in which case there's no harm, just a missed perf improvement.

So...can this just be a comment in pch.h, to keep CONTRIBUTING from getting bloated with obscure details that most people don't need to care about?

src/CalcManager/pch.h
```diff
@@ -3,27 +3,15 @@
...
-#endif
-
-// Windows headers define min/max macros.
-// Disable it for project code.
-#define NOMINMAX
-
+#include <algorithm>
+#include <array>
 #include <assert.h>
-#include <windows.h>
```
mcooley: So we've eliminated all our dependencies on windows.h? That's exciting!!

CONTRIBUTING.md
danbelcher-MSFT: I see your point about it feeling out-of-place here and I'm fine with moving the comment.  @HowardWolosky, thoughts?

src/CalcManager/pch.h
danbelcher-MSFT: On the top-level windows.h header, yes, and it is exciting!  We still have winerror.h, although #211 makes that available cross-platform.  It should be simple to take it a step further and eliminate winerror as well.

src/CalcManager/CEngine/calc.cpp
```diff
@@ -1,9 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "pch.h"
+#include <assert.h>
```
janisozaur:
```suggestion
#include <cassert>
```

janisozaur: There are at least two other places in this PR with same change

janisozaur:
This is great! I'll check GCC and clang builds and leave my review at some point today. Many thanks for implementing my idea.

src/CalcManager/CEngine/calc.cpp
danbelcher-MSFT: Thanks!

CONTRIBUTING.md
danbelcher-MSFT: Move to the CalcManager pch.

src/CalcManager/pch.h
janisozaur: `intsafe.h` is on the chopping block soon too: https://github.com/Microsoft/calculator/pull/398. That would leave `ppltasks.h` as the only Windows include here.

danbelcher-MSFT: #211 gets rid of ppltasks so we're almost there.

src/CalcManager/CalculatorManager.cpp
```diff
@@ -1,7 +1,6 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "pch.h"
 #include "Header Files/CalcEngine.h"
```
janisozaur:
```suggestion
#include <climits> // for UCHAR_MAX
#include "Header Files/CalcEngine.h"
```

src/CalcManager/CalculatorVector.h
```diff
@@ -1,8 +1,11 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 #pragma once
 
+#include <string>
+#include <vector>
+#include <winerror.h>
 #include "Ratpack/CalcErr.h"
```
janisozaur:
```suggestion
#include "Ratpack/CalcErr.h"
#include <stdexcept> // for std::out_of_range
#include <sal.h> // for SAL
```

src/CalcManager/Ratpack/basex.cpp
```diff
@@ -14,7 +14,6 @@
 //  internal base is a power of 2.
 //
 //-----------------------------------------------------------------------------
-#include "pch.h"
 #include "ratpak.h"
```
janisozaur:
```suggestion
#include "ratpak.h"
#include <cstring> // for memmove
```

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -17,7 +17,9 @@
 //
 //---------------------------------------------------------------------------
 
-#include "pch.h"
+#include <algorithm>
+#include <winerror.h>
+#include <sstream>
```
janisozaur:
```suggestion
#include <sstream>
#include <cstring> // for memmove, memcpy
```

src/CalcManager/Ratpack/num.cpp
```diff
@@ -17,7 +17,7 @@
 //
 //
 //-----------------------------------------------------------------------------
-#include "pch.h"
+#include <list>
```
janisozaur:
```suggestion
#include <list>
#include <cstring> // for memmove
```

src/CalcManager/Ratpack/ratpak.h
```diff
@@ -17,6 +17,8 @@
 //
 //-----------------------------------------------------------------------------
 
+#include <algorithm>
+#include <string>
 #include "CalcErr.h"
```
janisozaur:
```suggestion
#include "CalcErr.h"
#include <cstring> // for memmove
#include <sal.h> // for SAL
```

src/CalcManager/Ratpack/support.cpp
```diff
@@ -18,7 +18,7 @@
 //
 //----------------------------------------------------------------------------
 
-#include "pch.h"
+#include <string>
```
janisozaur:
```suggestion
#include <string>
#include <cstring> // for memmove
#include <iostream> // for wostream
```

src/CalcManager/UnitConverter.cpp
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#include "pch.h"
+#include <cassert>
+#include <sstream>
```
janisozaur:
```suggestion
#include <sstream>
#include <algorithm> // for std::sort
```

src/CalcManager/UnitConverter.h
```diff
@@ -1,8 +1,12 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 #pragma once
 
+#include <vector>
+#include <unordered_map>
+#include <ppltasks.h>
```
janisozaur:
```suggestion
#include <ppltasks.h>
#include <sal.h> // for SAL
#include <memory> // for std::shared_ptr
```

src/CalcManager/pch.h
janisozaur: Indeed, just noticed. :tada:

danbelcher-MSFT:
@janisozaur, how are you identifying these extra header dependencies?  I rebased my branch on top of #211 and I was able to build with CMake/Clang with no problems.

janisozaur:
GCC/libstdc++ and lots of time spent on writing code for the trinity of compilers.

danbelcher-MSFT:
I see.  I don't have GCC set up so I can't verify your changes right now.  I'll have to trust you :)
I do have a question.. For example, in UnitConverter.h, you suggested to add `<memory>` because the header uses `std::shared_ptr`.  You _didn't_ suggest `<string>` even though the header uses `std::wstring`.  Why the discrepancy?

danbelcher-MSFT:
Second suggest..  How do you feel about taking this change so we can unblock #211, then we can add onto it to support GCC?

janisozaur:
It must've been transitively included via other file. I don't claim this is exhaustive, but it works in my case. Of you find any more missing headers, I'd suggest you include them. I might spend more time on that in the future.

It's not really blocking #211 in my opinion, but I can see how having this merged first would lead to having nicer history.

janisozaur:
Actually, #211 seems to do away with some headers. It might actually be required for this to go in first (with comments addressed) or #211 to add some includes.

danbelcher-MSFT:
@janisozaur, can you take a second look?  I tried requesting your review again through GitHub but the 'refresh' button doesn't seem to be working for me right now.  Sorry if I spammed you with notifications!

@mcooley, I need your review so I can merge.

janisozaur:
Sure, I'll re-review.

danbelcher-MSFT:
@janisozaur, I'm ready to merge this change now.  I took your suggestions without a way to validate with GCC.  If you find that I broke something, feel free to comment on this PR or supply a fix.

PR #442, opened 2019-04-06
Make it obvious that the window doesn't have focus + simplify how the app manages the title bar.

### LINKED ISSUE #407: Code Cleaning: AppBar, OperatorTextBox and OperandTextBox, TitleBar never used
**Cleaning**
The 3 following controls in Calculator/Controls are never used in the application.
- AppBar: https://github.com/Microsoft/calculator/search?q=AppBar&unscoped_q=AppBar
- OperatorTextBox: https://github.com/Microsoft/calculator/search?q=OperatorTextBox
- OperandTextBox: https://github.com/Microsoft/calculator/search?q=OperandTextBox

TitleBar and TitleBarHelper are used in different parts of the application (Memory, HistoryList, ...), but it seems like these controls have no effects on the app (old code from Windows 8/Windows Phone 8.1?)

rudyhuyn:
- Remove TitleBarHelper and all `<Border x:Name="CustomTitleBar" />`
- Let the system defines the draggable region
- Centralize all events and functions associated to the title bar in a single control `TitleBar` instead of code splitted between MainPage/TitleBar/HistoryList/Memory.
- Use the standard title bar when high contrast is activated instead of the custom one.
- Modify the color of the title when the window doesn't have focus
- Fix the right padding of the title bar with high contrast

- Manually tested with LtR and RtL languages
- Manually tested with high contrast
- Tested when History and Memory flyout are opened

Window without focus - dark
image

Window without focus - light
image

Window without focus - high contrast
image

Right padding fixed with high contrast
image


src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -1,90 +1,86 @@
...
+        m_uiSettings->ColorValuesChanged -= m_colorValuesChangedToken;
+        m_colorValuesChangedToken.Value = 0;
+        m_accessibilitySettings->HighContrastChanged -= m_accessibilitySettingsToken;
+        m_accessibilitySettingsToken.Value = 0;
+        Window::Current->Activated -= m_windowActivatedToken;
+        m_windowActivatedToken.Value = 0;
     }
 
-    void TitleBar::SetTitleBarHeight(double height)
+    void TitleBar::SetTitleBarExtenView()
```
danbelcher-MSFT:
```suggestion
    void TitleBar::SetTitleBarExtendView()
```

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -103,14 +99,69 @@ namespace CalculatorApp
...
+            applicationTitleBar->ButtonHoverBackgroundColor = hoverbgColor;
+            applicationTitleBar->ButtonHoverForegroundColor = hoverfgColor;
+            applicationTitleBar->ButtonPressedBackgroundColor = pressedbgColor;
+            applicationTitleBar->ButtonPressedForegroundColor = pressedfgCoolor;
+        }
     }
+
 }
 
+void CalculatorApp::TitleBar::OnHighContrastChanged(_In_ AccessibilitySettings ^ /*sender*/, _In_ Object ^ /*args*/)
```
danbelcher-MSFT: Move these into the `CalculatorApp` namespace above.  Let's keep the style consistent to one linebreak between each function definition.

danbelcher-MSFT:
Sidenote, I'd be perfectly fine if this change also disabled the framerate counter in App.xaml.cpp.  It's covering the titlebar so I can't see your change! :)  In general, I doubt anyone uses it in their day-to-day and if they have need of it, it can be easily enabled.

rudyhuyn:
> Hey Rudy, it looks like this change regresses the position of the text that displays when the docked History/Memory list is empty:

Oops! The worst part is that I saw it when I took the screenshots and i thought to myself: "What a weird UI, we should change it"... 🤦‍♂️🤦‍♂️🤦‍♂️ I will update it tonight

rudyhuyn:
> Sidenote, I'd be perfectly fine if this change also disabled the framerate counter in App.xaml.cpp. It's covering the titlebar so I can't see your change! :) In general, I doubt anyone uses it in their day-to-day and if they have need of it, it can be easily enabled.

Yours is giant! 
image

But I agree, it's not really something we use daily

src/Calculator/Calculator.vcxproj.filters
```diff
@@ -471,15 +463,13 @@
...
-    <Page Include="Views\TitleBar.xaml">
-      <Filter>Views</Filter>
-    </Page>
     <Page Include="Views\MemoryListItem.xaml">
       <Filter>Views</Filter>
     </Page>
     <Page Include="Views\OperatorsPanel.xaml">
       <Filter>Views</Filter>
     </Page>
+    <Page Include="Views\TitleBar.xaml" />
```
danbelcher-MSFT: This should go under the `Views` filter with the rest of the XAML pages.

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -103,14 +99,67 @@ namespace CalculatorApp
...
+            SetTitleBarControlColors();
+            SetTitleBarExtendView();
+            SetTitleBarVisibility();
+        }));
     }
-}
 
+    void TitleBar::OnWindowActivated(_In_ Object ^ /*sender*/, _In_ WindowActivatedEventArgs ^e)
+    {
+        VisualStateManager::GoToState(this, e->WindowActivationState == CoreWindowActivationState::Deactivated ? L"WindowNotFocused" : L"WindowFocused", false);
```
danbelcher-MSFT:
```suggestion
        VisualStateManager::GoToState(this, e->WindowActivationState == CoreWindowActivationState::Deactivated ? WindowNotFocused->Name : WindowFocused->Name, false);
```
Safer :)

danbelcher-MSFT:
I'll be honest, I don't feel like changing the color of the titlebar text makes much of a difference in noticing if the window is focused or not.  Acrylic effect is removed from apps when they lose focus, which is much more noticeable.  I do like the rest of the changes in this PR.

grochocki:
I agree that we should not change the title bar visuals when the acrylic effect is enabled, but there are a number of scenarios the system will disable acrylic (i.e., Battery Saver or when transparency effects are explicilty disabled by the user in settings). In those cases, we should make sure that we have a visual indication of window focus.

danbelcher-MSFT:
I wasn't suggesting that we should not change the title bar visuals when acrylic effect is enabled.  I was only commenting that I personally don't feel that changing the text color from black to gray is that noticeable.  If we really need to indicate window focus as a feature requirement (shouldn't Windows handle this for apps?), it seems like we could go futher, for example all text could gray and we could gray the accent color.  A visual change like that pops across the whole app instead of just the text in the titlebar.  I also realize that this is a much more exhaustive change so I'm not expecting Rudy or anyone else to go implement that.  Anyways, I wasn't trying to be contentious. Just voicing one opinion.

rudyhuyn:
As Dave explained, there are a lot of scenarios where acrylic is not available, so we need to have a way to distinguish if the windows are focused or not.

in my opinion, for acrylic and titlebar, we should still modify the title bar color even if acrylic is activated:
- have a consistent UI whether acrylic is enabled or not (in both cases, the unfocused window will look exactly the same)
- match with colors of minimize, maximize, close buttons
- what if the battery saver turns on while Calculator is not focused?
- have a consistent UI with all other applications not using Acrylic
- make it future proof, Acrylic is currently enabled only when a window is focused (for performance reasons), if the OS team decides to change this behavior in a future update, we would be ready.

But, it's also true that a lot of first party applications using Acrylic don't modify the title color when they lose focus (Settings, Maps, Clock, Movie, Photos, Snip...), but they also have the same issue we are trying to fix, so we should probably not use them as a reference.

I let you decide what you prefer, we are very busy this week with @Turtleship, I can modify the diff tomorrow night with the behavior you prefer.

rudyhuyn:
> I wasn't suggesting that we should not change the title bar visuals when acrylic effect is enabled. I was only commenting that I personally don't feel that changing the text color from black to gray is that noticeable. If we really need to indicate window focus as a feature requirement (shouldn't Windows handle this for apps?), it seems like we could go futher, for example all text could gray and we could gray the accent color. A visual change like that pops across the whole app instead of just the text in the titlebar. I also realize that this is a much more exhaustive change so I'm not expecting Rudy or anyone else to go implement that. Anyways, I wasn't trying to be contentious. Just voicing one opinion.

If you gray everything, users will think that the actions/buttons are disabled, while they are not, users should still be able to distinct an enabled/disabled button even when the app is not focused. If it was necessary to click 2 times to do an action when the window isn't focused (the first click enabling the window but the event not forwarded to the UI) , I would agree, but (luckily) it's not how Windows works, even if the app is not focused, a single click on a button will activate the window and generate a click.

danbelcher-MSFT:
> If you gray everything, users will think that the actions/buttons are disabled, while they are not, users should still be able to distinct an enabled/disabled button even when the app is not focused.
Great point, let's not have that confusion.

> match with colors of minimize, maximize, close buttons
This seems like the best behavior and it's what your PR captures.  I checked with a blank UWP app, and the standard titlebar text changes color the same way we do here.  The other apps that you called out must have custom titlebars.  

> I can modify the diff tomorrow night with the behavior you prefer
No need, I think everyone is fine with this change.  I did have some feedback to be addressed and then we can get this merged.

src/Calculator/Views/TitleBar.xaml.cpp
rudyhuyn: 👍

mdtauk:
Once Microsoft implements the depth shadows with the windowing, that will help make the focused and unfocused Windows more noticeable I think.

DragoCubed:
Windows already has a larger/darker shadow for active windows. Windows needs more than that though.

mdtauk:
> Windows already has a larger/darker shadow for active windows. Windows needs more than that though.

I know they have some shadowing there, but I am sure I saw it mentioned at Build, that there are plans to bring in the new depth shadows that have been added into XAML for the actual shell windows.

src/Calculator/Calculator.vcxproj.filters
danbelcher-MSFT: @rudyhuyn, thoughts?

danbelcher-MSFT:
@rudyhuyn, I have two comments at the top of the thread that I'm still wondering your thoughts on.  Looks like there's also merge conflicts that should be fixed up.  Almost ready to check this in.

src/Calculator/Calculator.vcxproj.filters
rudyhuyn: You are right! (if only VS were able to handle file creation and filters a little better)

src/Calculator/Views/MainPage.xaml
```diff
@@ -37,12 +36,11 @@
...
     <Grid>
         <Grid.RowDefinitions>
-            <RowDefinition Height="Auto"/>
             <!--
-                This row is padding for the SystemFocusVisuals,
+                This row has a min height of 3 for the SystemFocusVisuals,
                 otherwise the focus rectangles render under the title bar controls.
             -->
-            <RowDefinition x:Name="FocusVisualMargin" Height="3"/>
+            <RowDefinition Height="Auto" MinHeight="3"/>
```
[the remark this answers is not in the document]
rudyhuyn: yes, it doesn't seem necessary anymore (tested in tablet mode, high-constrast, etc...)

danbelcher-MSFT:
Hey @rudyhuyn, we're ready to merge this change but there's conflicts to resolve.

rudyhuyn:
> Hey @rudyhuyn, we're ready to merge this change but there's conflicts to resolve.

thanks for the alert, I will merge tonight!

src/Calculator/Calculator.vcxproj.filters
danbelcher-MSFT: I agree...

PR #444, opened 2019-04-07
Add support for Pyeong, a Korean floorspace unit.

### LINKED ISSUE #382: Adding support for Pyeong, a floorspace unit specific to Korea
**Problem Statement**

Pyeong (abbreviation py) is a Korean unit of area and floorspace ([wikipedia](https://en.wikipedia.org/wiki/Pyeong)). The unit is the standard traditional measure for real estate floorspace - we use it whenever we talk about living space, office space, or even a city size.

Microsoft Calculator currently doesn't support it, and it would be great if it supports the unit.
Our target audience are Koreans, people living in Korea, and people looking at Korean real estates.

**Proposal**
Let's display Pyeong under **Converter -> Length**  when a user's current locale is Korean.

**Goals**
User can convert Pyeong to other area units.

**Non-Goals**
Supporting other units specific to Korea. I would like to keep the PR for this issue simple.

**Low-Fidelity Concept**
Code change should be fairly simple.
An approach I have in mind is to modify [UnitConverterDataLoader.cpp
](https://github.com/Microsoft/calculator/blob/1113ff4b8673b1dc59d8da91ae3189905b9199d9/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp) Specifically, I would modify `GetUnits` method as below

```
void UnitConverterDataLoader::GetUnits(_In_ unordered_map<ViewMode, vector<OrderedUnit>>& unitMap)
{
   ...
   bool useKoreanCustomary = m_currentRegionCode == L"KP" || m_currentRegionCode == L"KR";
   ...
   areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Pyung, GetLocalizedStringName(L"UnitName_Pyeong"), GetLocalizedStringName(L"UnitAbbreviation_Pyeong"), ...});
   ...
}
```

_Note: Once this feature request gets approved, I would like to follow up with a PR._

TurtleShip:
- Add Pyeong as an Area conversion unit.
- Pyeong shows up only if the user's current **region** is Korea ( i.e. region is either `KP` or `KR` ).
- Added Korean translation for Pyeong (평). For other locales, we default to English ( Pyeong ).

Manually tested the below
1) For non-Korean regions, Pyeong does not show up.

2) Korean region with Korean locale => Pyeong shows up and Pyeong is correctly translated.
pyeong_Korean

3) Korean region with English locale => Pyeong shows up and Pyeong is in English.
pyeong_English

4) Korean region with simplified Chinese locale => Pyeong shows up and Pyeong is in English.
pyeong_Chinese

### DIFF
--- a/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
+++ b/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
@@ -120,6 +120,12 @@ void UnitConverterDataLoader::LoadData()
 
                 for (auto itr = unitConversions.begin(); itr != unitConversions.end(); ++itr)
                 {
+                    if (idToUnit.find(itr->first) == idToUnit.end())
+                    {
+                        // Optional units will not be in idToUnit but can be in unitConversions.
+                        // For optional units that did not make it to the current set of units, just continue.
+                        continue;
+                    }
                     UCM::ConversionData parsedData = { 1.0, 0.0, false };
                     assert(itr->second > 0); // divide by zero assert
                     parsedData.ratio = unitFactor / itr->second;
@@ -175,6 +181,10 @@ void UnitConverterDataLoader::GetUnits(_In_ unordered_map<ViewMode, vector<Order
 
     bool useWattInsteadOfKilowatt = m_currentRegionCode == "GB";
 
+    // Use Pyeong, a Korean floorspace unit.
+    // https://en.wikipedia.org/wiki/Korean_units_of_measurement#Area
+    bool usePyeong = m_currentRegionCode == L"KP" || m_currentRegionCode == L"KR";
+    
     vector<OrderedUnit> areaUnits;
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Acre, GetLocalizedStringName(L"UnitName_Acre"), GetLocalizedStringName(L"UnitAbbreviation_Acre"), 9 });
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Hectare, GetLocalizedStringName(L"UnitName_Hectare"), GetLocalizedStringName(L"UnitAbbreviation_Hectare"), 4 });
@@ -190,6 +200,10 @@ void UnitConverterDataLoader::GetUnits(_In_ unordered_map<ViewMode, vector<Order
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Paper, GetLocalizedStringName(L"UnitName_Paper"), GetLocalizedStringName(L"UnitAbbreviation_Paper"), 12, false, false, true });
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_SoccerField, GetLocalizedStringName(L"UnitName_SoccerField"), GetLocalizedStringName(L"UnitAbbreviation_SoccerField"),13, false, false, true });
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Castle, GetLocalizedStringName(L"UnitName_Castle"), GetLocalizedStringName(L"UnitAbbreviation_Castle"), 14, false, false, true });
+    if (usePyeong)
+    {
+        areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Pyeong, GetLocalizedStringName(L"UnitName_PYeong"), GetLocalizedStringName(L"UnitAbbreviation_Pyeong"), 15, false, false, false });
+    }
     unitMap.emplace(ViewMode::Area, areaUnits);
 
     vector<OrderedUnit> dataUnits;
@@ -384,6 +398,7 @@ void UnitConverterDataLoader::GetConversionData(_In_ unordered_map<ViewMode, uno
             { ViewMode::Area, UnitConverterUnits::Area_Paper, 0.06032246 },
             { ViewMode::Area, UnitConverterUnits::Area_SoccerField, 10869.66 },
             { ViewMode::Area, UnitConverterUnits::Area_Castle, 100000 },
+            { ViewMode::Area, UnitConverterUnits::Area_Pyeong, 3.305785 }, // apprx of 400/121
 
             { ViewMode::Data, UnitConverterUnits::Data_Bit, 0.000000125 },
             { ViewMode::Data, UnitConverterUnits::Data_Byte, 0.000001 },
--- a/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
+++ b/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
@@ -203,7 +203,7 @@ void UnitConverterDataLoader::GetUnits(_In_ unordered_map<ViewMode, vector<Order
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Castle, GetLocalizedStringName(L"UnitName_Castle"), GetLocalizedStringName(L"UnitAbbreviation_Castle"), 14, false, false, true });
     if (usePyeong)
     {
-        areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Pyeong, GetLocalizedStringName(L"UnitName_PYeong"), GetLocalizedStringName(L"UnitAbbreviation_Pyeong"), 15, false, false, false });
+        areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Pyeong, GetLocalizedStringName(L"UnitName_Pyeong"), GetLocalizedStringName(L"UnitAbbreviation_Pyeong"), 15, false, false, false });
     }
     unitMap.emplace(ViewMode::Area, areaUnits);
 
--- a/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
+++ b/src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
@@ -118,18 +118,19 @@ void UnitConverterDataLoader::LoadData()
                 unordered_map<int, double> unitConversions = categoryToUnitConversionDataMap.at(categoryViewMode);
                 double unitFactor = unitConversions[unit.id];
 
-                for (auto itr = unitConversions.begin(); itr != unitConversions.end(); ++itr)
+                 for (const auto&[id, conversionFactor] : unitConversions)
                 {
-                    if (idToUnit.find(itr->first) == idToUnit.end())
+                    if (idToUnit.find(id) == idToUnit.end())
                     {
                         // Optional units will not be in idToUnit but can be in unitConversions.
                         // For optional units that did not make it to the current set of units, just continue.
                         continue;
                     }
+
                     UCM::ConversionData parsedData = { 1.0, 0.0, false };
-                    assert(itr->second > 0); // divide by zero assert
-                    parsedData.ratio = unitFactor / itr->second;
-                    conversions.insert(pair<UCM::Unit, UCM::ConversionData>(idToUnit.at(itr->first), parsedData));
+                    assert(conversionFactor > 0); // divide by zero assert
+                    parsedData.ratio = unitFactor / conversionFactor;
+                    conversions.insert(pair<UCM::Unit, UCM::ConversionData>(idToUnit.at(id), parsedData));
                 }
             }
             else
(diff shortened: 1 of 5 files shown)


src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -384,6 +398,7 @@ void UnitConverterDataLoader::GetConversionData(_In_ unordered_map<ViewMode, uno
             { ViewMode::Area, UnitConverterUnits::Area_Paper, 0.06032246 },
             { ViewMode::Area, UnitConverterUnits::Area_SoccerField, 10869.66 },
             { ViewMode::Area, UnitConverterUnits::Area_Castle, 100000 },
+            { ViewMode::Area, UnitConverterUnits::Area_Pyeong, 3.305785 }, // apprx of 400/121
```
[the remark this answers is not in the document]
TurtleShip: Thanks, I updated it! 😄

TurtleShip:
Thanks @rudyhuyn !

So I've been reading through how Calculator preferences work, and I noticed 1 bug and 1 possible improvement. But before I go into them, I would like to note that `ignore an optional unit if not available for the current culture` already happens through `UnitConverter::InitializeSelectedUnits()`.

So in case a user choose Pyeong, closes Calculator, changes the current region to non-Korea (e..g US), and starts up Calculator, the app won't attempt to load Pyeong. In fact, at the moment, the app will fall back to the default conversion values ( it looks like that may not be intended behaviour though - I think the current intended behaviour, at least from the code, was to show no unit (`EMPTY_UNIT` and let the user pick one ).

Circling back to the bug & improvement I mentioned earlier, here are what I found.
I also uploaded demo videos [here](https://www.dropbox.com/sh/metv3rc2hyzivnj/AACHVzL9_ZT7Bg2wIkWDCqsEa?dl=0) for bug & possible improvement. Gif I attached in the comment are of poor quality so I recommend viewing the videos.

### Bug 1: saving user preference doesn't seem to work correctly.
Specifically,
1) `UnitConverter::RestoreUserPreferences` is working correctly.
2) `UnitConverter::SaveUserPreferences()` is working correctly.
3) However, after restoring saved values on calculator startup, Something is overriding the restored value to the default value ☹️ 
Here is a demo
preferences_bug_demo

### Possible improvement 1: Instead of showing no unit, we should display a default value (or the first available value if any) if we fail to find units that we saved from last session.
So this behaviour where we don't show any unit probably never really got observed in real life because of Bug 1. I made a little tweak, and the below is what is looks like if we actually defaulted to `EMPTY_UNIT`.
cal_no_units

I am not sure if the above was a product decision though ( hence I labeled it as a possible improvement instead of a bug ).

@grochocki What are your opinions on the above mentioned bug & possible improvement?
I can file them separately and make PRs as well unless of course they are known issues and already being addressed.

Cheers!

TurtleShip:
> [CLA assistant check](https://cla.opensource.microsoft.com/Microsoft/calculator?pullRequest=444) 
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.microsoft.com/Microsoft/calculator?pullRequest=444) before we can accept your contribution.
> 
> ❌ TurtleShip [sign now](https://cla.opensource.microsoft.com/Microsoft/calculator?pullRequest=444)
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.microsoft.com/check/Microsoft/calculator?pullRequest=444) it.

Hmm, I tried 3 times already with an hour or so interval, but `I agree` button seems perma disabled. I can't click it. 😢

src/Calculator/Resources/ko-KR/Resources.resw
```diff
@@ -2907,4 +2907,12 @@
...
     <comment>Displayed on a link to the Microsoft Services Agreement in the about this app information</comment>
   </data>
+  <data name="UnitAbbreviation_Pyeong" xml:space="preserve">
+    <value>평</value>
+    <comment>An abbreviation for a measurement unit of area.</comment>
+  </data>
+  <data name="UnitName_Pyeong" xml:space="preserve">
+    <value>평</value>
+    <comment>A measurement unit for area.</comment>
+  </data>
```
grochocki: Unfortunately, our Localizaton system is not yet able to directly accept localization changes directly through this repo, so we should only add the english resource.

grochocki:
> So in case a user choose Pyeong, closes Calculator, changes the current region to non-Korea (e..g US), and starts up Calculator, the app won't attempt to load Pyeong. In fact, at the moment, the app will fall back to the default conversion values

We do not expect users to change region often, so resetting to the defaults is fine.

> What are your opinions on the above mentioned bug & possible improvement?
I can file them separately and make PRs as well unless of course they are known issues and already being addressed.

Good catch! There is not anything tracking this, so lets create a new issue and handle that separately from this PR. 

IRT CLA agreement issues, I would maybe try again and see if it was just a temporary issue. Otherwise, @HowardWolosky, do you have any ideas?

rudyhuyn:
> So in case a user choose Pyeong, closes Calculator, changes the current region to non-Korea (e..g US), and starts up Calculator, the app won't attempt to load Pyeong.

Luckily (or unfortunately) `InitializeSelectedUnits` is called just after `RestoreUserPreferences`, resetting `FromUnit` and `ToUnit` to their default values. It seems to not be an intended behavior, this is due to  the boolean `m_IsFirstTime` and `OnCategoryChanged` called twice when the application starts.

Any future PR (to fix the restoration or any changes on `UnitConverter`) can modify this behavior, so it 's better to prevent `FromUnit` and `ToUnit` to be set to an unsupported unit when we restore them. The same issue would happen if we decide a unit is removed by an update, the app would unfortunately crash.

image

src/Calculator/Resources/ko-KR/Resources.resw
TurtleShip: Gotcha. I removed translations for Korean :)

TurtleShip:
@grochocki 
Looks like I have triggered an abuse detection mechanism.
I will try signing it again tmrw

image

TurtleShip:
@rudyhuyn  I addressed your comment regarding not restoring from saved units if they are not valid with the current available units!

TurtleShip:
@grochocki 
I created issue for not restoring user preferences #445 .

I haven't created one for replacing `EMPTY_UNIT` with a valid unit. Shall I file it under Bug or Feature request?
Also, this issue where we may show no unit when calculator launches (because of `EMPTY_UNIT`) is not currently happening, but will start happening once #445 is fixed. Is there a way to strongly suggest to fix this issue first before #445, or concurrently?

Cheers!

rudyhuyn:
> Also, this issue where we may show no unit when calculator launches (because of `EMPTY_UNIT`) is not currently happening, but will start happening once #445 is fixed. Is there a way to strongly suggest to fix this issue first before #445, or concurrently?
> 
> Cheers!

If `UnitConverter::RestoreUserPreference` isn't able to set `m_fromType` or `m_toType` (because the saved value is Empty or the unit not available), we should not modify them and keep values set by `InitializeSelectedUnits`. IMO, if the issue is currently not happening, it should not change with #445.

TurtleShip:
Thanks, @rudyhuyn !

So to confirm my understanding of what you said, an issue where no unit is selected should not occur  after fixing user preferences issue because of the below flow?
1) Calculator fails to restore saved values ( either there are no saved values or saved values are not valid )
2) Calculator will select default values ( this seems to be the case from manual testing but I haven't confirmed in code )
3) `InitializeSelectedUnits` will choose default values

rudyhuyn:
I'm AFK, but in my memory `InitializeSelectedUnits` is first called, then `RestoreUserPreference`, so if we can't restore, we just need to keep the current values as is. The issue is that OnCategoryChanged is called twice and the way we detect if the category has been changed by the code or by the user (via UI) is wrong (see code related to `m_isFirstTime`). Let's chat about it tomorrow IRL 😀📦

TurtleShip:
For prosperity, @rudyhuyn and I talked and we agree that an issue where no unit is selected is not a concern ( my understanding of  `InitializeSelectedUnits` was incorrect ).

So to sum up, the only real issue is preferences not being correctly restored, and we created issue for it -  #445.

TurtleShip:
btw, is there a further action I need to take for review? 😄

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -120,6 +120,12 @@ void UnitConverterDataLoader::LoadData()
 
                 for (auto itr = unitConversions.begin(); itr != unitConversions.end(); ++itr)
```
danbelcher-MSFT: There's an opportunity to improve the readability of the code here with:
```suggestion
                 for (const auto&[id, conversionFactor] : unitConversions)
```
Then replace `itr->first` with `id` and `itr->second` with `conversionFactor`.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -3379,4 +3379,12 @@
     <value>Microsoft Services Agreement</value>
     <comment>Displayed on a link to the Microsoft Services Agreement in the about this app information</comment>
   </data>
-</root>
+  <data name="UnitAbbreviation_Pyeong" xml:space="preserve">
+    <value>Pyeong</value>
```
danbelcher-MSFT: Wikipedia lists the abbreviation as `py`.  Are you sure we should use `Pyeong` explicitly?  Americans would recognize `ft` as `feet`.  Would Koreans recognize `py` as `Pyeong`?

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -190,6 +200,10 @@ void UnitConverterDataLoader::GetUnits(_In_ unordered_map<ViewMode, vector<Order
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Paper, GetLocalizedStringName(L"UnitName_Paper"), GetLocalizedStringName(L"UnitAbbreviation_Paper"), 12, false, false, true });
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_SoccerField, GetLocalizedStringName(L"UnitName_SoccerField"), GetLocalizedStringName(L"UnitAbbreviation_SoccerField"),13, false, false, true });
     areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Castle, GetLocalizedStringName(L"UnitName_Castle"), GetLocalizedStringName(L"UnitAbbreviation_Castle"), 14, false, false, true });
+    if (usePyeong)
+    {
+        areaUnits.push_back(OrderedUnit{ UnitConverterUnits::Area_Pyeong, GetLocalizedStringName(L"UnitName_PYeong"), GetLocalizedStringName(L"UnitAbbreviation_Pyeong"), 15, false, false, false });
```
danbelcher-MSFT: You're using `UnitName_PYeong` but the resource key is `UnitName_Pyeong`.  Looks like it still works, but we should fix the typo anyways.

src/CalcViewModel/DataLoaders/UnitConverterDataLoader.cpp
```diff
@@ -120,6 +120,12 @@ void UnitConverterDataLoader::LoadData()
 
                 for (auto itr = unitConversions.begin(); itr != unitConversions.end(); ++itr)
                 {
+                    if (idToUnit.find(itr->first) == idToUnit.end())
+                    {
+                        // Optional units will not be in idToUnit but can be in unitConversions.
+                        // For optional units that did not make it to the current set of units, just continue.
+                        continue;
+                    }
```
danbelcher-MSFT: Add a blank line after closing a scope

src/Calculator/Resources/en-US/Resources.resw
TurtleShip: Hmm, good question.
I am not sure if Koreans would recognize `py` mainly because they would just write it in Korean ( 평 ).
I don't feel strongly one way or the other though.
Would you like me to update it to `py`?

TurtleShip:
Hi, @danbelcher-MSFT 

Thanks for your reviews! :) 
I've addressed your comments and left one clarifying question.

src/Calculator/Resources/en-US/Resources.resw
danbelcher-MSFT: If we're uncertain, let's go with the full name.

PR #447, opened 2019-04-08
Clicking on the same element in the hamburger view should re-open that view, like the menu items do

### LINKED ISSUE #437: Clicking on the same element in the hamburger view should re-open that view, like the menu items do
**Problem Statement**

The pain point is that when opening the hamburger menu, it is unclear what view I am currently on. Re-selecting the same view that I'm already on (e.g. Currency) does nothing. but the menu item registers a click animation. Clicking out of the hamburger menu reveals that I'm already on that view.

**Evidence or User Insights**

When selecting a menu item that has already been selected (e.g. US Dollars) the menu item closes, even though it has already been selected as a conversion currency. This behavior should be more universal within the app.

Microsoft Outlook, Gmail, and Github desktop allow re-selecting an item from the hamburger menu that is already open and it will return you to the same view while closing the menu. I did some very primitive research and found a paper which concluded that clicking on something  should make progress towards an existing goal according to (Wang 2008, 409) and other criteria such as discover-ability. However, the size of the study was very small (n=5) so the results may not be statistically significant.

Additionally, if the menu item does close when it is clicked and is in the same view, this might not cause the user to perform a needed context switch when they switch to a different view if the view does not change but the menu item closes.

Wang X. (2008) Design and Evaluation of Intelligent Menu Interface through Cognitive Walkthrough Procedure and Automated Logging for Management Information System. In: Shen W., Yong J., Yang Y., Barthès JP.A., Luo J. (eds) Computer Supported Cooperative Work in Design IV. CSCWD 2007. Lecture Notes in Computer Science, vol 5236. Springer, Berlin, Heidelberg

**Proposal**

This will make the menu items more universal, as they would behave more similar. If someone is converting multiple values with different units, they might forget what the last unit that they converted was; trying to re-select the same menu item will do nothing.

**Goals**

Goal is for the user to be able to select a menu item and go to that view, even if they are already on that view. 

**Non-Goals**

Right-clicking on a menu item should not close the menu because it could have been an accidental click.

**Low-Fidelity Concept**

cheezwhines:
## Fixes #437.
Clicking on the same element in the hamburger view should re-open that view, like the menu items do

-Fixed the bug that was listed

-manual

### DIFF
--- a/src/Calculator/Views/MainPage.xaml.cpp
+++ b/src/Calculator/Views/MainPage.xaml.cpp
@@ -604,9 +604,5 @@ void MainPage::AnnounceCategoryName()
 
 void MainPage::OnNavItemInvoked(MUXC::NavigationView^ /*sender*/, _In_ MUXC::NavigationViewItemInvokedEventArgs^ e)
 {
-    auto item = dynamic_cast<MUXC::NavigationViewItem^>(e->InvokedItemContainer);
-    if (item != nullptr)
-    {
-        NavView->IsPaneOpen = false;
-    }
+    NavView->IsPaneOpen = false;
 }
--- a/src/Calculator/Views/MainPage.xaml.cpp
+++ b/src/Calculator/Views/MainPage.xaml.cpp
@@ -607,11 +607,6 @@ void MainPage::OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationVi
     auto item = dynamic_cast<MUXC::NavigationViewItem^>(e->InvokedItemContainer);
     if (item != nullptr)
     {
-        auto selectedItem = static_cast<NavCategory^>(item->DataContext);
-
-        if (selectedItem->Mode == Model->Mode)
-        {
-            NavView->IsPaneOpen = false;
-        }
+        NavView->IsPaneOpen = false;
     }
 }
--- a/src/Calculator/Views/MainPage.xaml.cpp
+++ b/src/Calculator/Views/MainPage.xaml.cpp
@@ -602,7 +602,7 @@ void MainPage::AnnounceCategoryName()
     NarratorNotifier->Announce(announcement);
 }
 
-void MainPage::OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewItemInvokedEventArgs^ e)
+void MainPage::OnNavItemInvoked(MUXC::NavigationView^ /*sender*/, _In_ MUXC::NavigationViewItemInvokedEventArgs^ e)
 {
     auto item = dynamic_cast<MUXC::NavigationViewItem^>(e->InvokedItemContainer);
     if (item != nullptr)
--- a/src/Calculator/Views/MainPage.xaml.cpp
+++ b/src/Calculator/Views/MainPage.xaml.cpp
@@ -601,3 +601,17 @@ void MainPage::AnnounceCategoryName()
     NarratorAnnouncement^ announcement = CalculatorAnnouncement::GetCategoryNameChangedAnnouncement(categoryName);
     NarratorNotifier->Announce(announcement);
 }
+
+void MainPage::OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewItemInvokedEventArgs^ e)
+{
+    auto item = dynamic_cast<MUXC::NavigationViewItem^>(e->InvokedItemContainer);
+    if (item != nullptr)
+    {
+        auto selectedItem = static_cast<NavCategory^>(item->DataContext);
+
+        if (selectedItem->Mode == Model->Mode)
+        {
+            NavView->IsPaneOpen = false;
+        }
+    }
+}
--- a/src/Calculator/Views/MainPage.xaml.h
+++ b/src/Calculator/Views/MainPage.xaml.h
@@ -52,8 +52,7 @@ namespace CalculatorApp
         void OnNavPaneOpened(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Platform::Object^ args);
         void OnNavPaneClosed(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Platform::Object^ args);
         void OnNavSelectionChanged(_In_ Platform::Object^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewSelectionChangedEventArgs^ e);
-        void OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewItemInvokedEventArgs^ e);
-
+        void OnNavItemInvoked(Microsoft::UI::Xaml::Controls::NavigationView^ /*sender*/, _In_ Microsoft::UI::Xaml::Controls::NavigationViewItemInvokedEventArgs^ e);
 
         void OnAboutButtonClick(_In_ Platform::Object^ sender, _In_ Windows::UI::Xaml::Controls::ItemClickEventArgs^ e);
         void OnAboutFlyoutOpened(_In_ Platform::Object^ sender, _In_ Platform::Object^ e);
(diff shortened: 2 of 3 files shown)


src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -601,3 +601,17 @@ void MainPage::AnnounceCategoryName()
     NarratorAnnouncement^ announcement = CalculatorAnnouncement::GetCategoryNameChangedAnnouncement(categoryName);
     NarratorNotifier->Announce(announcement);
 }
+
+void MainPage::OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewItemInvokedEventArgs^ e)
```
rudyhuyn: Can you remove `Microsoft::UI::Xaml::Controls::`? 
(I would like Visual Studio to automatically check if `using namespace Microsoft::UI::Xaml::Controls` exists, but unfortunately, it's not the case :( )

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -601,3 +601,17 @@ void MainPage::AnnounceCategoryName()
...
+{
+    auto item = dynamic_cast<MUXC::NavigationViewItem^>(e->InvokedItemContainer);
```
rudyhuyn: I think it's safe to simplify this code and always call `NavView->IsPaneOpen = false;` without checking the value of Mode.

What's your opinion?

rudyhuyn:
> ## Fixes #.
> Clicking on the same element in the hamburger view should re-open that view, like the menu items do #437
> 
> ### Description of the changes:
> ## -Fixed the bug that was listed
> 
> ### How changes were validated:
> ## -manual

Congratulation on your first PR!

The PR description has some formatting issue, here is the fixed version:

```
## Fixes #437
Clicking on the same element in the hamburger view should re-open that view, like the menu items do. 
-Fixed the bug that was listed 
-manual
```

I also let some comments, don't hesitate to comment back if you have questions/feedback! 

Good work!

src/Calculator/Views/MainPage.xaml.cpp
cheezwhines: I agree with this.

src/Calculator/Views/MainPage.xaml.cpp
cheezwhines: I was not able to remove `Microsoft::UI::Xaml::Controls`.

src/Calculator/Views/MainPage.xaml.cpp
rudyhuyn: Great! The less we use `DataContext` in the code, the better!

src/Calculator/Views/MainPage.xaml.cpp
rudyhuyn: Because the namespace is already declared , you can write:

`void MainPage::OnNavItemInvoked(_In_ NavigationView^ sender, _In_ NavigationViewItemInvokedEventArgs^ e)`

cheezwhines: I tried that. For some reason I got errors.

danbelcher-MSFT:
I think I'm considering this a bug in the NavigationView control.  We can take the fix here and I'll pass on the bug to the controls team.

src/Calculator/Views/MainPage.xaml.h
```diff
@@ -52,6 +52,8 @@ namespace CalculatorApp
         void OnNavPaneOpened(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Platform::Object^ args);
         void OnNavPaneClosed(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Platform::Object^ args);
         void OnNavSelectionChanged(_In_ Platform::Object^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewSelectionChangedEventArgs^ e);
+        void OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender, _In_ Microsoft::UI::Xaml::Controls::NavigationViewItemInvokedEventArgs^ e);
+
```
danbelcher-MSFT: nit: please remove this extra whitespace.  We'll keep just one blank line separating blocks of functions.

src/Calculator/Views/MainPage.xaml.cpp
rudyhuyn: @danbelcher-MSFT, I think you removed the `_In_` by mistake

`void MainPage::OnNavItemInvoked(_In_opt_  MUXC::NavigationView^ /*sender*/, _In_ MUXC::NavigationViewItemInvokedEventArgs^ e)`

danbelcher-MSFT: Actually, it was intentional.  `_In_opt_` describes a parameter that may or may not have a value for use by the function.  We are explicitly discarding this parameter so I figured there's no point to decorating it with SAL (imo).  `_In_opt_` doesn't seem to be the right fit and there isn't any other SAL equivalent.

rudyhuyn: This is a (very) tricky question and open to debate.

Imo, I don't think it's a best practice to comment/discard a parameter in a function declaration (but it's totally fine with function definition), it makes the development harder when using IntelliSense (because the parameter is not named):

image

Another issue is virtual methods, if the method is overridden, the developer have no idea if the parameter is In, Out, In/Out, optional, etc...

```
class TestBase
{
public:
    virtual int TestFunc(int * /*test*/, _In_ int * test2) noexcept;
};

class Test: public TestBase
{
public:
    virtual int TestFunc(_InOut_ int * test, _In_ int * test2) noexcept;
};
```

in this case, a developer will have no idea if the first parameter is in/out/inout when using a TestBase* object.

For these 2 reasons (and probably more), I think we should not comment/discard a argument in a function declaration, so we need to use SAL for this parameter in the declaration.

Visual Studio forcing SAL to be in the function definition when set in the declaration, we are forced to do it in the function definition too.

So in our case:

.h
```
void OnNavItemInvoked(_In_ Microsoft::UI::Xaml::Controls::NavigationView^ sender...)
```

.cpp
```
void MainPage::OnNavItemInvoked(_In_ NavigationView^ /*sender.*/..)
```

danbelcher-MSFT: On the other hand, if we're consistent about naming parameters that are used and not naming parameters that are unused, then seeing a function with an unnamed parameter is a clear indication that the parameter is not used by that function.  In this case, I can simply supply `nullptr`.  Given that SAL describes how a parameter is used, for an unnamed parameter there would be no need for SAL.

This doesn't feel like an appropriate place for a long discussion of the semantics (I can see arguments for both sides).  In this case, let's let @cheezwhines resolve as they'd like and we revisit this elsewhere, if we have need.

danbelcher-MSFT:
Controls bug is tracked by Microsoft/microsoft-ui-xaml#573

danbelcher-MSFT:
Hey @cheezwhines, can you address @rudyhuyn's simplification he suggested here? https://github.com/Microsoft/calculator/pull/447#discussion_r273354119

src/Calculator/Views/MainPage.xaml.cpp
rudyhuyn: > then seeing a function with an unnamed parameter is a clear indication that the parameter is not used by that function.

This isn't true with virtual methods, a unused parameter in a base class can be in reality used by a function overriding it in a child class.

```
class WaterParkBase
{
public:
    virtual int GetPrice(_In_ int * numberAdults, int * /*numberKids*/) noexcept
    {
          m_price = *numberAdults * 50;
    }
...
};

class SFWaterPark: public c
{
public:
    virtual int GetPrice(_In_ int * numberAdults, _In_ int * numberKids) noexcept
    {
          m_price = *numberAdults * 35 + *numberKids *10;
    }
};
```
in this case:

```
SFWaterPark t = new SFWaterPark();
t->GetPrice(a, b);
``` 

Intellisense will show `GetPrice(int * numberAdults, int *)` even if `numberKids` will be used.

danbelcher-MSFT: Hey Rudy, I hear your perspective and I think I agree we should be consistent in naming parameters for declarations but remove them for definitions.  For now I'm going to take the change as is, to prevent more work for the author.  We'll endeavour to be consistent about this rule for future changes.  Thanks for sharing the examples!

Issue #428, opened 2019-04-04
Update Calculator keyboard for greater consistency and extensibility


MicrosoftIssueBot:
> This feature is in pre-production. Contribute to the in-progress spec **[here](https://github.com/Microsoft/calculator-specs/tree/keyboardrefresh/active/KeyboardRefresh)**.

**Problem Statement**
The Calculator keyboard requires some changes to meet the demands of upcoming features, including graphing mode and always-on-top mode. Furthermore, the current standard keyboard could be better at accommodating scaled screen sizes, which improves overall accessibility of the app. There are a number of other problems this spec seeks to address, including reducing confusion over the up arrow used as the shift key and reducing the number of keyboard layout states.

**Evidence or User Insights**
The current keyboard is not scalable to support new minimum size requirements for CompactOverlay mode, nor is it extensible to support additional functionality in graphing calculator.

**Proposal**
Update they keyboard app-wide for increased consistency and greater extensibility to prepare for upcoming feature improvements.

**Goals**
* No functionality takebacks in existing calculator modes
* Opportunistically add missing functionality across modes
* Increase clarity around shifted operations

**Non-Goals**
* Changes to display option buttons are out of scope (e.g., BYTE/WORD mode in programmer calculator), unless the button is better suited to be included in the keyboard

**Low-Fidelity Concept**
![image](https://user-images.githubusercontent.com/1414423/55523781-7ce7b880-563f-11e9-8a71-3359ba754766.png)

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

rudyhuyn:
The concept already looks better than the current version (especially font sizes and the EQU button using the accent color).

What's the purpose of the `2nd` button?

Do you plan to support long-press actions (or right click if used with a mouse) to display more options? For example:
- long-press on `Mod` would display `Rem`
- long-press on "Pi" would display more Mathematical constant (some examples here: https://en.wikipedia.org/wiki/Mathematical_constant)
- long-press on `square root` would display `cube root`
etc...

grochocki:
> What's the purpose of the 2nd button?

Updated the mock above to show what 2nd button in action.

> Do you plan to support long-press actions (or right click if used with a mouse) to display more options?

Interesting question! I think some of these can be addressed by 2nd button. I am working to publish more details on my thinking here soon.

michael-hawker:
[...]

PR #449, opened 2019-04-10
Simplify some of the calc engine string logic

joseartrivera:
For #428 some new buttons/functions will need to be added. In preparation, this PR attempts to make that process easier.

Currently Calculator handles strings by defining integers for each type of function that can be performed, this integer will eventually correspond with an index in s_engineStrings which holds the corresponding display string for each function. Some functions such as Sin can have multiple strings (degrees, rads, grads, inverse). Functions like Sin are mapped to another array called "rgUfne" where a new integer is given depending on the output string which will then be given to s_engineStrings. The new integer returned by the "rgUfne" array runs the risk of overlapping with any new functions that may be added in CCommand.h. Furthermore, it is expected that the strings in s_engineStrings and rgUfne are defined in a particular order (not necessarily sequential), otherwise the logic will break. This makes adding new strings for new functions confusing and difficult, since a lot of the logic is not clearly defined.

This PR attempts to make this a bit simpler by changing the s_engineStrings and rgUfne arrays to be unordered_maps instead of arrays. For s_engineStrings the keys will now be strings, allowing the existing logic for indexing to be used by simply converting the number into a string to access the value. This will also allow us to create keys in the future that are not limited to integers but to strings that hold more meaning.

The rgUfne array will also be updated to be a map that will take in an integer and give you the corresponding string that can be passed to s_engineStrings. The UFNE object in the rgUfne array will also be updated to hold all the possible string keys for a function, instead of indexing them on other numbers that may overlap with existing definitions.

Now to add a new string for a new IDC_FOO function, we would just need to add the "FooString" resource keys to the g_sids array and use the updated rgUfne map to link the IDC_FOO value to the corresponding "FooString" resource key. This way the resource key can be a meaningful string, and not an integer that must be in any particular order.

Tested each function manually in standard, scientific, and programmer modes.


src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -870,153 +863,91 @@ void CCalcEngine::DisplayAnnounceBinaryOperator()
...
+    wstring degreeString; // Used by default if there are no rad or grad specific strings.
+    wstring inverseDegreeString; // Will fall back to degreeString if empty
+
+    wstring radString;
+    wstring inverseRadString; // Will fall back to radString if empty
+
+    wstring gradString;
+    wstring inverseGradString; // Will fall back to gradString if empty
+
+    bool hasAngleStrings = ((!radString.empty()) || (!inverseRadString.empty()) || (!gradString.empty()) || (!inverseGradString.empty()));
```
danbelcher-MSFT: If just one of these strings is not empty, it's sufficient to say the struct `hasAngleStrings`?  Do `degreeString` and `inverseDegreeString` need to be checked here as well?

src/CalcManager/CEngine/calc.cpp
```diff
@@ -25,13 +25,18 @@ static constexpr wstring_view DEFAULT_NUMBER_STR = L"0";
 // Read strings for keys, errors, trig types, etc.
 // These will be copied from the resources to local memory.
 
-array<wstring, CSTRINGSENGMAX> CCalcEngine::s_engineStrings;
+unordered_map<wstring, wstring> CCalcEngine::s_engineStrings;
 
 void CCalcEngine::LoadEngineStrings(CalculationManager::IResourceProvider& resourceProvider)
 {
-    for (size_t i = 0; i < s_engineStrings.size(); i++)
+    for (size_t i = 0; i < g_sids.size(); i++)
```
danbelcher-MSFT:
```suggestion
    for (const auto& locKey : g_sids)
```

src/CalcManager/Header Files/EngineStrings.h
```diff
@@ -214,7 +165,8 @@
 #define SIDS_ERR_INPUT_OVERFLOW L"119"
 #define SIDS_ERR_OUTPUT_OVERFLOW L"120"
 
-__declspec(selectany) std::wstring g_sids[] =
+// Include the resource key ID from above into this vector to load it into memory for the engine to use
+__declspec(selectany) std::vector<std::wstring> g_sids =
```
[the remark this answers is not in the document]
joseartrivera: Some of the function require these to be convertible to wstring or convertible to Platform::String, can this be done with wstring_view?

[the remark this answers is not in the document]
danbelcher-MSFT: I would update those functions to `wstring_view` as appropriate.  If you need a `wstring`, you can construct one from a `wstring_view`:
```C++
wstring foo(const wstring_view& sv) {
    return wstring{ sv };
}
```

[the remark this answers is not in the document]
joseartrivera: I made the change, I didn't update GetCEngineString to accept a wstring_view since it would need to call c_str() anyway to pass a Platform::String to the ResourceLoader

src/CalcManager/CEngine/scicomm.cpp
joseartrivera: Yes it is sufficient for the purpose we are using it for degree strings act as the default, we will only look for rad/grad versions if hasAngleStrings is true.

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -394,7 +394,7 @@ String^ UnitConverterViewModel::ConvertToLocalizedString(const std::wstring& str
 
 void UnitConverterViewModel::DisplayPasteError()
 {
-    String^ errorMsg = AppResourceProvider::GetInstance().GetCEngineString(SIDS_DOMAIN); /*SIDS_DOMAIN is for "invalid input"*/
+    String^ errorMsg = AppResourceProvider::GetInstance().GetCEngineString(ref new String (SIDS_DOMAIN)); /*SIDS_DOMAIN is for "invalid input"*/
```
danbelcher-MSFT:
```suggestion
    String^ errorMsg = AppResourceProvider::GetInstance().GetCEngineString(StringReference(SIDS_DOMAIN)); /*SIDS_DOMAIN is for "invalid input"*/
```

PR #458, opened 2019-04-15
Add Pan/Zoom support for the graph surface

danbelcher-MSFT:
Add Pan/Zoom support for the graph surface.
- Currently only supports Mouse/Pen/Touch interactions.  Keyboard support will be added separately.

Manual

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -80,6 +80,20 @@ namespace GraphControl
         }
     }
 
+    void Grapher::MoveRangeByRatio(double ratioX, double ratioY)
+    {
+        if (m_graph)
+        {
+            if (auto renderer = m_graph->GetRenderer())
+            {
+                if (SUCCEEDED(renderer->MoveRangeByRatio(ratioX, ratioY)))
+                {
+                    m_renderMain->RunRenderPass();
+                }
+            }
+        }
+    }
+
     void Grapher::OnApplyTemplate()
     {
         auto swapChainPanel = dynamic_cast<SwapChainPanel^>(GetTemplateChild(StringReference(s_templateKey_SwapChainPanel)));
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -66,6 +66,20 @@ namespace GraphControl
         }
     }
 
+    void Grapher::ScaleRange(double centerX, double centerY, double scale)
+    {
+        if (m_graph)
+        {
+            if (auto renderer = m_graph->GetRenderer())
+            {
+                if (SUCCEEDED(renderer->ScaleRange(centerX, centerY, scale)))
+                {
+                    m_renderMain->RunRenderPass();
+                }
+            }
+        }
+    }
+
     void Grapher::OnApplyTemplate()
     {
         auto swapChainPanel = dynamic_cast<SwapChainPanel^>(GetTemplateChild(StringReference(s_templateKey_SwapChainPanel)));
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -417,4 +417,32 @@ namespace GraphControl
             e->Handled = true;
         }
     }
+
+    void Grapher::OnPointerWheelChanged(PointerRoutedEventArgs^ args)
+    {
+        PointerPoint^ currentPointer = args->GetCurrentPoint(/*relative to*/ this);
+
+        double delta = currentPointer->Properties->MouseWheelDelta;
+
+        // The maximum delta is 120 according to:
+        // https://docs.microsoft.com/en-us/uwp/api/windows.ui.input.pointerpointproperties.mousewheeldelta#Windows_UI_Input_PointerPointProperties_MouseWheelDelta
+        // Apply a dampening effect so that small mouse movements have a smoother zoom.
+        constexpr double scrollDamper = 0.15;
+        double scale = 1.0 + (abs(delta) / WHEEL_DELTA) * scrollDamper;
+
+        // positive delta if wheel scrolled away from the user
+        if (delta >= 0)
+        {
+            scale = 1.0 / scale;
+        }
+
+        // For scaling, the graphing engine interprets x,y position between the range [-1, 1].
+        // Translate the pointer position to the [-1, 1] bounds.
+        double centerX = (currentPointer->Position.X - ActualWidth / 2) / (ActualWidth / 2);
+        double centerY = (ActualHeight / 2 - currentPointer->Position.Y) / (ActualHeight / 2);
+
+        ScaleRange(centerX, centerY, scale);
+
+        args->Handled = true;
+    }
 }
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -7,8 +7,10 @@ using namespace GraphControl::DX;
 using namespace Platform;
 using namespace Platform::Collections;
 using namespace std;
+using namespace Windows::Devices::Input;
 using namespace Windows::Foundation;
 using namespace Windows::Foundation::Collections;
+using namespace Windows::System;
 using namespace Windows::UI;
 using namespace Windows::UI::Input;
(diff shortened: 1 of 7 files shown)


src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
...
+                    const auto& pos = args->Position;
+                    const auto[centerX, centerY] = PointerPositionToGraphPosition(pos.X, pos.Y, width, height);
+
+                    if (FAILED(renderer->ScaleRange(centerX, centerY, scale)))
+                    {
+                        return;
+                    }
+                }
+
+                m_renderMain->RunRenderPass();
```
joseartrivera: Is it possible for neither PointerPositionToGraphPosition and MoveRangeByRatio to not be called? If so do we still need to call RunRenderPass?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
...
+    void Grapher::OnPointerPressed(PointerRoutedEventArgs^ e)
+	{
+        // Set the pointer capture to the element being interacted with so that only it
+        // will fire pointer-related events
+        CapturePointer(e->Pointer);
+	}
+
+    void Grapher::OnPointerReleased(PointerRoutedEventArgs^ e)
+	{
+        // Release the pointer
```
joseartrivera: Don't need this comment imo

src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
...
+    void Grapher::OnPointerWheelChanged(PointerRoutedEventArgs^ args)
+    {
+        PointerPoint^ currentPointer = args->GetCurrentPoint(/*relative to*/ this);
+
+        double delta = currentPointer->Properties->MouseWheelDelta;
+
+        // The maximum delta is 120 according to:
+        // https://docs.microsoft.com/en-us/uwp/api/windows.ui.input.pointerpointproperties.mousewheeldelta#Windows_UI_Input_PointerPointProperties_MouseWheelDelta
+        // Apply a dampening effect so that small mouse movements have a smoother zoom.
+        constexpr double scrollDamper = 0.15;
```
joseartrivera: Curious as to when constexpr makes sense within functions instead of above where the other constexpr are defined

src/GraphControl/Control/Grapher.cpp
```diff
@@ -66,6 +83,34 @@ namespace GraphControl
...
 
+    void Grapher::ScaleRange(double centerX, double centerY, double scale)
+    {
+        if (m_graph)
+        {
+            if (auto renderer = m_graph->GetRenderer())
+            {
+                if (SUCCEEDED(renderer->ScaleRange(centerX, centerY, scale)))
+                {
+                    m_renderMain->RunRenderPass();
```
joseartrivera: In other places you are checking if m_renderMain is null, should this happen here too?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
...
+        {
+            if (auto renderer = m_graph->GetRenderer())
+            {
+                const double width = ActualWidth;
+                const double height = ActualHeight;
+
+                const auto& translation = args->Delta.Translation;
+                double translationX = translation.X;
+                double translationY = translation.Y;
+                if (translationX != 0 || translation.Y != 0)
```
joseartrivera: Use translationY here instead

src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
...
+                    // The graphing engine interprets scale amounts as the inverse of the value retrieved
+                    // from the ManipulationUpdatedEventArgs. Invert the scale amount for the engine.
+                    scale = 1.0 / scale;
+
+                    // The graphing engine interprets x,y position between the range [-1, 1].
+                    // Translate the pointer position to the [-1, 1] bounds.
+                    const auto& pos = args->Position;
+                    const auto[centerX, centerY] = PointerPositionToGraphPosition(pos.X, pos.Y, width, height);
+
+                    if (FAILED(renderer->ScaleRange(centerX, centerY, scale)))
```
joseartrivera: Why not use the ScaleRange function you defined earlier?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
...
+                double translationY = translation.Y;
+                if (translationX != 0 || translation.Y != 0)
+                {
+                    // The graphing engine pans the graph according to a ratio for x and y.
+                    // A value of +1 means move a half screen in the positive direction for the given axis.
+                    // Convert the manipulation's translation values to ratios for the engine.
+                    translationX /= -width;
+                    translationY /= height;
+
+                    if (FAILED(renderer->MoveRangeByRatio(translationX, translationY)))
```
joseartrivera: Same comment here, why not use the function that handles this from earlier?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -389,4 +434,98 @@ namespace GraphControl
             e->Handled = true;
         }
     }
+
+    void Grapher::OnPointerWheelChanged(PointerRoutedEventArgs^ args)
```
joseartrivera: Be consistent and also call this variable 'e' like the other functions

src/GraphControl/Control/Grapher.cpp
```diff
@@ -7,32 +7,42 @@ using namespace GraphControl::DX;
...
     constexpr auto s_propertyName_EquationsSource = L"EquationsSource";
-
-    DependencyProperty^ Grapher::s_forceProportionalAxesTemplateProperty;
     constexpr auto s_propertyName_ForceProportionalAxes = L"ForceProportionalAxes";
 
+    // Helper function for converting a pointer position to a position that the graphing engine will understand.
+    // posX/posY are the pointer position elements and width,height are the dimensions of the graph surface.
+    __inline pair<double, double> PointerPositionToGraphPosition(double posX, double posY, double width, double height)
+    {
+        return make_pair(2 * posX / width - 1, 1 - 2 * posY / height);
```
joseartrivera: Add parenthesis here for readability

src/GraphControl/Control/Grapher.cpp
danbelcher-MSFT: Yes, constexpr makes just as much sense here.  The compiler will still perform [constant initialization](https://en.cppreference.com/w/cpp/language/constant_initialization) of the variable which, in this case, means placing a definition in the program's .rdata section.  We should also prefer having the constant here instead of at the top, because of http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-introduce

src/GraphControl/Control/Grapher.cpp
danbelcher-MSFT: The reason is that those public-facing functions call a render pass before they return.  If I used them here, that would mean a single manipulation event would result in two render passes which is suboptimal.  Currently there aren't any clients of the public-facing functions so I can just remove them.  I used them when I was prototyping and playing around with the rendering APIs.

src/GraphControl/Control/Grapher.cpp
danbelcher-MSFT: The reason is that those public-facing functions call a render pass before they return.  If I used them here, that would mean a single manipulation event would result in two render passes which is suboptimal.  Currently there aren't any clients of the public-facing functions so I can just remove them.  I used them when I was prototyping and playing around with the rendering APIs.

src/GraphControl/Control/Grapher.cpp
danbelcher-MSFT: I'm not sure this is possible because _something_ happened to generate a ManipulationUpdated event.  I added the optimization just in case.

src/GraphControl/Control/Grapher.cpp
danbelcher-MSFT: Answered below

PR #464, opened 2019-04-17
Fix _In_ attributes in CalculatorHistory

tniessen:

These parameters have `_In_` attributes in the implementation, but not in the header file, causing a warning in the static code analysis.

This change causes the warnings to disappear.


janisozaur:
This is already covered by https://github.com/Microsoft/calculator/pull/400, in particular https://github.com/Microsoft/calculator/pull/400/files#r270185887

HowardWolosky:
Thanks @janisozaur -- As you've already mentioned, this is being handled in #400, so we'll let it get resolved as part of that PR.

PR #465, opened 2019-04-18
Fix test code that never executes

### LINKED ISSUE #175: Test code never executes - ill defined loop
[`NumbersAndOperatorsEnum::Add` is never `<= NumbersAndOperatorsEnum::None`](https://github.com/Microsoft/calculator/blob/057401f5f2b4bb1ea143da02c773ac18d1bb9a2e/src/CalculatorUnitTests/UnitConverterViewModelUnitTests.cpp#L500), [so the test never executes](https://docs.microsoft.com/en-us/visualstudio/code-quality/c6294).

**Describe the bug**
Someone wanted to write some unit tests, and idk, maybe they thought that `CM::Command::CommandNULL` was `>=NumbersAndOperatorsEnum::Add` because it was [lower on the screen](https://github.com/Microsoft/calculator/blob/057401f5f2b4bb1ea143da02c773ac18d1bb9a2e/src/CalcViewModel/Common/CalculatorButtonUser.h#L63)... but, alas, it's not:
enum_add
enum_none

**Steps To Reproduce**

I just enabled code analysis. C'mon, Microsoft, it's your own tool!

EriWong:
Test code never executes - ill defined loop

## Fixes #175.
Fixes loop in test code to verify that commands not supported by the unit converter viewmodel result in no-op.

Replaced loop of range of enums with a list of all currently supported commands in check that those commands are no-ops.

Ran modified test to ensure it passes


src/CalculatorUnitTests/UnitConverterViewModelUnitTests.cpp
```diff
@@ -454,57 +454,189 @@ namespace CalculatorUnitTests
...
+                NumbersAndOperatorsEnum::BINPOS59,
+                NumbersAndOperatorsEnum::BINPOS60,
+                NumbersAndOperatorsEnum::BINPOS61,
+                NumbersAndOperatorsEnum::BINPOS62,
+                NumbersAndOperatorsEnum::BINPOS63,
+                NumbersAndOperatorsEnum::BINEND,
+                NumbersAndOperatorsEnum::Hyp
+            };
+
+            for (const auto& button : buttons)
```
EriWong: So I'm not entirely convinced it's useful to make sure that we don't actually do anything when any of the above commands are sent.  I'm ok with just removing the loop entirely if this test doesn't make sense and just keep the part of this test that covers supported commands.

HowardWolosky: Based on the observation in the original bug (#175) and the position of `Zero` and `None` in [`CalculatorButtonUser.h`](https://github.com/Microsoft/calculator/blob/057401f5f2b4bb1ea143da02c773ac18d1bb9a2e/src/CalcViewModel/Common/CalculatorButtonUser.h#L63), I don't believe that the _intention_ by the original author at the time was to test _all_ of these commands, but rather just the ones between `Add` and `None`.  I did check the archive, and it's been this way since the test was first introduced (at that time, `NumbersAndOperatorsEnum` ended after `Cube`, so there _were_ commands after `None`).

I think it _can_ be reasonable to have tests that validate that certain commands do nothing in certain modes, however I think they only make sense if it would be clear to the maintainers what the failing test would indicate.  So, if we had this test currently pass because `Add` does nothing, but then in some mythical future we added the ability to add measurements together which would cause this test to fail, what would our action item be?  We'd probably just remove `Add` from `buttons` and give it its own unique case.  So, did it add value?  Not substantially.  Would there be value for still running this loop, but only for the subset of buttons between `Add` and `None` vs all of them?  Probably not.

I'd say go ahead and remove `buttons` and the loop entirely.

EriWong: Cool.  I'll go ahead and remove the loop.

PR #467, opened 2019-04-19
Collapse CurrencySecondaryStatus when no text is present in the text block

### LINKED ISSUE #313: In Scan/Item mode, Narrator focus navigates to hidden element “No next item” after “Update rates” link in "Currency Converter" window
**Repro Steps**
* Launch narrator.
* Launch Calculator app.
* Navigate to "Open navigation" button and activate it.
* Navigate to "Currency Converter" list item and activate it.
* Currency Converter window will open.
* Further navigate in "Currency Converter" window in scan mode.
* Verify that narrator focus navigates to hidden element “No next item” after “Update rates” link.

**Actual Result**
In Scan/Item mode, Narrator focus navigates to hidden element “No next item” after “Update rates” link.

**Expected Result**
In Scan/Item mode, Narrator focus should not navigate to hidden element “No next item” after “Update rates” link.

App Version: 10.1902.16.0

EriWong:
## Fixes #313 
In Scan/Item mode, Narrator focus navigates to hidden element “No next item” after “Update rates” link in "Currency Converter" window #313

- Adds an x:Name to the CurrencySecondaryStatus text block
- Adds a NormalCurrencyStatus visual state to the CurrencySecondaryStatusStates
- Adds functionality to the CurrencySecondaryStatusStates to show or hide the CurrencySecondaryStatus text block.

- Verified that the textblock is not visible in the accessibility tree via inspect.exe from the windows sdk.
- Verified that Narrator also does not stop on the block when in scan mode.
- Verified that the textblock is visible in the accessibility tree and read out in Narrator when the ChargesMayApplyCurrencyStatus or FailedCurrencyStatus viewstates are set.

### DIFF
--- a/src/Calculator/Views/UnitConverter.xaml
+++ b/src/Calculator/Views/UnitConverter.xaml
@@ -428,14 +428,21 @@
                 </VisualState>
             </VisualStateGroup>
             <VisualStateGroup x:Name="CurrencySecondaryStatusStates">
+                <VisualState x:Name="NormalCurrencyStatus">
+                    <VisualState.Setters>
+                        <Setter Target="CurrencySecondaryStatusBlock.Visibility" Value="Collapsed"/>
+                    </VisualState.Setters>
+                </VisualState>
                 <VisualState x:Name="ChargesMayApplyCurrencyStatus">
                     <VisualState.Setters>
                         <Setter Target="CurrencySecondaryStatus.Foreground" Value="{ThemeResource AppControlPageTextBaseHighColorBrush}"/>
+                        <Setter Target="CurrencySecondaryStatusBlock.Visibility" Value="Visible"/>
                     </VisualState.Setters>
                 </VisualState>
                 <VisualState x:Name="FailedCurrencyStatus">
                     <VisualState.Setters>
                         <Setter Target="CurrencySecondaryStatus.Foreground" Value="{ThemeResource AppControlPageTextRedColorBrush}"/>
+                        <Setter Target="CurrencySecondaryStatusBlock.Visibility" Value="Visible"/>
                     </VisualState.Setters>
                 </VisualState>
             </VisualStateGroup>
@@ -615,7 +622,8 @@
                                          x:Uid="RefreshButtonText"
                                          Foreground="{ThemeResource SystemControlHyperlinkBaseHighBrush}"
                                          Click="CurrencyRefreshButton_Click"/>
-                        <TextBlock Margin="0,7,0,0" Style="{ThemeResource CaptionTextBlockStyle}">
+                        <TextBlock x:Name="CurrencySecondaryStatusBlock" Margin="0,7,0,0"
+                                   Style="{ThemeResource CaptionTextBlockStyle}">
                             <Run x:Name="Spacing" Text="&#x200A;"/>
                             <Run x:Name="CurrencySecondaryStatus"
                                  FontWeight="SemiBold"
--- a/src/Calculator/Views/UnitConverter.xaml
+++ b/src/Calculator/Views/UnitConverter.xaml
@@ -428,11 +428,7 @@
                 </VisualState>
             </VisualStateGroup>
             <VisualStateGroup x:Name="CurrencySecondaryStatusStates">
-                <VisualState x:Name="NormalCurrencyStatus">
-                    <VisualState.Setters>
-                        <Setter Target="CurrencySecondaryStatusTextBlock.Visibility" Value="Collapsed"/>
-                    </VisualState.Setters>
-                </VisualState>
+                <VisualState x:Name="NormalCurrencyStatus"/>
                 <VisualState x:Name="ChargesMayApplyCurrencyStatus">
                     <VisualState.Setters>
                         <Setter Target="CurrencySecondaryStatus.Foreground" Value="{ThemeResource AppControlPageTextBaseHighColorBrush}"/>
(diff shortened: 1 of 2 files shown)


src/Calculator/Views/UnitConverter.xaml
```diff
@@ -615,7 +622,8 @@
                                          x:Uid="RefreshButtonText"
                                          Foreground="{ThemeResource SystemControlHyperlinkBaseHighBrush}"
                                          Click="CurrencyRefreshButton_Click"/>
-                        <TextBlock Margin="0,7,0,0" Style="{ThemeResource CaptionTextBlockStyle}">
+                        <TextBlock x:Name="CurrencySecondaryStatusTextBlock" Margin="0,7,0,0"
```
danbelcher-MSFT: Install the XAML Styler extension so you get automatic formatting of XAML files.  See [README.md](/README.md).

src/Calculator/Views/UnitConverter.xaml
```diff
@@ -428,14 +428,21 @@
                 </VisualState>
             </VisualStateGroup>
             <VisualStateGroup x:Name="CurrencySecondaryStatusStates">
+                <VisualState x:Name="NormalCurrencyStatus">
+                    <VisualState.Setters>
+                        <Setter Target="CurrencySecondaryStatusTextBlock.Visibility" Value="Collapsed"/>
```
danbelcher-MSFT: For `Normal` states, we typically set the appropriate properties directly on the element and define an empty `<VisualState x:Name="NormalCurrencyStatus"/>`

src/Calculator/Views/UnitConverter.xaml.cpp
```diff
@@ -152,18 +152,18 @@ void UnitConverter::OnOfflineNetworkAccess()
...
 void UnitConverter::SetNormalCurrencyStatus()
 {
-    CurrencySecondaryStatus->Text = L"";
+    VisualStateManager::GoToState(this, NormalCurrencyStatus->Name, false);
 }
 
 void UnitConverter::SetChargesMayApplyStatus()
 {
-    VisualStateManager::GoToState(this, L"ChargesMayApplyCurrencyStatus", false);
+    VisualStateManager::GoToState(this, ChargesMayApplyCurrencyStatus->Name, false);
```
danbelcher-MSFT: nitpicky, but probably a better pattern to set the text before changing the visual state, ie, do the right work before displaying.

danbelcher-MSFT: Thanks for cleaning up my mess 👍 😄

src/Calculator/Views/UnitConverter.xaml
EriWong: Super useful.  Thanks

src/Calculator/Views/UnitConverter.xaml
EriWong: Makes sense.

src/Calculator/Views/UnitConverter.xaml.cpp
EriWong: Agreed.

src/Calculator/Views/UnitConverter.xaml.cpp
```diff
@@ -152,19 +152,19 @@ void UnitConverter::OnOfflineNetworkAccess()
 
 void UnitConverter::SetNormalCurrencyStatus()
 {
-    CurrencySecondaryStatus->Text = L"";
+    VisualStateManager::GoToState(this, NormalCurrencyStatus->Name, false);
```
rudyhuyn: We should still do `CurrencySecondaryStatus->Text = L"";` even if the control will be collapsed

src/Calculator/Views/UnitConverter.xaml
```diff
@@ -615,7 +618,10 @@
                                          x:Uid="RefreshButtonText"
                                          Foreground="{ThemeResource SystemControlHyperlinkBaseHighBrush}"
                                          Click="CurrencyRefreshButton_Click"/>
-                        <TextBlock Margin="0,7,0,0" Style="{ThemeResource CaptionTextBlockStyle}">
+                        <TextBlock x:Name="CurrencySecondaryStatusTextBlock"
+                                   Margin="0,7,0,0"
```
rudyhuyn: Because the control will be collapsed, we are losing the margin (7 pixels vertically) between the refresh button and the bottom of the window (or the numpad)

rudyhuyn: (this margin should be removed and `Margin="0,0,0,7"` should be added to CurrencyRefreshBlockControl)

rudyhuyn: My bad, it was a horizontal stackpanel!

rudyhuyn:
Alternative solution to solve the root cause of the issue (the narrator thinking that the textblock isn't empty because of the half space character):

Replace:

```
                        <TextBlock Margin="0,7,0,0" Style="{ThemeResource CaptionTextBlockStyle}">
                            <Run x:Name="Spacing" Text="&#x200A;"/>
                            <Run x:Name="CurrencySecondaryStatus"
                                 FontWeight="SemiBold"
                                 Text=""/>
                        </TextBlock>
```
by

```
                        <TextBlock x:Name="CurrencySecondaryStatus"
                                   Margin="3,7,0,0"
                                   Style="{ThemeResource CaptionTextBlockStyle}"
                                   FontWeight="SemiBold"/>
``` 

(`Margin="3,8,0,0"` would be even better to correctly align the button and the textblock)

src/Calculator/Views/UnitConverter.xaml.cpp
EriWong: I'm happy to make this change, but for my own learning, I'd like to know the reasoning behind it.  If the textblock is not visible, what benefit is there to changing it's text?

EriWong: Went ahead with your other suggestion as that seems like a better fix.

PR #473, opened 2019-04-23
[JaEra] Calc: subtracting 1-year from date during Reiwa 1 yields unexpected results.

EriWong:
## Fixes 471#.
[JaEra] Calc: subtracting 1-year from date during Reiwa 1 yields unexpected results.

- When doing addition or subtraction of dates, if using the Japanese Era Calendar system, use the gregorian system to perform year addition/subtraction.

- Manually verified that the transition between Reiwa and Heisei eras are handled correctly
- Wrote tests covering an already established era transition between Heisei and Showa eras.

### DIFF
--- a/src/CalcViewModel/Common/DateCalculator.cpp
+++ b/src/CalcViewModel/Common/DateCalculator.cpp
@@ -20,13 +20,27 @@ DateCalculationEngine::DateCalculationEngine(_In_ String^ calendarIdentifier)
 // Returns: True if function succeeds to calculate the date else returns False
 bool DateCalculationEngine::AddDuration(_In_ DateTime startDate, _In_ const DateDifference& duration, _Out_ DateTime *endDate)
 {
+    auto currentCalendarSystem = m_calendar->GetCalendarSystem();
+
     try
     {
         m_calendar->SetDateTime(startDate);
 
         if (duration.year != 0)
         {
+            // The Japanese Era system can have multiple year partitions within the same year.
+            // For example, April 30, 2019 is denoted April 30, Heisei 31; May 1, 2019 is denoted as May 1, Reiwa 1.
+            // The Calendar treats Heisei 31 and Reiwa 1 as separate years, which results in some unexpected behaviors where subtracting a year from Reiwa 1 results in a date in Heisei 31.
+            // To provide the expected result across era boundaries, we first convert the Japanese era system to a Gregorian system, do date math, and then convert back to the Japanese era system.
+            // This works because the Japanese era system maintains the same year/month boundaries and durations as the Gregorian system and is only different in display value.
+            if (currentCalendarSystem == CalendarIdentifiers::Japanese)
+            {
+                m_calendar->ChangeCalendarSystem(CalendarIdentifiers::Gregorian);
+            }
+
             m_calendar->AddYears(duration.year);
+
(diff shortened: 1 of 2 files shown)


src/CalcViewModel/Common/DateCalculator.cpp
```diff
@@ -20,13 +20,27 @@ DateCalculationEngine::DateCalculationEngine(_In_ String^ calendarIdentifier)
...
         m_calendar->SetDateTime(startDate);
 
         if (duration.year != 0)
         {
+            // The Japanese Era system can have multiple year partitions within the same year.
+            // For example, April 30, 2019 is denoted April 30, Heisei 31; May 1, 2019 is denoted as May 1, Reiwa 1.
+            // The Calendar treats Heisei 31 and Reiwa 1 as separate years, which results in some unexpected behaviors where subtracting a year from Reiwa 1 results in a date in Heisei 31.
+            // To provide the expected result across era boundaries, we first convert the Japanese era system to a Gregorian system, do date math, and then convert back to the Japanese era system.
+            // This works because the Japanese era system maintains the same year/month boundaries and durations as the Gregorian system and is only different in display value.
+            if (currentCalendarSystem == CalendarIdentifiers::Japanese)
```
danbelcher-MSFT: Does this special logic need to be specific to adding/subtracing years?  Why not months or days?

src/CalcViewModel/Common/DateCalculator.cpp
```diff
@@ -20,13 +20,27 @@ DateCalculationEngine::DateCalculationEngine(_In_ String^ calendarIdentifier)
...
+            {
+                m_calendar->ChangeCalendarSystem(CalendarIdentifiers::Gregorian);
+            }
+
             m_calendar->AddYears(duration.year);
+
+            m_calendar->ChangeCalendarSystem(currentCalendarSystem);
```
danbelcher-MSFT: Could we use something like [`scope_exit`](https://github.com/Microsoft/wil/blob/5a6e8df54674299929b80735e61e0923394404b1/include/wil/resource.h#L517-L520) so we don't have to duplicate logic in the golden path and if an exception is thrown?

src/CalcViewModel/Common/DateCalculator.cpp
EriWong: The issue is really only when we deal with years because Calendar treats Heisei 31 and Reiwa 1 as 2 different years.  When we add a month to a date in the last month of Heisei, it transitions correctly (ie: the same way we transition correctly when we add a month to a date in December on a Gregorian calendar)

src/CalcViewModel/Common/DateCalculator.cpp
EriWong: That's a good pattern to use if we see this pattern up in multiple places...  At the moment, I think it's out of scope for this particular fix.  I'm leaning towards not doing this right now.

EriWong: Actually, looking through the codebase, it looks like there are a number of places that would benefit from this change.  I'll take a look at including it.

EriWong: Looks like the other cases I was thinking about are being dealt with in issue #61. I'm gonna go back to saying we won't need this just yet.

PR #474, opened 2019-04-23
Allow vcxproj and sln files to merge as text.

danbelcher-MSFT:
This will cause Git to insert conflict markers in the file and make it easier to identify and fix merge conflicts. Previously, Git would complain about conflicts with these files, but no markers were inserted.

Manual.
- Checkout feature/GraphingCalculator and merge in upstream/master.  Git identifies merge conflicts present for Calculator.vcxproj but there are no conflict markers in the file.  Abort the merge.
- Merge in dabelc/MergeProjectFilesAsText.  Confirm Git is able to merge Calculator.vcxproj.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,29 +13,6 @@
 *.cs     diff=csharp
 
 ###############################################################################
-# Set the merge driver for project and solution files
-#
-# Merging from the command prompt will add diff markers to the files if there
-# are conflicts (Merging from VS is not affected by the settings below, in VS
-# the diff markers are never inserted). Diff markers may cause the following 
-# file extensions to fail to load in VS. An alternative would be to treat
-# these files as binary and thus will always conflict and require user
-# intervention with every merge. To do so, just uncomment the entries below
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -12,29 +12,6 @@
...
-# Set the merge driver for project and solution files
-#
-# Merging from the command prompt will add diff markers to the files if there
-# are conflicts (Merging from VS is not affected by the settings below, in VS
-# the diff markers are never inserted). Diff markers may cause the following 
-# file extensions to fail to load in VS. An alternative would be to treat
-# these files as binary and thus will always conflict and require user
-# intervention with every merge. To do so, just uncomment the entries below
-###############################################################################
-*.sln       merge=binary
```
danbelcher-MSFT: Instead of deleting this, we could also simply comment them out so that future readers will know that they have the option to return to binary merge driver.

mcooley: My preference is to just delete.

Issue #416, opened 2019-03-30
Temperature/Angle/Power conversion outputs change if there are more than three zeros after the input value, even though the value are mathematically equivalent


alexyorke:
**Describe the bug**
When appending zeros to a number when converting a temperature, the output suddenly becomes incorrect, even though the value is the same.

**Steps To Reproduce**
1. Open Calculator
2. Switch to temperature view
3. Switch the first temperature type to Celculus
4. Switch the second temperature type to Kelvin
5. Type in `273.00000000000`
6. The output will change from `0.15` to `0.149999999998` when adding additional zeros beyond three.

Note: Also happens with angle and power conversions.

**Expected behavior**
The temperature should remain at `0.15`, irregardless of how many zeros are in the input.

**Device and Application Information (please complete the following information):**
- OS Build: 10.0.17134.0
- Architecture: X64
- Application Version: 10.1812.10048.0

**Screenshots**
![image](https://user-images.githubusercontent.com/7844441/55282883-62a29780-532c-11e9-8215-42e1a71140ac.png)

![image](https://user-images.githubusercontent.com/7844441/55282904-d3e24a80-532c-11e9-9ae8-7a9df146c4c6.png)

**Additional information**

I had a look at IEEE 754 (https://en.wikipedia.org/wiki/Floating-point_arithmetic#IEEE_754:_floating_point_in_modern_computers), and it seems to cover situations where adding very small numbers together causes rounding errors, and not ones which end with just zeros.

> A team member has edited this comment to update repro steps for clarity.

rudyhuyn:
Nice found! I agree, it's a bug, the calculator should ignore the trailing zero. Probably a bug with `Rational`

grochocki:
Thanks for catching this!

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

Decagon:
Also happens with Angle and Power view, but not Speed view, strangely.

Issue #424, opened 2019-04-03
Application fails to take float value for Currency converter when Japanese Calendar is enabled


MicrosoftIssueBot:
A team member will update this issue with more details soon.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

Geveffe:
**Describe the bug**

When you update your calendar settings to Japanese, Calculator's currency converter stops letting you input values after a decimal point.

**Steps To Reproduce**

1. Open intl.cpl
2. Select "Japanese (Japan)" from the "Format" drop down
3. Select "Additional Settings" Date
4. In the "Calendar Type" dropdown, select the 2nd calendar "和暦", which is the Japanese
5. Select "Apply" and “Ok" buttons.
6. Launch Calculator
7. From menu, select currency convertor
8. Type in digits, then a decimal point, then type in more digits

**Expected behavior**

The additional digits should get used as part of the number, but don't.

**Device and Application Information (please complete the following information):**

OS Build: 10.0.18362.0
Architecture: X64
Application Version: 10.1902.42.0

Geveffe:
I looked into this one a bit more. The reason that this happens is that when you change your Format to Japanese (Japan) the Currency Format changes by default to have 0 "No. of digits after decimal". This may or may not be correct, but Calculator is correctly using the format. I will follow up with the intl.cpl team form them to decide if the current intl.cpl behavior is correct or not.

rudyhuyn:
This issue showcases a wider issue.

The Currency Format dialog gives information about the locale currency, not all currencies.

When you use Japanese yens, the lower coin is 1 yen, cents don't exist, so it seems normal that the number of digits after decimal equals 0 when the region is Japan. But should we use this settings for all other currencies? Probably not.

When we convert Yen in Dollar, should we display: `25 Yen == 0 Dollar` or `25 Yen == 0.22 Dollar`?
Because Yen doesn't feature cents, should we allow users to write "212.40 Yens" when this value doesn't exist?

Similarly, if a US user wants to know how many dollars his/her Kuwaiti dinars are worth, we should probably allow him/her to type `12.005 Dinar` (because the currency includes millesimal) 

We should probably associate a number of decimals to each currency and use these values instead of Windows settings for the currency converter.

rudyhuyn:
[...]

PR #498, opened 2019-05-10
Unit Converter - Calculate the rounding precision of results based on the source

### LINKED ISSUE #455: Converter should progressively increase the number of significant digits (instead of switching from 7 to 15 instantly)
**Describe the improvement**
Based on the length of the string associated to the entry value, the converter will switch from 7 to 15 significant digits when the entry value length is higher than 7.

**Steps To Reproduce**
- Open the Length converter
- Select `centimeters` and `inches`
- Type `1.22222` in the centimeter field
- add 1 extra `2`
The result will switch from `0.481189` to `0.48118976377953`

**Expected behavior**
- based on the length of the entry value, we should progressively increase the number of significant digits, but not an instant switch.

**Screenshots**

**Current behavior:**

**Proposal:**

 - OS Build: 10.0.18362.0
 - Architecture: X64
 - Application Version: 10.1903.21.0
 - Region: en-US
 - Dev Version Installed: True

**Additional context**

The diff will also fix https://github.com/Microsoft/calculator/issues/416 at the same time.

rudyhuyn:
- Ignore trailing zeros (issue #416), negative sign, decimal separator of the source when calculating the precision of the result 
- Round the result number based on the number of digits in the source value

Manually tested, including the 2 repros from #455 and #416

### DIFF
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -885,6 +885,8 @@ void UnitConverter::Calculate()
                 }
                 else
                 {
+					// Fewer digits are needed following the decimal if the number is large,
+					// we calculate the number of decimals necessary based on the number of digits in the integer part.
                     precision = max(0, max(OPTIMALDIGITSALLOWED, min(MAXIMUMDIGITSALLOWED, currentNumberSignificantDigits)) - numPreDecimal);
                 }
 
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -178,7 +178,7 @@ void UnitConverter::SwitchActive(const wstring& newValue)
     swap(m_currentHasDecimal, m_returnHasDecimal);
     m_returnDisplay = m_currentDisplay;
     m_currentDisplay = newValue;
-    m_currentHasDecimal = (m_currentDisplay.find(L'.') != m_currentDisplay.npos);
+    m_currentHasDecimal = (m_currentDisplay.find(L'.') != wstring::npos);
     m_switchedActive = true;
 
     if (m_currencyDataLoader != nullptr && m_vmCurrencyCallback != nullptr)
@@ -202,7 +202,7 @@ vector<wstring> UnitConverter::StringToVector(const wstring& w, const wchar_t* d
     size_t delimiterIndex = w.find(delimiter);
     size_t startIndex = 0;
     vector<wstring> serializedTokens = vector<wstring>();
-    while (delimiterIndex != w.npos)
+    while (delimiterIndex != wstring::npos)
     {
         serializedTokens.push_back(w.substr(startIndex, delimiterIndex - startIndex));
         startIndex = delimiterIndex + (int)wcslen(delimiter);
@@ -891,7 +891,7 @@ void UnitConverter::Calculate()
                 m_returnDisplay = RoundSignificant(returnValue, precision);
                 TrimString(m_returnDisplay);
             }
-            m_returnHasDecimal = (m_returnDisplay.find(L'.') != m_returnDisplay.npos);
+            m_returnHasDecimal = (m_returnDisplay.find(L'.') != wstring::npos);
         }
     }
     UpdateViewModel();
@@ -930,11 +930,11 @@ unsigned int UnitConverter::GetNumberSignificantDigits(std::wstring value)
 {
     TrimString(value);
     unsigned int numberSignificantDigits = static_cast<unsigned int>(value.size());
-    if (value.find(L'.') != value.npos)
+    if (value.find(L'.') != wstring::npos)
     {
         --numberSignificantDigits;
     }
-    if (value.find(L'-') != value.npos)
+    if (value.find(L'-') != wstring::npos)
     {
         --numberSignificantDigits;
     }
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -882,7 +882,7 @@ void UnitConverter::Calculate()
 
             auto currentDisplayTrimmed = m_currentDisplay;
             TrimString(currentDisplayTrimmed);
-            int currentNumberSignificantDigits = currentDisplayTrimmed.size();
+            int currentNumberSignificantDigits = static_cast<int>(currentDisplayTrimmed.size());
             if (currentDisplayTrimmed.find(L'.') != currentDisplayTrimmed.npos)
                 --currentNumberSignificantDigits;
             if (currentValue < 0)
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -848,7 +848,7 @@ void UnitConverter::Calculate()
     }
 
     unordered_map<Unit, ConversionData, UnitHash> conversionTable = m_ratioMap[m_fromType];
-	if (AnyUnitIsEmpty() || (conversionTable[m_toType].ratio == 1.0 && conversionTable[m_toType].offset == 0.0))
+    if (AnyUnitIsEmpty() || (conversionTable[m_toType].ratio == 1.0 && conversionTable[m_toType].offset == 0.0))
     {
         m_returnDisplay = m_currentDisplay;
         m_returnHasDecimal = m_currentHasDecimal;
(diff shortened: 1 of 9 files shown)


src/CalcManager/UnitConverter.cpp
```diff
@@ -879,16 +879,26 @@ void UnitConverter::Calculate()
         else
         {
             returnValue = stod(m_returnDisplay);
-            wstring returnString;
-            if (m_currentDisplay.size() <= OPTIMALDIGITSALLOWED && abs(returnValue) >= OPTIMALDECIMALALLOWED)
+
+            auto currentDisplayTrimmed = m_currentDisplay;
+            TrimString(currentDisplayTrimmed);
+            int currentNumberSignificantDigits = static_cast<int>(currentDisplayTrimmed.size());
+            if (currentDisplayTrimmed.find(L'.') != currentDisplayTrimmed.npos)
```
joseartrivera: Doesn't the '.' character change based on the display language or does that happen later?

src/CalcManager/UnitConverter.cpp
```diff
@@ -879,16 +879,26 @@ void UnitConverter::Calculate()
...
+                precision = MAXIMUMDIGITSALLOWED;
             }
             else
             {
-                returnString = RoundSignificant(returnValue, MAXIMUMDIGITSALLOWED - min(numPreDecimal, MAXIMUMDIGITSALLOWED));
+                precision = max(0, max(OPTIMALDIGITSALLOWED, min(MAXIMUMDIGITSALLOWED, currentNumberSignificantDigits)) - numPreDecimal);
             }
-            m_returnDisplay = returnString;
+       
+            m_returnDisplay = RoundSignificant(returnValue, precision);
```
joseartrivera: Is this the expected behavior in currency mode? I would think we would at least always want a precision of 2 in that mode? Left is after this change, right is before. 

image

rudyhuyn: You are right, for currency, I will force the display of X decimals based on the system locale settings (while waiting for the Calculator to include the number of digit to use per currency, see https://github.com/microsoft/calculator/issues/424#issuecomment-479823338)

src/CalcManager/UnitConverter.cpp
rudyhuyn: This is safe, this string is locale-independent

src/CalcManager/UnitConverter.cpp
```diff
@@ -900,9 +901,9 @@ void UnitConverter::Calculate()
 /// Trims out any trailing zeros or decimals in the given input string
 /// </summary>
 /// <param name="input">wstring to trim</param>
-void UnitConverter::TrimString(wstring& returnString)
+void UnitConverter::TrimString(_Inout_ wstring& returnString)
 {
-    if (returnString.find(L'.') == m_returnDisplay.npos)
```
rudyhuyn: Bug fixed!

danbelcher-MSFT: What was the bug?  [`npos`](https://en.cppreference.com/w/cpp/string/basic_string/npos) is a static member of `basic_string`.  My understanding is that the condition should be `returnString.find(L'.') == wstring::npos`.  You can write `m_returnDisplay.npos` or `returnString.npos`, but I believe it should be the same either way because they are referring to the same static member.  We should still prefer `wstring::npos` because it makes it clear that the member is static instead of the current form which implies non-static.

src/CalcManager/UnitConverter.cpp
```diff
@@ -922,6 +923,24 @@ void UnitConverter::TrimString(wstring& returnString)
...
 }
 
+/// <summary>
+/// Get number of significant digits (integer part + fractional part) of a number</summary>
+/// <param name="value">the number</param>
+unsigned int UnitConverter::GetNumberSignificantDigits(std::wstring value)
+{
+    TrimString(value);
+    unsigned int numberSignificantDigits = static_cast<unsigned int>(value.size());
+    if (value.find(L'.') != value.npos)
```
danbelcher-MSFT: Same comment here and below.  My understanding is this should be `wstring::npos`

src/CalcManager/UnitConverter.h
```diff
@@ -292,6 +290,10 @@ namespace UnitConversionManager
         std::shared_ptr<IConverterDataLoader> GetDataLoaderForCategory(const Category& category);
         std::shared_ptr<ICurrencyConverterDataLoader> GetCurrencyConverterDataLoader();
 
+        static void TrimString(_Inout_ std::wstring& input);
```
danbelcher-MSFT: Are these functions specific to UnitConverter?  Would it make more sense to put them as free functions in Utils?

src/CalcManager/UnitConverter.cpp
```diff
@@ -922,6 +923,24 @@ void UnitConverter::TrimString(wstring& returnString)
     }
 }
 
+/// <summary>
+/// Get number of significant digits (integer part + fractional part) of a number</summary>
+/// <param name="value">the number</param>
+unsigned int UnitConverter::GetNumberSignificantDigits(std::wstring value)
+{
+    TrimString(value);
+    unsigned int numberSignificantDigits = static_cast<unsigned int>(value.size());
```
danbelcher-MSFT: What's wrong with `size_t`?  :)

rudyhuyn: The function returns a unsigned int, so a cast is necessary.

rudyhuyn: The function returns an unsigned int, so a cast is necessary.

danbelcher-MSFT: I was mostly poking for fun :)  Although, if the return value is derived from a type with a higher precision there's no imperative to convert to a lower precision type.  It would indicate to me that the return type should also be `size_t`.  That said, I'm not sure how many clients there are of this function and maybe you didn't want to go update.

src/CalcManager/UnitConverter.h
rudyhuyn: Not specific but only used by UnitConverter.

There is no way to make a free-standing function private, so I would prefer to not do it, especially with this library which has a high chance of being used by other projects. 

- free-standing in global namespace -> can create collision when another app uses this library
- free-standing in namespace -> still visible
- utility class -> better to avoid.

There is probably some functions (including from other classes) we can group to create a new class that will make more sense (`NumberStringOperation`?), but it can be part of another PR.

For this diff, I hesitated to remove these functions from the header file and only keep the implementation, do you have a guidance/best practice for that?

src/CalcManager/UnitConverter.cpp
rudyhuyn: The bug wasn't technical but logical, it was weird to use an instance we don't even use in this function to get access to npos.

All methods in this class get access to `wstring::npos` via instances, I was intrigued but I thought it was made on purpose for a reason I ignored. But it looks like it's the only class doing that, so I'm totally fine with using wstring:npos instead!

src/CalcManager/UnitConverter.h
danbelcher-MSFT: I don't see a problem with having the functions be visible to other projects.  We _want_ them to be visible so that they can be reused.  If the functions don't have anything specific to UnitConverter, then it doesn't make sense to tie them to the UnitConverter class.  If they were free functions we would still do our due diligence and put them in an appropriate namespace.  Like many of my comments, I'm not necessarily asking for a change here, but I find the discussion valuable.  In this case, I don't see the concern with string manipulation functions being free and grouped under an appropriate namespace.

Regarding your last comment, you wanted to remove the functions from the header file and only keep the implementation..  did you mean remove them as static members of UnitConverter and keep the definitions in the source file?  I don't have a problem with that, although I would have the same question of whether it was more appropriate to move them to a common namespace/class.  If you were to make that change, please place the definitions in an anonymous namespace.  One reason to keep the functions as private static members would be so that they can be unit tested.

rudyhuyn: > I don't see a problem with having the functions be visible to other projects.

In this case, we are fine, it was my main concern. I will update the PR.

For the namespace (unfortunately CalcManager does not have a clean namespace structure) what do you think about `CalculationManager::StringUtils`

danbelcher-MSFT: Thanks, @rudyhuyn!  `CalcManager` would also be fine.  It matches the project name and we abbreviate elsewhere  as with `CalcViewModel`.  Either way, adding namespaces for the CalcManager project is much appreciated.

grochocki:
ping: @joseartrivera

src/CalcManager/UnitConverter.cpp
```diff
@@ -842,100 +844,59 @@ void UnitConverter::Calculate()
...
-                returnString = RoundSignificant(returnValue, MAXIMUMDIGITSALLOWED - min(numPreDecimal, MAXIMUMDIGITSALLOWED));
+                int currentNumberSignificantDigits = GetNumberDigits(m_currentDisplay);
+                int precision = 0;
+                if (abs(returnValue) < OPTIMALDECIMALALLOWED)
+                {
+                    precision = MAXIMUMDIGITSALLOWED;
+                }
+                else
+                {
+                    precision = max(0, max(OPTIMALDIGITSALLOWED, min(MAXIMUMDIGITSALLOWED, currentNumberSignificantDigits)) - numPreDecimal);
```
joseartrivera: I think this makes sense, can we add a small comment giving an explanation for the logic here?

rudyhuyn:
> Hi Rudy, sorry for the delay on my response. Just a few small requests before merging this in. Thanks again for taking on this work! :)

all done!

PR #501, opened 2019-05-10
Added Calculator Standard Mode UI Tests

### LINKED ISSUE #74: Long term plan for testing needs
- Identify test gaps in calculator in existing test coverage.
- Updating manual tests to provide better E2E coverage.
- Create new automated tests to fill identified gaps.

sanderl:
- Added the CalculatorUIFramework to handle the WinAppDriver logic.
- Added Standard Mode smoke tests and BVTs to the CalculatorUITests project.
- Removed old UI tests that did not use the CalculatorUIFramework

- Used VSTest to run the UI tests and unit tests


src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CalculatorUITestFramework
+{
+    public class HistoryPanel : ApplicationBase
```
mcooley: Shouldn't inherit from ApplicationBase here. (Prefer composition over inheritance.)

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -0,0 +1,51 @@
...
+        public void OpenHistoryPanel()
+        {
+            HistoryLabel.Click();
+        }
+
+        /// <summary>
+        /// Gets all of the history items listed in the History Pane.
+        /// </summary>
+        /// <returns>A readonly collection of history items.</returns>
+        public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
```
mcooley: Are the returned elements actually of type WindowsElement?

src/CalculatorUITestFramework/NumerPad.cs
```diff
@@ -0,0 +1,77 @@
+using System;
+using OpenQA.Selenium.Appium.Windows;
+
+namespace CalculatorUITestFramework
+{
+    public class NumerPad : ApplicationBase
```
mcooley: nit: spelling "NumberPad"

src/CalculatorUITestFramework/HistoryPanel.cs
mcooley: Same feedback for the rest of the classes.

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -0,0 +1,51 @@
...
+        public WindowsElement HistoryLabel => TryFindElementByAccessibilityId("HistoryLabel");
+        public WindowsElement HistoryListView => TryFindElementByAccessibilityId("HistoryListView");
+        public WindowsElement ClearHistoryButton => TryFindElementByAccessibilityId("ClearHistory");
+
+        /// <summary>
+        /// Opens the History Pane by clicking the History pivot label.
+        /// </summary>
+        public void OpenHistoryPanel()
+        {
+            HistoryLabel.Click();
```
mcooley: nit: our team's C# style in other projects prefers explicit "this." for member access; let's adopt that style here.

(We should also file a work item to set up StyleCopAnalyzers in this project.)

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -0,0 +1,51 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
mcooley: We'll need copyright headers in all these files.

(We should set up StyleCopAnalyzers, which will automatically enforce this and help you add them.)

src/CalculatorUITestFramework/NavigationMenu.cs
```diff
@@ -0,0 +1,105 @@
...
+        Energy,
+        Area,
+        Speed,
+        Time,
+        Power,
+        Data,
+        Pressure,
+        Angle
+    }
+    public class NavigationMenu : ApplicationBase
```
mcooley: If this class is not meant to be instantiated, make it a static class.

Same feedback for the rest of the framework classes.

src/CalculatorUITestFramework/NavigationMenu.cs
```diff
@@ -0,0 +1,105 @@
...
+                    modeAccessibilityId = "Data";
+                    break;
+                case CalculatorMode.Pressure:
+                    modeAccessibilityId = "Pressure";
+                    break;
+                case CalculatorMode.Angle:
+                    modeAccessibilityId = "Angle";
+                    break;
+                default:
+                    modeAccessibilityId = string.Empty;
```
mcooley: If we get here, it's developer error. Throw an exception and crash!

src/CalculatorUITestFramework/NumerPad.cs
```diff
@@ -0,0 +1,77 @@
...
+        public WindowsElement Num8Button => TryFindElementByAccessibilityId("num8Button");
+        public WindowsElement Num9Button => TryFindElementByAccessibilityId("num9Button");
+        public WindowsElement DecimalButton => TryFindElementByAccessibilityId("decimalSeparatorButton");
+        public WindowsElement NegateButton => TryFindElementByAccessibilityId("negateButton");
+
+        /// <summary>
+        /// Translates a number into the Calculator button clicks.
+        /// </summary>
+        /// <param name="number">Number to be entered into the calculator.</param>
+        public void TranslateNumberToButtonClicks(double number)
```
mcooley: naming: "Click" would be a better verb than "Translate". Suggestion: "ClickNumberPadButtons(double number)"

src/CalculatorUITestFramework/NumerPad.cs
```diff
@@ -0,0 +1,77 @@
...
+                        break;
+                    case '9':
+                        Num9Button.Click();
+                        break;
+                    case '.':
+                        DecimalButton.Click();
+                        break;
+                    case '-':
+                        NegateButton.Click();
+                        break;
```
mcooley: Add a default case which throws an exception

src/CalculatorUITestFramework/NumerPad.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Translates a number into the Calculator button clicks.
+        /// </summary>
+        /// <param name="number">Number to be entered into the calculator.</param>
+        public void TranslateNumberToButtonClicks(double number)
+        {
+            string numberStr = number.ToString();
+            if (numberStr.StartsWith("-"))
+            {
+                numberStr = numberStr.Remove(0, 1);
```
mcooley: There's a slightly simpler one-liner for this:

`numberStr = numberStr.Substring(1) + "-";`

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        public static StandardOperatorsPanel StandardOperators = new StandardOperatorsPanel();
+        public static MemoryPanel MemoryPanel = new MemoryPanel();
+        public static HistoryPanel HistoryPanel = new HistoryPanel();
+
+        public static void StandardCalculatorSetup(TestContext context)
+        {
+            // Create session to launch a Calculator window
+            ApplicationSetup(context);
+            // Identify calculator mode by locating the header
```
mcooley: Dead comment?

src/CalculatorUITestFramework/StandardOperatorsPanel.cs
```diff
@@ -0,0 +1,156 @@
...
+        public WindowsElement EqualButton => TryFindElementByAccessibilityId("equalButton");
+        public WindowsElement ClearEntryButton => TryFindElementByAccessibilityId("clearEntryButton");
+        public WindowsElement ClearButton => TryFindElementByAccessibilityId("clearButton");
+        public WindowsElement BackSpaceButton => TryFindElementByAccessibilityId("backSpaceButton");
+
+        /// <summary>
+        /// Uses the Calculator UI to add a list of numbers.
+        /// </summary>
+        /// <param name="numbers">List of numbers to be added.</param>
+        public void Add(List<double> numbers)
```
mcooley: Add, Subtract, etc. feel like the wrong level of abstraction to me. Some options I would like more:

1) remove a layer of abstraction and move this logic directly into the test code
2) create one method, with an implementation similar to NumberPad.TranslateToButtonClicks, which can handle all of these cases, like:
StandardOperatorsPanel.ClickButtons("1+2=");
StandardOperatorsPanel.ClickButtons("3-4=");

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <summary>
+    /// This class contains the UI automation objects and helper methods available when the Calculator is in Standard Mode.
+    /// </summary>
+    public class StandardCalculatorPage : ApplicationBase
+    {
+        public static StandardOperatorsPanel StandardOperators = new StandardOperatorsPanel();
+        public static MemoryPanel MemoryPanel = new MemoryPanel();
+        public static HistoryPanel HistoryPanel = new HistoryPanel();
+
+        public static void StandardCalculatorSetup(TestContext context)
```
mcooley: I don't think StandardCalculatorPage should inherit from ApplicationBase, and it follows that ApplicationSetup should get moved elsewhere (into the test code, probably). After that, this method should get renamed. Suggested name: `public static void NavigateToStandardCalculator()`

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -0,0 +1,269 @@
...
+            StandardCalculatorPage.TearDown();
+        }
+
+        /// <summary>
+        /// Ensures the calculator is in a cleared state
+        /// </summary>
+        [TestInitialize]
+        public void TestInit()
+        {
+            StandardCalculatorPage.ClearAll(false);
```
mcooley: nit: explicit names for boolean parameters for readability: `.ClearAll(clearHistory: false)`

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -0,0 +1,269 @@
...
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            StandardCalculatorPage.ClearAll();
+        }
+
+        #region Basic Arthimetic Tests
+        [TestMethod]
+        public void BVT_Reciprocal()
```
mcooley: I feel like "BVT" is a bit too jargon-y. Consider "Operator_Reciprocal", "Operator_Squared", etc.?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -0,0 +1,269 @@
...
+        public void KeyboardInput_Percent()
+        {
+            StandardCalculatorPage.SendKeys("34+7%=");
+            Assert.AreEqual("36.38", StandardCalculatorPage.GetCalculatorResultText());
+        }
+
+        [Ignore]
+        [TestMethod]
+        public void KeyboardInput_Delete()
+        {
```
mcooley: Don't check in dead code unless we have a bug tracking getting these working again (or Assert.Inconclusive(), or something else which is harder to forget about)

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -0,0 +1,269 @@
...
+            StandardCalculatorPage.MemoryPanel.MemButton.Click();
+            standardOperators.NumberPad.TranslateNumberToButtonClicks(3.3);
+            StandardCalculatorPage.MemoryPanel.MemoryClear.Click();
+            try
+            {
+                StandardCalculatorPage.MemoryPanel.GetAllMemoryListViewItems();
+            }
+            catch (NullReferenceException ex)
+            {
+                if (!ex.StackTrace.Contains("MemoryPanel.GetAllMemoryListViewItems()"))
```
mcooley: What is this supposed to do?

src/CalculatorUITests/StandardModeSmokeTests.cs
```diff
@@ -0,0 +1,92 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Collections.Generic;
+using System;
+using CalculatorUITestFramework;
+
+namespace CalculatorUITests
+{
+    [TestClass]
+    public class StandardModeSmokeTests
```
mcooley: I don't think that we need to separate smoke tests from more thorough tests in the code. We currently have enough capacity to run all the tests all the time, and if we ever need to get more selective I would prefer that we do so by annotating individual tests by priority rather than having separate suites.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -0,0 +1,269 @@
...
+    [TestClass]
+    public class StandardModeFunctionalTests
+    {
+        private StandardOperatorsPanel standardOperators => StandardCalculatorPage.StandardOperators;
+
+        /// <summary>
+        /// Initializes the WinAppDriver web driver session.
+        /// </summary>
+        /// <param name="context"></param>
+        [ClassInitialize]
```
EriWong: >[ClassInitialize] [](start = 8, length = 17)

Should we put this in Assembly Initialize instead?  This will result in us using the same calculator instance across classes, but I think that's ok.

src/CalculatorUITests/StandardModeSmokeTests.cs
EriWong: We can leverage testcategory attributes to mark specific test cases if that's the direction we want to take it in.  

---
In reply to: [283480690](https://github.com/microsoft/calculator/pull/501#discussion_r283480690) [](ancestors = 283480690)

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading.Tasks;
+
+namespace CalculatorUITestFramework
+{
+    public class HistoryPanel : ApplicationBase
+    {
+        public WindowsElement HistoryLabel => TryFindElementByAccessibilityId("HistoryLabel");
+        public WindowsElement HistoryListView => TryFindElementByAccessibilityId("HistoryListView");
+        public WindowsElement ClearHistoryButton => TryFindElementByAccessibilityId("ClearHistory");
+
```
EriWong: Do you think people will need to access these directly in test code?  Might be worth making private instead.

src/CalculatorUITestFramework/NumerPad.cs
EriWong: I kind of like "Input".   Tests then look like "NumberPad.Input(500);".  It's a little less descriptive, but also a lot less verbose.  Thoughts?  

---
In reply to: [283470690](https://github.com/microsoft/calculator/pull/501#discussion_r283470690) [](ancestors = 283470690)

src/CalculatorUITestFramework/ApplicationBase.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        /// <summary>
+        /// Initalizes the WinAppDriver CalculatorSession, starts the session and launches the app.
+        /// </summary>
+        /// <param name="context">The test context passed into VSTest.</param>
+        public static void ApplicationSetup(TestContext context)
+        {
+            // Launches the WinAppDriver.exe in a new Process.
+            winAppDriverProcess = new Process();
+            winAppDriverProcess.StartInfo.FileName = @"c:\Program Files (x86)\Windows Application Driver\winappdriver.exe";
```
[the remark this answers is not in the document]
EriWong: Can we have a fallback of launching this ourselves?  (Or maybe and app.config that configures this).  If we don't have this here, local development gets harder.

---
In reply to: [283453418](https://github.com/microsoft/calculator/pull/501#discussion_r283453418) [](ancestors = 283453418)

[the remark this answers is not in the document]
sanderl: What about just hiding it behind an #if DEBUG?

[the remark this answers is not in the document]
mcooley: I did some more research into how other projects handle this.

If you look [upstream in appium-dotnet-driver](https://github.com/appium/appium-dotnet-driver), there's already code to start and stop Appium--see AppiumLocalService and AppiumServiceBuilder. That won't work for us since we're using WinAppDriver directly instead of full appium, but we could reuse most of that code to build WindowsDriverLocalService and WindowsDriverServiceBuilder. And eventually those patches could be contributed upstream.

As for choosing when to start the service ourselves vs. rely on the WinAppDriver Azure Pipelines task to do it for us...I don't want to make the assumption that Debug===local development in VS and Release===CI pipeline. It would be better to plumb through a flag from the build environment like $(BuildingInVisualStudio). But if we have code in our process which does all the things that AppiumLocalService does (starting and stopping the service in a robust way, reading output from stdout, etc.) then maybe there's no need for the winappdriver pipelines task any more and we don't need the conditional.

cc: @hassanuz

src/CalculatorUITests/StandardModeFunctionalTests.cs
sanderl: This code was refactored to check for the MemoryPaneEmptyLabel instead.

src/CalculatorUITestFramework/HistoryPanel.cs
sanderl: The could need to. It depends on what they are trying to test in the History panel.

src/CalculatorUITestFramework/ApplicationBase.cs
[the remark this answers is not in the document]
sanderl: I've removed the WinAppDriver process start up. I asked @joseartrivera to take a look at the AppiumLocalService and AppiumServiceBuilder.

src/CalculatorUITestFramework/ApplicationBase.cs
```diff
@@ -0,0 +1,205 @@
...
+            // switch the session to that window as follows. You can repeat this logic with any top window with the same
+            // process id (any entry of allWindowHandles)
+            CalculatorSession.SwitchTo().Window(allWindowHandles[0]);
+        }
+
+        /// <summary>
+        /// Sends keyboard key presses to the calculator app.
+        /// </summary>
+        /// <param name="keysToSend"></param>
+        public static void SendKeys(string keysToSend)
```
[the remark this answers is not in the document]
sanderl: I moved this to the WindowsElementExtensions, so now any WindowsElement can call SendKeys.

src/CalculatorUITests/StandardModeFunctionalTests.cs
sanderl: I think we could consider that once we have more classes if we find they all use the same ClassInitialize, but for now, I think it's ok to leave it here.

src/Calculator.sln
```diff
@@ -103,22 +105,6 @@ Global
...
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Debug|x86.ActiveCfg = Debug|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Debug|x86.Build.0 = Debug|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|ARM.ActiveCfg = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|ARM.Build.0 = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|ARM64.ActiveCfg = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|ARM64.Build.0 = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|x64.ActiveCfg = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|x64.Build.0 = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|x86.ActiveCfg = Release|Any CPU
-		{B2C5ADFF-D6B5-48C1-BB8C-571BFD583D7F}.Release|x86.Build.0 = Release|Any CPU
```
mcooley: These lines probably need to stay.

src/CalculatorUITestFramework/WindowsDriverServiceBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+
+namespace CalculatorUITestFramework
+{
+    public class WindowsDriverServiceBuilder
```
mcooley: >WindowsDriverServiceBuilder [](start = 17, length = 27)

Since this includes code from appium-dotnet-driver we need to make sure we include the proper third-party-notices in our project (possibly temporarily, if we ask to take this functionality upstream)

src/Calculator.sln
mcooley: This is why the project isn't building

src/CalculatorUITestFramework/HistoryPanel.cs
sanderl: No, the FindElementsByClassName returns a collection of AppiumWebElements. We can convert them if we want consistency.

src/CalculatorUITestFramework/ApplicationBase.cs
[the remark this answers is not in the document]
joseartrivera: I added this in, as for the command line output it looks like we would only be able to see the output for failed test cases. I think this is probably fine since we would likely only dig into the logs on test failures.

See microsoft/azure-pipelines-tasks#3588

src/CalculatorUITestFramework/NavigationMenu.cs
sanderl: It gets instantiated by the StandardCalculatorPage.cs. My thought was to have the StandardCalculatorPage.cs be the class that brought all of the components together to be used. My thought was that it would be clearer what was available to each mode this way.

src/CalculatorUITestFramework/WindowsElementExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <summary>
+        /// Waits for an element to be displayed until the timeout is reached.
+        /// </summary>
+        /// <param name="element">WindowsElement in the Calculator application.</param>
+        /// <param name="timeout">Timeout in ms.</param>
+        public static void WaitForDisplayed(this WindowsElement element, int timeout = 2000)
+        {
+            Stopwatch timer = new Stopwatch();
+            timer.Reset();
```
EriWong: Is this line needed?

sanderl: The timer was not being reset properly without explicitly calling it.

src/CalculatorUITestFramework/WindowsDriverExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+using OpenQA.Selenium;
+using OpenQA.Selenium.Appium.Windows;
+
+namespace CalculatorUITestFramework
+{
+    public static class WindowsDriverExtensions
+    {
+        /// <summary>
+        /// Wraps the WindowsDriver.FindElementByAccessibilityId and adds retry logic for when the element cannot be found due to WinAppDriver losing the window.
+        /// If FindElementByAccessibilityId fails for a different reason log the error and return null.
```
mcooley: > and return null [](start = 86, length = 16)

nit: this comment is no longer correct (here and below)

src/CalculatorUITestFramework/WindowsElementExtensions.cs
mcooley: How do you know? This should not be needed--I'm wondering if there's something else happening.

---
In reply to: [294546554](https://github.com/microsoft/calculator/pull/501#discussion_r294546554) [](ancestors = 294546554)

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -0,0 +1,62 @@
...
+        public NavigationMenu NavigationMenu = new NavigationMenu();
+        public WindowsElement Header
+        {
+            get
+            {
+                try
+                {
+                    return this.session.TryFindElementByAccessibilityId("Header");
+                }
+                catch
```
mcooley: Should have a comment here explaining why fallback to ContentPresenter is a reasonable thing to do. Also, is there a more specific exception type we can catch?

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public void NavigateToStandardCalculator()
+        {
+            // Ensure that calculator is in standard mode
+            this.NavigationMenu.ChangeCalculatorMode(CalculatorMode.StandardCalculator);
+            Assert.IsNotNull(CalculatorResult);
+        }
+
+        /// <summary>
+        /// Clear the Calculatory display, Memory Panel and optionally the History Panel
```
mcooley: nit: spelling "Calculatory"

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        /// <param name="clearHistory">Bool specifying if the History Panel should be cleared; true by default.</param>
```
mcooley: Param no longer exists

src/CalculatorUITestFramework/WinAppDriver.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.Diagnostics;
+
+namespace CalculatorUITestFramework
+{
+    public sealed class WinAppDriver
+    {
+        private WindowsDriverLocalService windowsDriverService = null;
+        private const string calculatorAppId = "Microsoft.WindowsCalculator.Dev_8wekyb3d8bbwe!App";
+        private static WinAppDriver instance = null;
```
mcooley: In this case, it's appropriate to collapse the singleton boilerplate into a one-liner:

public static WinAppDriver Instance { get; } => new WinAppDriver();

src/CalculatorUITestFramework/WindowsElementExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+            timer.Stop();
+            Assert.Fail(String.Format("{0} was not displayed in {1} ms", element, timeout));
+        }
+
+        /// <summary>
+        /// Sends keyboard key presses to the calculator app.
+        /// </summary>
+        /// <param name="keysToSend"></param>
+        public static void SendKeys(this WindowsElement element, string keysToSend)
```
mcooley: Why is this extension method necessary?

sanderl: I just tested it and it's not needed.

src/CalculatorUITestFramework/WinAppDriver.cs
sanderl: I think this is creating a new instance of WinAppDriver each time Instance is called. I tried this and it caused a bunch of null reference exceptions. So I decided to stay with the original Singleton implementation.

src/CalculatorUITestFramework/StandardCalculatorPage.cs
sanderl: I got this code from the original UI Tests project. I missed it when updating the other properties. It is updated to match the format of the other properties.

src/CalculatorUITestFramework/WinAppDriver.cs
mcooley: I mistyped the comment above. I should have said:

public static WinAppDriver Instance { get; } = new WinAppDriver();

...which will only call the constructor once.

PR #503, opened 2019-05-11
Added the name property for right/left scroll buttons on screen

### LINKED ISSUE #482: Name property of a focusable element must not be null (Scrollable Input)
**Repro Steps**
* Click on the Menu options
* Click on Programmer options
* Chose “BIN” from the list
* Enter number 1 (around >50 times) in the input field to invoke the scroll bar
* Observe Hyperlink for scrolling right/left is missing an accessible name
* Observe the issue using inspect.exe (available in C:\Program Files (x86)\Windows Kits\10\bin\<SDK Version>\x64)

**Actual Result**
Name Property is not defined for left/right buttons.

**Expected Result**
Name Property should be defined

image

MelulekiDube:
-  Added the 'AutomationProperties.Name' tag to the HyperlinkButton template used to create the button
- The above was done for both the right and left buttons
- The left and right buttons now have accessible names as intended.

-Manually
- The fix was tested by running the application and trying to reproduce the original problem described in issue #482. 
- The expected results as specified by the report on the issue was that the 'Name Property should be defined' for both the right and left buttons and this was achieved and tested by using the inspector tool.

### DIFF
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -400,6 +400,7 @@
                                                TextWrapping="NoWrap"/>
                                 </ScrollViewer>
                                 <HyperlinkButton x:Name="ScrollLeft"
+                                                 x:Uid="ScrollLeft"
                                                  Grid.Column="0"
                                                  Width="20"
                                                  MinWidth="20"
@@ -411,7 +412,7 @@
                                                  VerticalContentAlignment="Center"
                                                  Foreground="{ThemeResource SystemControlForegroundAccentBrush}"
                                                  BorderThickness="0"
-                                                 AutomationProperties.Name="ScrollLeft"
+                                                 AutomationProperties.AutomationId="CalculationResultScrollLeft"
                                                  Visibility="Collapsed">
                                     <FontIcon x:Name="ScrollLeftText"
                                               FontFamily="{ThemeResource SymbolThemeFontFamily}"
@@ -419,6 +420,7 @@
                                               Glyph="&#xE26C;"/>
                                 </HyperlinkButton>
                                 <HyperlinkButton x:Name="ScrollRight"
+                                                 x:Uid="ScrollRight"
                                                  Grid.Column="2"
                                                  Width="20"
                                                  MinWidth="20"
@@ -430,7 +432,7 @@
                                                  VerticalContentAlignment="Center"
                                                  Foreground="{ThemeResource SystemControlForegroundAccentBrush}"
                                                  BorderThickness="0"
-                                                 AutomationProperties.Name="ScrollRight"
+                                                 AutomationProperties.AutomationId="CalculationResultScrollRight"
                                                  Visibility="Collapsed">
                                     <FontIcon x:Name="ScrollRightText"
                                               FontFamily="{ThemeResource SymbolThemeFontFamily}"
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -400,7 +400,7 @@
                                                TextWrapping="NoWrap"/>
                                 </ScrollViewer>
                                 <HyperlinkButton x:Name="ScrollLeft"
-                                                 x:Uid="ScrollLeft"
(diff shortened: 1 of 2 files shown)


rudyhuyn:
Hi @MelulekiDube and congratulations for your first PR in this repository!

First of all, I think we can both agree that the Issue filled by @MicrosoftIssueBot wasn't clear.

So here is a better way to explain it: 
In order to help users with visual impairment for example, many tools exist: braille terminals, [Narrator](https://support.microsoft.com/en-us/help/4028598/windows-10-start-or-stop-narrator),... But to be compatible with them, the XAML code must describe what buttons do, especially when they only contain a symbol. In our case, the synthetic voice (or Braille code) will only say _"Link"_ when the focus is on it, not very helpful.

To solve that, we can use the property `AutomationProperties.Name` to describe the action in natural language (with localized strings). When you are used to the `x:Name`, I agree, it's not clear that the value expected is a localized strings and not an id.

Even if the action is only visual, we should keep in mind that 1) it can help someone with visual impairment to show a result to another person. 2) Accessibility benefits everyone, not just those with limited vision and including a lot of scenarios we aren't aware.

So here's what you have to do:
- Add a `x:uid` to these controls to be able to localize their properties
- in the resource file "/strings/en-us/Resources.resw", add a line
`<UidYouCreated>.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name` with a description of the action in English + a comment to help translators to contextualize the text. (you don't need to add it to other .resw, it will be automatically added).

In addition to the name, I think it might also be a good idea to also set an Automation ID via `AutomationProperties.AutomationId` and using a very descriptive name like "CalculationResultScrollLeft" and "CalculationResultScrollRight" to simplify navigation in the visual tree.

I'll be glad to answer any of your questions if you need!

grochocki:
Thanks, @rudyhuyn for that fantastic explanation. Here is another great article on the topic: [Giving your XAML element an accessible name: Part 1 – Introduction](https://blogs.msdn.microsoft.com/winuiautomation/2015/03/19/giving-your-xaml-element-an-accessible-name-part-1-introduction/). I used to work with the author and I learned a lot about building [accessible experiences](https://docs.microsoft.com/en-us/windows/uwp/design/accessibility/accessibility) from him.

MelulekiDube:
@rudyhuyn Thanks for the response. So After reading your response and the articles suggested by @grochocki I realized that the localized strings already existed in the resource file with the ```ScrollLeft``` and ```ScrollRight```. So I have then set the ```x:uid``` and this has resulted in the Accessible name showing up. I was not sure if these string in the resource related to these buttons but I then tried to check were else they might be used. I found that they are not being used for anything but I want o clarify that with you aswell.

So for solving this what I have done now is:

1. Set ```x:uid``` for ```ScrollLeft``` and ```SrollRight``` Buttons to ```ScrollLeft``` and ```ScrollRight``` respectively.
1.  Did not add the line suggested as there was already an entry for that in the  "/strings/en-us/Resources.resw" file.
1. Added the ```AutomationProperties.AutomationId``` for the ```ScrollLeft``` and ```SrollRight``` Buttons within the App.xaml file(I hope this is okay as I assumed since its an ID we don't need to initialize it with a localized String).

I tested the changes with the Narrator as well it seems to be working.
I hope these changes are on the right track.

rudyhuyn:
@MelulekiDube, thank you for the commit!

You are right, there are 2 `scrollLeft`/`scrollRight` keys in the resw but they are in fact not associated to the same control. Moreover, there was a case-issue between your XAML (`ScrollLeft`) and the resw file (`scrollLeft`)

image

`scrollLeft`/`scrollRight` are used by the first line in this image (the expression), while the issue we are fixing is on the second line (`CalculationResult`).

Consequently, we need to create 2 new uid keys (we can share the same uid between controls but it's not a good practice), you can use the same name that you used for AutomationId (`CalculationResultScrollLeft`, `CalculationResultScrollRight`) and add 2 new lines in the resw.

rudyhuyn:
Perfect!! Good job!

PR #507, opened 2019-05-15
Bound the calendar picker to the control above

### LINKED ISSUE #387: Name property of a focusable element must not be null (Calendar date picker)
**Repro Steps**
* Launch Calculator app
* Navigate to Menu and Select Date calculation
* Select Add or subtract days from the combo box
* Observe the issue using inspect.exe (available in C:\Program Files (x86)\Windows Kits\10\bin\<SDK Version>\x64)

**Actual Result**
Name Property is not defined for "From" Calendar picker.

**Expected Result**
Name Property should be defined for "From" Calendar picker. We should be leveraging the control header property instead of drawing our own label.

**Test Environment**
App Name: Calculator 
App Version: 10.1902.16.0

dkheri:
-Initially name of the "From" Calendar picker was bound using the name of the above control
-Bound the name of "From" calendar picker using the id of the above control 

-manually using the inspect tool

### DIFF
--- a/src/Calculator/Views/DateCalculator.xaml
+++ b/src/Calculator/Views/DateCalculator.xaml
@@ -502,7 +502,7 @@
                                 Grid.Row="2"
                                 Margin="0,0,0,0"
                                 Style="{StaticResource DateCalculation_CalendarPickerStyle}"
-                                AutomationProperties.LabeledBy="{x:Bind AddSubtract_Date_FromLabel}"
+                                AutomationProperties.LabeledBy="{x:Bind Date_FromLabel}"
                                 CalendarViewStyle="{StaticResource DateCalculation_CalendarViewStyle}"
                                 Closed="CalendarFlyoutClosed"
                                 DateChanged="AddSubtract_DateChanged"/>
--- a/src/Calculator/Views/DateCalculator.xaml
+++ b/src/Calculator/Views/DateCalculator.xaml
@@ -417,19 +417,13 @@
             </Grid.RowDefinitions>
 
             <!-- From Date -->
-            <TextBlock x:Name="Date_FromLabel"
-                       x:Uid="Date_FromLabel"
-                       Grid.Row="1"
-                       Margin="0,0,0,2"
-                       Foreground="{ThemeResource SystemControlPageTextBaseHighBrush}"
-                       TextWrapping="Wrap"/>
             <CalendarDatePicker x:Name="DateDiff_FromDate"
                                 Grid.Row="2"
                                 Style="{StaticResource DateCalculation_CalendarPickerStyle}"
-                                AutomationProperties.LabeledBy="{x:Bind Date_FromLabel}"
                                 CalendarViewStyle="{StaticResource DateCalculation_CalendarViewStyle}"
                                 Closed="CalendarFlyoutClosed"
-                                DateChanged="FromDate_DateChanged"/>
+                                DateChanged="FromDate_DateChanged"
+                                Header="From"/>
 
             <!-- To Date -->
             <TextBlock x:Name="Date_ToLabel"
@@ -502,7 +496,6 @@
                                 Grid.Row="2"
                                 Margin="0,0,0,0"
                                 Style="{StaticResource DateCalculation_CalendarPickerStyle}"
-                                AutomationProperties.LabeledBy="{x:Bind Date_FromLabel}"
                                 CalendarViewStyle="{StaticResource DateCalculation_CalendarViewStyle}"
                                 Closed="CalendarFlyoutClosed"
                                 DateChanged="AddSubtract_DateChanged"/>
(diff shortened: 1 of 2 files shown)


src/Calculator/Views/DateCalculator.xaml
```diff
@@ -502,7 +502,7 @@
                                 Grid.Row="2"
                                 Margin="0,0,0,0"
                                 Style="{StaticResource DateCalculation_CalendarPickerStyle}"
-                                AutomationProperties.LabeledBy="{x:Bind AddSubtract_Date_FromLabel}"
+                                AutomationProperties.LabeledBy="{x:Bind Date_FromLabel}"
```
grochocki: `CalendarDatePicker` supports [setting header text](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.CalendarDatePicker#setting-a-header-and-placeholder-text), which would collapse this `TextBox` + `CalendarDatePicker` into a single element (and address the accessibility bug as well). I would consider this approach instead of using a custom header `TextBlock` with `LabeledBy`. If this works out, we could simplify the two other instances of `CalendarDatePicker` on this page by leveraging this approach as well.

dkheri:
@grochocki i tried your suggestion and it seems to work. Can you give me feedback if i need to change the other calendar picker as well as part of this pull request or that will be on a separate issue

grochocki:
Awesome! For consistency, let's update the other instances as well as part of this PR so we do not accidentally regress this behavior down the road.

src/Calculator/Views/DateCalculator.xaml
```diff
@@ -417,35 +417,23 @@
...
-                       TextWrapping="Wrap"/>
             <CalendarDatePicker x:Name="DateDiff_FromDate"
                                 Grid.Row="2"
                                 Style="{StaticResource DateCalculation_CalendarPickerStyle}"
-                                AutomationProperties.LabeledBy="{x:Bind Date_FromLabel}"
                                 CalendarViewStyle="{StaticResource DateCalculation_CalendarViewStyle}"
                                 Closed="CalendarFlyoutClosed"
-                                DateChanged="FromDate_DateChanged"/>
+                                DateChanged="FromDate_DateChanged"
+                                Header="From"/>
```
grochocki: In order to make sure Calculator works great across all languages, we avoid putting string resources directly in the XAML. Instead, we reference a list of strings in a separate resource file that we translate into all supported languages so that we can pull from the right language based on the user's locale. Notice how we did not specify the `Text` property directly in the TextBlocks we are removing (i.e., "Date_FromLabel")?

So here is what you need to do for each CalendarDatePicker:
- Add `x:Uid` to each CalendarDatePicker so we can localize the header property (e.g., "DateDiff_FromHeader")
- We can reuse the existing resources, since we are using the same strings but just in a different way. In the resource file "/strings/en-us/Resources.resw", update the resource names of the TextBlock's we are removing to the new UID we are adding to CalendarDatePicker. Here is an example.

Before:

```
<data name="Date_FromLabel.Text" xml:space="preserve">
  <value>From</value>
  <comment>From Date label for Date Picker</comment>
</data>
```

After:

```
<data name="DateDiff_FromHeader.Header" xml:space="preserve">
  <value>From</value>
  <comment>From Date header for Date Picker</comment>
</data>
```

src/Calculator/Views/DateCalculator.xaml
```diff
@@ -417,35 +417,23 @@
             </Grid.RowDefinitions>
 
             <!-- From Date -->
-            <TextBlock x:Name="Date_FromLabel"
-                       x:Uid="Date_FromLabel"
-                       Grid.Row="1"
```
grochocki: Since we are removing some rows, we should update `Grid.RowDefinitions` to remove those rows as well. That also means we should update `Grid.Row` of the controls we are not removing to make sure they align with the new row definitions (e.g., the first CalendarDatePicker would now be `Grid.Row="1"` instead of `Grid.Row="2"`

src/Calculator/Views/DateCalculator.xaml
```diff
@@ -502,7 +490,6 @@
                                 Grid.Row="2"
```
grochocki: It looks like there is one more TextBlock we can consolidate/remove like the others here.

src/Calculator/Views/DateCalculator.xaml
dkheri: I remapped the grid components as per your instruction but components seem to appear to be squished together which i think is due to the absence of the extra padding that was present because of the text block before the calendar picker/text block consolidation.
 
Tried adding extra padding(space occupied by the removed textblocks) and the components seem to be nicely spaced out as before.So should i do this or is there another alternative?

grochocki: Hmm, I see what you mean. Instead of removing rows 0 and 3 (the rows that were adding that padding), try removing rows 1 and 4 (the rows that those removed TextBlocks were in). That _should_ preserve that spacing, but if not, I can take a closer look and suggest an alternative.

dkheri: That seemed to do the trick.

PR #513, opened 2019-05-21
Ensure the calculator uses correctly unary operations when the equal button is pressed

### LINKED ISSUE #264: Strange behaviour after using factorial
**Describe the bug**
Mysterious "+ 1" appears in the history panel after reproducing following steps which leads to an incorrect result.

**Click on buttons in a following order to reproduce**
a. 1
b. +
c. 1
d. n!
e. = (at this point the calculator shows correct result which is 2)
f.  √
g. x<sup>2</sup> (at this point the calculator still shows correct result which is 2)
h. = (at this point the calculator shows incorrect result 3 insted of 2 and "+ 1" appears at the end of the expression in the history panel)

**Screenshots**
obraz

 - OS Build: 10 0 17134 0
 - Architecture: X64

**Additional context**
Ones can be replaced with any other natural numbers and issue will still occur.
In that case mysterious "+ 1" after expression will be replaced with "+ whatever the result of previously counted factorial was".

rudyhuyn:
### Fix unary actions displaying faulty results when a binary operation was previously

a2019-03-08_4-36-01

- Type a random binary operation: `3 + 5` (for example)
- press Enter
- Use an unary action (Power or square root for example)
- press Enter

### Make unary operations (including <kbd>+/-</kbd> ) are repeatable when the equal button is pressed instead of using the last binary action done.

76B914F7-E585-488A-AEAC-906926574955

- Type a random binary operation: 3 + 5 (for example)
- press Enter
- Use an unary action (Power or square root for example)
- press Enter
- press Enter
- press Enter

- Add support of unary actions in `ResolveHighestPrecedenceOperation`
- Modify `ProcessCommandWorker` to set `m_nOpCode`, `m_bNoPrevEqu` when an unary action is used.

- manually tested

### DIFF
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -170,6 +170,8 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
     // BINARY OPERATORS:
     if (IsBinOpCode(wParam))
     {
+        m_bNoPrevEqu = true;
+
         // Change the operation if last input was operation.
         if (IsBinOpCode(m_nLastCom))
         {
@@ -274,13 +276,18 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
         m_lastVal = m_currentVal;
         m_nOpCode = (int)wParam;
         m_HistoryCollector.AddBinOpToHistory(m_nOpCode);
-        m_bNoPrevEqu = m_bChangeOp = true;
+        m_bChangeOp = true;
         return;
     }
 
     // UNARY OPERATORS:
     if (IsUnaryOpCode(wParam) || (wParam == IDC_DEGREES))
     {
+        m_bNoPrevEqu = true;
+        m_bChangeOp = false;
+        m_nPrevOpCode = m_nOpCode;
+        m_nOpCode = (INT)wParam;
+
         /* Functions are unary operations.                            */
         /* If the last thing done was an operator, m_currentVal was cleared. */
         /* In that case we better use the number before the operator  */
@@ -642,7 +649,11 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
             }
             break;
         }
-
+        else
+        {
+            m_nOpCode = IDC_SIGN;
+            m_bNoPrevEqu = true;
+        }
         // Doing +/- while in Record mode is not a unary operation
         if (IsBinOpCode(m_nLastCom))
         {
@@ -655,7 +666,6 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
         }
 
         m_currentVal = -(m_currentVal);
-
         DisplayNum();
         m_HistoryCollector.AddUnaryOpToHistory(IDC_SIGN, m_bInv, m_angletype);
     }
@@ -701,6 +711,13 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
         break;
 
     case IDC_PI:
+
+        m_bNoPrevEqu = true;
+        if (!m_bChangeOp)
+        {
+            m_nOpCode = 0;
+        }
+
         if (!m_fIntegerMode)
         {
             CheckAndAddLastBinOpToHistory(); // pi is like entering the number
@@ -760,12 +777,34 @@ void CCalcEngine::ResolveHighestPrecedenceOperation()
         {
             m_currentVal = m_holdVal;
             DisplayNum(); // to update the m_numberString
-            m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
-            m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            if (IsBinOpCode(m_nOpCode))
+            {
+                m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
+                m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            }
+            else if (m_nOpCode == IDC_SIGN || IsUnaryOpCode(m_nOpCode))
+            {
+                m_HistoryCollector.AddUnaryOpToHistory(m_nOpCode, m_bInv, m_angletype);
+            }
         }
 
         // Do the current or last operation.
-        m_currentVal = DoOperation(m_nOpCode, m_currentVal, m_lastVal);
+        if (!m_bNoPrevEqu && (m_nOpCode == IDC_SIGN || IsUnaryOpCode(m_nOpCode)))
+        {
+            if (m_nOpCode == IDC_SIGN)
+            {
+                m_currentVal = -m_currentVal;
+            }
+            else
+            {
+                m_currentVal = SciCalcFunctions(m_currentVal, (DWORD)m_nOpCode);
+            }
+            m_holdVal = m_currentVal;
+        }
+        else
+        {
+            m_currentVal = DoOperation(m_nOpCode, m_currentVal, m_lastVal);
+        }
         m_nPrevOpCode = m_nOpCode;
         m_lastVal = m_currentVal;
 
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -711,13 +711,6 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
         break;
 
     case IDC_PI:
-
-        m_bNoPrevEqu = true;
-        if (!m_bChangeOp)
-        {
-            m_nOpCode = 0;
-        }
-
         if (!m_fIntegerMode)
         {
             CheckAndAddLastBinOpToHistory(); // pi is like entering the number


HowardWolosky:
Locking this PR for the moment until the Issue is marked `triage approved`.

HowardWolosky:
> Locking this PR for the moment until #264 is marked `triage approved`.

cc: @grochocki

grochocki:
Unlocking PR after issue was approved.

rudyhuyn:
I will update this PR tonight and check if nothing broke it during last 2 months

rudyhuyn:
Merged with master, ready for code review

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -643,7 +650,11 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
             }
             break;
         }
-
+        else
```
HowardWolosky: We talked about _this specific_ aspect of your change this morning, and would like to revert it.  I realize that at the time you were making the change, @grochocki wrote this [comment](https://github.com/microsoft/calculator/issues/264#issuecomment-511076350) in the Issue:

> Based on the conversation here and after looking at what other calculators do for unary operators, I am marking this as approved and unlocking #513 so that hitting "=" will always repeat the last operation.

However, after we played with this change, we realize it doesn't make sense for the <kbd>+/-</kbd> button to benefit from this change.  In fact, I'll be opening up a separate issue shortly to track stopping it from adding `negate()` to the history at all, and simply have it change the sign of the current input/result.  We don't want the <kbd>+/-</kbd> button to be treated like an operator at all, but rather just like an input key, the same way the <kbd>.</kbd> button is.

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -761,12 +771,34 @@ void CCalcEngine::ResolveHighestPrecedenceOperation()
...
             m_currentVal = m_holdVal;
             DisplayNum(); // to update the m_numberString
-            m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
-            m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            if (IsBinOpCode(m_nOpCode))
+            {
+                m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
+                m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            }
+            else if (m_nOpCode == IDC_SIGN || IsUnaryOpCode(m_nOpCode))
```
HowardWolosky: The code here (and below) will have to change as a result of the feedback above.

src/CalcManager/CEngine/scicomm.cpp
rudyhuyn: I'm fine with this change! I will update the PR later tonight

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -170,6 +170,8 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
     // BINARY OPERATORS:
     if (IsBinOpCode(wParam))
     {
+        m_bNoPrevEqu = true;
```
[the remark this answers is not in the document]
rudyhuyn: For binary operators, the current code sets `m_bNoPrevEqu` to True only when the previous operator is not binary (see the `return;`  line 200 in the conditional `if (IsBinOpCode(m_nLastCom)){...}``. 

This shouldn't be necessary, if the previous key is a binary operator, `m_bNoPrevEqu' *should* already be true, but it's better and safer to not make this kind of assumption (a special case that we missed, a bug in another part of the code,...) especially when we do nothing to make sure this assumption is always true.

I guess that's why I moved the code up, to make sure it's still set.

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -761,12 +767,28 @@ void CCalcEngine::ResolveHighestPrecedenceOperation()
...
-            m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
-            m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            if (IsBinOpCode(m_nOpCode))
+            {
+                m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
+                m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            }
+            else if (IsUnaryOpCode(m_nOpCode))
+            {
+                m_HistoryCollector.AddUnaryOpToHistory(m_nOpCode, m_bInv, m_angletype);
```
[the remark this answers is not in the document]
rudyhuyn: IMO this security makes sense and I should also do it in this PR.

`%` is a fake unitary operator, in reality, the result depends on 2 values + an additional operator to calculate the result. 

This PR created a regression with <kbd>%</kbd> I will fix asap:
<kbd>50</kbd> <kbd>+</kbd> <kbd>50</kbd> <kbd>%</kbd>
- before: 75
- after: 25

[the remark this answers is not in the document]
rudyhuyn: - Regression fixed:
<kbd>40</kbd> <kbd>+</kbd> <kbd>50</kbd> <kbd>%</kbd> <kbd>=</kbd> -> 60

- `%` not added to the history. 

We should not. The value calculated after pressing <kbd>%</kbd> is already added to the history, so we shouldn't add the operator.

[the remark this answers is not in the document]
rudyhuyn: Behavior tested with a physical Finance Calculator:
<kbd>40</kbd> <kbd>+</kbd> <kbd>50</kbd> <kbd>%</kbd>
<kbd>=</kbd> -> 60
<kbd>=</kbd> -> 80
<kbd>=</kbd> -> 100

Behavior tested with the Microsoft Calculator (including this PR): same.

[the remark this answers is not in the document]
grochocki: FYI, see [here}(https://github.com/microsoft/calculator/issues/655#issuecomment-527471016) for detailed analysis of how percent key should behave.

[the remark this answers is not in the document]
HowardWolosky: Great, thanks Rudy.

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -761,12 +766,28 @@ void CCalcEngine::ResolveHighestPrecedenceOperation()
...
-            m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
-            m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            if (IsBinOpCode(m_nOpCode))
+            {
+                m_HistoryCollector.AddBinOpToHistory(m_nOpCode, false);
+                m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal); // Adding the repeated last op to history
+            }
+            else if (IsUnaryOpCode(m_nOpCode) && m_nOpCode != IDC_PERCENT)
+            {
+                m_HistoryCollector.AddUnaryOpToHistory(m_nOpCode, m_bInv, m_angletype);
```
HowardWolosky: Sticking with this code from the previous comment...

https://github.com/microsoft/calculator/blob/80912a33d0a4a853ebb43e312999a9fd6197d32e/src/CalcManager/CEngine/scicomm.cpp#L294-L304

Can you clarify why you don't feel like we need to call `m_HistoryCollector.AddOpndToHistory(m_numberString, m_currentVal);` here the way we're doing for the binary operators?

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -761,12 +766,28 @@ void CCalcEngine::ResolveHighestPrecedenceOperation()
         {
```
HowardWolosky: Another side-effect I'm seeing of this change is that we'll still repeat the user's last action even if they deleted/cleared their history.  That behavior was undefined in the discussion within #264.  @grochocki, do you have thoughts on this?  I'm inclined to think that if the user cleared their history (for whatever reason it was that they had to do so), then they're not going to want their previous action to potentially be exposed by someone hitting <kbd>=</kbd>.  It feels like the ability to do this should be based on history being there.

That further opens the question, if the user deletes the most recent history entry, should hitting <kbd>=</kbd> repeat the _actual_ last action that they had, or should it repeat the last action that their _history_ says that they performed?

src/CalcManager/CEngine/scicomm.cpp
rudyhuyn: Of course! The context of this code and the one mentioned are different.

**Line 298:** 
**Context:** the user just entered a unitary operator
When the user enters a unitary operator, there is a high chance that the number he has previously typed (displayed in the CalculationResult control) has not been added to the history, so we must first check if we should add the number before adding the unitary operator.

**Line 774** 
**Context:** the user typed more than 1 time on <kbd>=</kbd>
Users has not typed a number since the last <kbd>=</kbd> (this code is only executed when nothing has been typed since the last one, so we are 100% sure there is no operands to add to the history. The only thing we must do is to add the unit operator.

src/CalcManager/CEngine/scicomm.cpp
rudyhuyn: Interesting question! For information, the current version of Calculator already behaves this way with binary operators:
- Type <kbd>3</kbd> <kbd>+</kbd> <kbd>2</kbd> <kbd>=</kbd>
- Click on `Clear all history` button (the trash icon)
- press <kbd>=</kbd>

In my opinion (for what it's worth), the history is a separate feature and should not change the behavior of the Calculator part but I also understand the possible confusion for users.

rudyhuyn: Another questionable behavior (public version of the app):
- Type <kbd>21</kbd>
- Press <kbd>x²</kbd>
- Press <kbd>CE</kbd>

-> sqrt(21) is added to the History

HowardWolosky: You're right that this is only perpetuating the existing disconnect between history and the repeat-action nature of <kbd>=</kbd>.  Given that, we can table this discussion in a separate Issue and can address it (if we choose to do so) in a separate PR.

rudyhuyn:
PR updated to take give the priority to the most recent binary operator over the last operator if unitary.

Example:
- type `12 + 4²`
- press <kbd>=</kbd>
 -> calculate 12 + 16 = 28
- press <kbd>=</kbd>
 -> must calculate `28 + 16` and not `28²`

000000000000000000000000000000000000004:
ritz car

Karan-05 (approved):

hanzhang54:
Hi @rudyhuyn , there are some conflicts for this PR, would you like to continue work on this? 😀

hanzhang54:
Close this PR as no update for long time. Please feel free to reopen it anytime if you'd like to continue work on this.

PR #524, opened 2019-05-30
Migrate currency converter endpoints

### LINKED ISSUE #77: Currency Endpoints: Update fwlinks to point to new service
The underlying service used for currency conversion will be changed in the coming months. Once the underlying service is updated, we will need to update our fwlinks to point to the new endpoints.

joseartrivera:
We are migrating the REST endpoint that Currency Converter uses.  The REST endpoint is not licensed for your use.  To continue to enable community participation in the development of the feature, developer builds will use an alternate REST endpoint that serves mock data (clearly identifiable as it references planets instead of countries).

Manual testing


src/CalcViewModel/CalcViewModel.vcxproj
```diff
@@ -408,6 +408,17 @@
       <Project>{311e866d-8b93-4609-a691-265941fee101}</Project>
     </ProjectReference>
   </ItemGroup>
+  <ItemDefinitionGroup Condition="!Exists('DataLoaders\DataLoaderConstants.h')">
```
HowardWolosky: It seems like you might be able to simplify this section by using a [`Choose/When/Otherwise`](https://docs.microsoft.com/en-us/visualstudio/msbuild/choose-element-msbuild?view=vs-2019), because then you'd only have to have a single condition check authored for `Condition="!Exists('DataLoaders\DataLoaderConstants.h')`

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -644,6 +644,18 @@ void CurrencyDataLoader::GuaranteeSelectedUnits()
             isConversionTargetSet = true;
         }
     }
+
+    // If still not set for either source or target, just select the first currency in the list
```
HowardWolosky: This change doesn't seem related to switching endpoints.  We should probably pull this specific change out to a separate PR.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -644,6 +644,18 @@ void CurrencyDataLoader::GuaranteeSelectedUnits()
...
+
+    if (!isConversionSourceSet && !m_currencyUnits.empty())
+    {
+        m_currencyUnits[0].isConversionSource = true;
+    }
+
+    if (!isConversionTargetSet && !m_currencyUnits.empty())
+    {
+        m_currencyUnits[0].isConversionTarget = true;
```
HowardWolosky: Seems like you could just have this immediately after `m_currencyUnits[0].isConversionSource = true;` -- no reason to have two identical if conditions.  I would also set `isConversionTargetSet = true;` for completeness/consistency in case something else gets added to this function later on.  Do we already have a separate issue tracking this problem?

src/CalcViewModel/DataLoaders/DataLoaderMockConstants.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+namespace CalculatorApp
+{
+    namespace DataLoaders
+    {
+        static constexpr auto sc_MetadataUriLocalizeFor = L"https://go.microsoft.com/fwlink/?linkid=2091028&localizeFor=";
```
HowardWolosky: nit: I am a _little_ confused why Venus' currency is "Earth Rocks".  It seems a bit odd for a civilization to name their currency after a _different_ civilization...or is it simply that they also use "Earth" to indicate the ground?

README.md
```diff
@@ -54,6 +54,9 @@ Read our [privacy statement](https://go.microsoft.com/fwlink/?LinkId=521839) to
 Telemetry is disabled in development builds by default, and can be enabled with the `SEND_TELEMETRY`
 build flag.
 
+## Currency Converter
+Windows Calculator includes a currency converter feature that uses mock data in developer builds.  The data that Microsoft uses for the currency converter feature (e.g., in the retail version of the application) is not licensed for your use.  The mock data will be clearly identifiable as it references planets instead of countries.
```
HowardWolosky: Please try to keep individual lines in the markdown to no longer than 100-120 characters.  Force a line-break after that.  It will still render just fine, but becomes easier to read (no horizontal scrolling) for those looking at the raw file.

Additionally, now that I see that the mock data is completely static regardless of the `GET` parameter, I would suggest this slightly altered statement (or some similar change):

```suggestion
Windows Calculator includes a currency converter feature that uses mock data in developer builds.  The data that Microsoft uses for the currency converter feature (e.g., in the retail version of the application) is not licensed for your use.  The mock data will be clearly identifiable as it references planets instead of countries, and remains static regardless of selected inputs.
```

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
joseartrivera: Switching the data set to one that does not contain the expected default values (in this case the dollar and the euro) results in the UI not selecting anything by default in the currency converter mode. Since this PR is changing the data set, I think it should be included in this PR.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
joseartrivera: They are not identical, one is setting the source and the other is setting the target. I factored out the check for m_currrencyUnits.empty() though.

src/CalcViewModel/DataLoaders/DataLoaderMockConstants.h
joseartrivera: Their favorite planet is earth, so earth rocks are pretty popular there... but apparently they also use Venusian Seashells according to @EriWong so I have updated the data.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
HowardWolosky: Fair enough.  Keep as-is then.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
HowardWolosky: Right you are.  I read too fast.

src/CalcViewModel/CalcViewModel.vcxproj
```diff
@@ -408,6 +408,23 @@
...
```
HowardWolosky: Any reason this can't go inside the `Choose` as well?

joseartrivera: I did try it out and it seems that choose doesn't allow for ItemDefinitionGroup to be used inside.

README.md
```diff
@@ -54,6 +54,12 @@ Read our [privacy statement](https://go.microsoft.com/fwlink/?LinkId=521839) to
 Telemetry is disabled in development builds by default, and can be enabled with the `SEND_TELEMETRY`
 build flag.
 
+## Currency Converter
+Windows Calculator includes a currency converter feature that uses mock data in developer builds.  The data that
```
mcooley: nit: single space between sentences (here and below)

PR #534, opened 2019-06-03
Wrap json parsing in try/catch

### LINKED ISSUE #77: Currency Endpoints: Update fwlinks to point to new service

joseartrivera:
There is a bug on the service side of the currency converter endpoints that returns an unnecessary null value. Work around this on the client side until a service side fix is found.

Manual testing


src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -502,7 +502,15 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String ^ rawJson, _Inout_ vecto
     staticData.resize(size_t{ data->Size });
     for (unsigned int i = 0; i < data->Size; i++)
     {
-        JsonObject ^ obj = data->GetAt(i)->GetObject();
+        JsonObject ^ obj;
+        try
+        {
+            obj = data->GetAt(i)->GetObject();
+        }
+        catch (...)
```
HowardWolosky: Any chance you can limit the catch to just the exception that you're trying to avoid here?  That would be best practice so that we don't try to continue in the event that we get something like an OOM (Out of Memory) which we clearly can't recover from.

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -531,7 +539,15 @@ bool CurrencyDataLoader::TryParseAllRatiosData(_In_ String ^ rawJson, _Inout_ Cu
     allRatios.clear();
     for (unsigned int i = 0; i < data->Size; i++)
     {
-        JsonObject ^ obj = data->GetAt(i)->GetObject();
+        JsonObject ^ obj;
+        try
+        {
+            obj = data->GetAt(i)->GetObject();
```
HowardWolosky: same

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -502,7 +502,22 @@ bool CurrencyDataLoader::TryParseStaticData(_In_ String ^ rawJson, _Inout_ vecto
     staticData.resize(size_t{ data->Size });
     for (unsigned int i = 0; i < data->Size; i++)
     {
-        JsonObject ^ obj = data->GetAt(i)->GetObject();
+        JsonObject ^ obj;
+        try
+        {
+            obj = data->GetAt(i)->GetObject();
+        }
+        catch (COMException ^ e)
```
HowardWolosky: Given that we're not modifying anything from it...
```suggestion
        catch (const COMException& e)
```

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -531,7 +546,22 @@ bool CurrencyDataLoader::TryParseAllRatiosData(_In_ String ^ rawJson, _Inout_ Cu
     allRatios.clear();
     for (unsigned int i = 0; i < data->Size; i++)
     {
-        JsonObject ^ obj = data->GetAt(i)->GetObject();
+        JsonObject ^ obj;
+        try
+        {
+            obj = data->GetAt(i)->GetObject();
+        }
+        catch (COMException^ e)
```
HowardWolosky: Given that we're not modifying anything from it...
```suggestion
        catch (const COMException& e)
```

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
HowardWolosky: As we've discussed, let's ignore this suggestion for now.

PR #540, opened 2019-06-05
Add CMake project, GCC support

janisozaur:
This is extract from #211 that enables compilation with GCC. With #211
now in the state of bitrot, I would rather try approaching it in smaller
steps that can be hopefully merged quicker, even if it does not provide
full support for all the features #211 provided.

This will _compile_ correctly with my (@janisozaur) GCC, but clang is
more picky about flexible array memebers and refuses to compile it yet.
I will extract remaining parts of #211 in future PRs.

I marked @fwcd as author, as he did most of the work in #211.


janisozaur:
I'm happy with the current state of this PR and I will work on extracting the changes to flexible array members next, in future PR.

CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.13)
```
mcooley: Can we move this one under src/ and avoid having a CMakeLists.txt in the root of the repo? I'd like this to work kind of like our Windows/VS setup does, where everything is contained in src/.

src/CalcManager/UnitConverter.cpp
```diff
@@ -229,7 +229,7 @@ Unit UnitConverter::StringToUnit(const wstring& w)
     vector<wstring> tokenList = StringToVector(w, L";");
     assert(tokenList.size() == EXPECTEDSERIALIZEDUNITTOKENCOUNT);
     Unit serializedUnit;
-    serializedUnit.id = _wtoi(Unquote(tokenList[0]).c_str());
+    serializedUnit.id = wcstol(Unquote(tokenList[0]).c_str(), 0, 10);
```
mcooley: nit: can we use nullptr instead of 0 for the second argument?

src/CalcManager/sal_cross_platform.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
```
mcooley: Do we need #include<sal.h> on Windows/MSVC?

CMakeLists.txt
janisozaur: I'm sure it's technically possible, but it would go against the convention, virtually every project based on cmake had this file in tour directory. I'll see if there's any technical advice for going one way or the other, feel free to the same. Let me know how you want this done.

mcooley: It's true that _more_ projects seem to have it in the root, but there is precedent for having it in a first-level subdirectory (see, for example, [libphonenumber](https://github.com/google/libphonenumber/blob/master/cpp/CMakeLists.txt) or [vcpkg](https://github.com/microsoft/vcpkg/blob/master/toolsrc/CMakeLists.txt)).

Given the layout we currently have for this repo, I'd still prefer it in src/.

janisozaur: While true, you seem to be skipping over the fact that both those repos (and I assume most others) have a main build _thing_ in the root directory. For libphonenumber, there's `pom.xml` and the library is primarily Java, while the cpp version is a port that could very well be extracted to its own repo, while for vcpkg there're `bootstrap-vcpkg.{bat,sh}` scripts in the root of repository.

Personally, I think it makes sense especially e.g. when you have `docs`, `tests`, `assets`, next to `src` that are also consumed when building. That said, I don't see cmake taking place of the already present visual studio project file and there are only files under`src` being used.

I will make the requested change later today.

janisozaur:
All done now.

PR #544, opened 2019-06-06
Port number's mantissa to std::vector

janisozaur:
These are the changes extracted from #211 that port number's mantissa
from flexible member arrays, a non-standard C++ extension, to
std::vector, which enables compilation and linkage with clang.

@fwcd is the original author of these changes, I (@janisozaur) merely
rebased them on top of current master.

The rebase was luckily fairly straight-forward, with the only real
conflict coming from 7a7ceb58882828096cac87494f5a1ffb5359e212
(https://github.com/microsoft/calculator/pull/412), but was trivial to
fix.

### DIFF
--- a/src/CalcManager/CEngine/Number.cpp
+++ b/src/CalcManager/CEngine/Number.cpp
@@ -19,28 +19,22 @@ namespace CalcEngine
     {
     }
 
-    Number::Number(PNUMBER p) noexcept
-        : m_sign{ p->sign }
-        , m_exp{ p->exp }
+    Number::Number(NUMBER p) noexcept
+        : m_sign{ p.sign }
+        , m_exp{ p.exp }
         , m_mantissa{}
     {
-        m_mantissa.reserve(p->cdigit);
-        copy(p->mant, p->mant + p->cdigit, back_inserter(m_mantissa));
+        m_mantissa.reserve(p.cdigit);
+        copy(p.mant.begin(), p.mant.begin() + p.cdigit, back_inserter(m_mantissa));
     }
 
-    PNUMBER Number::ToPNUMBER() const
+    NUMBER Number::ToNUMBER() const
     {
-        PNUMBER ret = _createnum(static_cast<uint32_t>(this->Mantissa().size()) + 1);
-        ret->sign = this->Sign();
-        ret->exp = this->Exp();
-        ret->cdigit = static_cast<int32_t>(this->Mantissa().size());
-
-        MANTTYPE* ptrRet = ret->mant;
-        for (auto const& digit : this->Mantissa())
-        {
-            *ptrRet++ = digit;
-        }
-
+        NUMBER ret = _createnum(static_cast<uint32_t>(this->Mantissa().size()) + 1);
+        ret.sign = this->Sign();
+        ret.exp = this->Exp();
+        ret.cdigit = static_cast<int32_t>(this->Mantissa().size());
+        ret.mant = this->Mantissa();
         return ret;
     }
 
--- a/src/CalcManager/Ratpack/conv.cpp
+++ b/src/CalcManager/Ratpack/conv.cpp
@@ -63,7 +63,7 @@ wchar_t g_decimalSeparator = L'.';
 #define CALC_ULONG_ERROR ((uint32_t)0xffffffffU)
 
 // Used to strip trailing zeros, and prevent combinatorial explosions
-bool stripzeroesnum(_Inout_ NUMBER pnum, int32_t starting);
+bool stripzeroesnum(_Inout_ NUMBER& pnum, int32_t starting);
 
 void SetDecimalSeparator(wchar_t decimalSeparator)
 {
@@ -947,7 +947,7 @@ int32_t numtoi32(_In_ const NUMBER& pnum, uint32_t radix)
 //
 //-----------------------------------------------------------------------------
 
-bool stripzeroesnum(_Inout_ NUMBER pnum, int32_t starting)
+bool stripzeroesnum(_Inout_ NUMBER& pnum, int32_t starting)
 {
     vector<MANTTYPE>::iterator pmant;
     int32_t cdigits;
--- a/src/CalcManager/Ratpack/conv.cpp
+++ b/src/CalcManager/Ratpack/conv.cpp
@@ -101,7 +101,7 @@ namespace
 }
 
 // Used to strip trailing zeros, and prevent combinatorial explosions
-bool stripzeroesnum(_Inout_ PNUMBER pnum, int32_t starting);
+bool stripzeroesnum(_Inout_ NUMBER pnum, int32_t starting);
 
 void SetDecimalSeparator(wchar_t decimalSeparator)
 {
@@ -128,30 +128,12 @@ void* zmalloc(size_t a)
 //
 //-----------------------------------------------------------------------------
 
-void _dupnum(_In_ PNUMBER dest, _In_ const NUMBER* const src)
+void _dupnum(_In_ NUMBER* dest, _In_ const NUMBER* const src)
 {
-    memcpy(dest, src, (int)(sizeof(NUMBER) + ((src)->cdigit) * (sizeof(MANTTYPE))));
-}
-
-//-----------------------------------------------------------------------------
-//
-//    FUNCTION: _destroynum
-//
-//    ARGUMENTS: pointer to a number
-//
-//    RETURN: None
-//
-//    DESCRIPTION: Deletes the number and associated allocation
-//
-//-----------------------------------------------------------------------------
-
-void _destroynum(_Frees_ptr_opt_ PNUMBER pnum)
-
-{
-    if (pnum != nullptr)
-    {
-        free(pnum);
-    }
+    dest->sign = src->sign;
+    dest->exp = src->exp;
+    dest->cdigit = src->cdigit;
+    dest->mant = src->mant;
 }
 
 //-----------------------------------------------------------------------------
@@ -172,9 +154,7 @@ void _destroyrat(_Frees_ptr_opt_ PRAT prat)
 {
     if (prat != nullptr)
     {
-        destroynum(prat->pp);
-        destroynum(prat->pq);
-        free(prat);
+        delete prat;
     }
 }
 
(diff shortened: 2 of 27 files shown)


src/CalcManager/Ratpack/conv.cpp
```diff
@@ -300,47 +263,46 @@ PRAT numtorat(_In_ PNUMBER pin, uint32_t radix)
...
+        cdigits = (uint32_t)a.cdigit;
     }
 
     // scale by the internal base to the internal exponent offset of the LSD
-    numpowi32(&powofnRadix, a->exp + (a->cdigit - cdigits), radix, precision);
+    numpowi32(&powofnRadix, a.exp + (a.cdigit - cdigits), radix, precision);
 
     // Loop over all the relative digits from MSD to LSD
-    for (ptr = &(a->mant[a->cdigit - 1]); cdigits > 0; ptr--, cdigits--)
+    for (ptr = &(a.mant[a.cdigit - 1]); cdigits > 0; ptr--, cdigits--)
```
mcooley: The unit tests are failing, and I think the problem is here--we're calling a.mant[-1].

One reasonable fix would be to rewrite this as a while loop so we check cdigits first:

while(cdigits > 0)
{
    ptr = &(a.mant[a.cdigit - 1]);
    // etc.
    ptr--;
    cdigits--;
}

janisozaur:
I rebased this on top of already approved https://github.com/microsoft/calculator/pull/545 to preempt minor conflict and fixed some issues I found. This might address the warnings raised previously, I'd like to hear back from the test suite, [would be nice to have it fixed](https://github.com/microsoft/calculator/pull/545#issuecomment-500518021)

mcooley:
Sorry for the test pipeline earlier. The unit tests are still failing with the same issue, unfortunately.

janisozaur:
Judging by the error message reported, the tests are failing because of infrastructure still.

> `artifact-engine: unhandled rejection Failed to extract package with error Error: ENOENT: no such file or directory, open 'D:\_work\1\a\drop.zip'`

mcooley:
Still working on this CI infrastructure issue--I'm waiting for another team to investigate.

I tried running the unit tests locally, and I'm still seeing failures. The next one I see is a vector subscript out of range in mulnumx, line 42. There wasn't an obvious logic change there, so a deeper investigation will be needed.

janisozaur:
How can I get size of mantissa in its previous (flexible array) form? The equivalent of `num.mant.size()` in the new form? What is the value of `num->mant[0]` in previous form, when default-initialised? Is that a legal access?

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -184,33 +164,19 @@ void _destroyrat(_In_ PRAT prat)
...
-        {
-            throw(CALC_E_OUTOFMEMORY);
-        }
-    }
-    else
-    {
-        throw(CALC_E_INVALIDRANGE);
-    }
-    return (pnumret);
+    NUMBER num = NUMBER();
```
janisozaur: This would create a default-constructed value, which will have `sign` equal 1: https://github.com/microsoft/calculator/blob/63e4a124ba08bc2ff0afd84164ea0fb9d05335cf/src/CalcManager/Ratpack/ratpak.h#L63

That's unlike the current/old code, which would initialise it to 0 (via `zmalloc` call, which calls `calloc` underneath).

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -1058,10 +1016,10 @@ bool stripzeroesnum(_Inout_ PNUMBER pnum, int32_t starting)
     if (fstrip)
     {
         // Remove them.
-        memmove(pnum->mant, pmant, (int)(cdigits * sizeof(MANTTYPE)));
+        copy(pmant, pmant + cdigits, pnum.mant.begin());
```
janisozaur: In case the elements overlap, the result of `std::copy` is undefined, unlike with `memmove`

src/CalcManager/Ratpack/num.cpp
```diff
@@ -363,116 +356,107 @@ void remnum(PNUMBER* pa, PNUMBER b, uint32_t radix)
...
-    {
-        memmove(c->mant, ptrc, (int)(cdigits * sizeof(MANTTYPE)));
-    }
-
-    // Cleanup table structure
-    for (auto& num : numberList)
+    if (c.mant.begin() != ++ptrc)
     {
-        destroynum(num);
+        copy(ptrc, ptrc + cdigits, c.mant.begin());
```
janisozaur: `memmove` vs `std::copy` again

src/CalcManager/Ratpack/ratpak.h
```diff
@@ -56,20 +58,15 @@ typedef enum eANGLE_TYPE ANGLE_TYPE;
...
-    // This is actually allocated as a continuation of the
-    // NUMBER structure.
-} NUMBER, *PNUMBER, **PPNUMBER;
-#pragma warning(pop)
+    int32_t sign = 1;   // The sign of the mantissa, +1, or -1
+    int32_t cdigit = 0; // The number of digits, or what passes for digits in the
+                        // radix being used.
+    int32_t exp = 0;    // The offset of digits from the radix point
+                        // (decimal point in radix 10)
+    std::vector<MANTTYPE> mant;
```
janisozaur: It seems `cdigit` is used as count of entries in the flexible array in the long form, is that correct? Additionally, `cdigits` usage seems to suggest such arrays used to have at least one element defaulting to zero, is that correct? Since `vector` by default doesn't contain any values, perhaps default-initialising it with a single zero would address the current problems:
```suggestion
    std::vector<MANTTYPE> mant{0};
```

janisozaur:
@mcooley I highlighted some possible differences, could you review them?

mcooley:
Okay, looks like we’ve finally worked around the CI infrastructure issues.

Thanks for flagging all these spots to double-check. I am about to head out on vacation for a few days, but I will take a closer look at these next week.

janisozaur:
@mcooley is there any way I can get Azure to produce actual error messages, instead of the rather unhelpful
```
Error Message:
 Exception Code: C0000005
```
?

src/CalcManager/CEngine/Number.cpp
```diff
@@ -19,28 +19,22 @@ namespace CalcEngine
...
-        : m_sign{ p->sign }
-        , m_exp{ p->exp }
+    Number::Number(NUMBER p) noexcept
+        : m_sign{ p.sign }
+        , m_exp{ p.exp }
         , m_mantissa{}
     {
-        m_mantissa.reserve(p->cdigit);
-        copy(p->mant, p->mant + p->cdigit, back_inserter(m_mantissa));
+        m_mantissa.reserve(p.cdigit);
```
AjayVV: Why not use ``resize`` and thence avoid ``back_inserter`` ?

janisozaur:
All the CI runs are green now! @mcooley can you review now?

src/CalcManager/CEngine/Number.cpp
```diff
@@ -19,28 +19,22 @@ namespace CalcEngine
...
-        ret->exp = this->Exp();
-        ret->cdigit = static_cast<int32_t>(this->Mantissa().size());
-
-        MANTTYPE* ptrRet = ret->mant;
-        for (auto const& digit : this->Mantissa())
-        {
-            *ptrRet++ = digit;
-        }
-
+        NUMBER ret = _createnum(static_cast<uint32_t>(this->Mantissa().size()) + 1);
```
danbelcher-MSFT: Can we get rid of _createnum and use a constructor instead?

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -436,15 +356,14 @@ PRAT StringToRat(bool mantissaIsNegative, wstring_view mantissa, bool exponentIs
     else
     {
         // Mantissa specified, convert to number form.
-        PNUMBER pnummant = StringToNumber(mantissa, radix, precision);
-        if (pnummant == nullptr)
+        optional<NUMBER> pnummant = StringToNumber(mantissa, radix, precision);
```
danbelcher-MSFT: `std::optional` works well with `auto` and pointer semantics:
```suggestion
        auto pnummant = StringToNumber(mantissa, radix, precision);
        if (!pnummant)
        {
            ...
        }
```

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -1149,11 +1068,10 @@ wstring NumberToString(_Inout_ PNUMBER& pnum, int format, uint32_t radix, int32_
         }
     }
 
-    if (round != nullptr)
+    if (round.has_value())
```
danbelcher-MSFT:
```suggestion
    if (round)
```

src/CalcManager/Ratpack/num.cpp
```diff
@@ -302,50 +300,45 @@ void _mulnum(PNUMBER* pa, PNUMBER b, uint32_t radix)
...
         }
 
-        if (lessnum(*pa, tmp))
+        if (lessnum(*pa, *tmp))
         {
             // too far, back up...
-            destroynum(tmp);
             tmp = lasttmp;
-            lasttmp = nullptr;
+            lasttmp = optional<NUMBER>();
```
danbelcher-MSFT:
```suggestion
            lasttmp = nullopt;
```

src/CalcManager/Ratpack/ratpak.h
```diff
@@ -80,8 +77,8 @@ typedef struct _number
 
 typedef struct _rat
 {
-    PNUMBER pp;
-    PNUMBER pq;
+    NUMBER pp;
```
danbelcher-MSFT:
```suggestion
    NUMBER p;
    NUMBER q;
```

The names `pp` and `pq` are shorthand for `pointer to p` and `pointer to q`.  There will be a lot of change in the codebase if you rename these, but we're also adding confusion now that the names have embedded type information that isn't actually correct.

src/CalcManager/Ratpack/ratpak.h
```diff
@@ -192,18 +186,14 @@ extern PRAT rat_min_i32;
...
         OutputDebugString(outputString.str().c_str());                                                                                                         \
     }
-#define destroynum(x)                                                                                                                                          \
-    {                                                                                                                                                          \
-        std::wstringstream outputString;                                                                                                                       \
-        outputString << "destroynum " << x << " file= " << __FILE__ << ", line= " << __LINE__ << "\n";                                                         \
-        OutputDebugString(outputString.str().c_str());                                                                                                         \
-    }                                                                                                                                                          \
-    _destroynum(x), (x) = nullptr
+{
```
danbelcher-MSFT: I don't think this is actually valid?

janisozaur:
I've rebased it and updated to include changes required after #498, but haven't addressed the review comments yet.

janisozaur:
I started writing  some benchmarks using google benchmark, but have never really finished them to a point where they would be worth submitting and eventually lost interest in that, as other things took my time away. I can't say I fully agree with the argument about possible performance loss, when there is no benchmark available in the upstream project, so I will simply address the review comments as they are now.

mcooley:
Closing this since it's become stale and we haven't had time to thoroughly investigate the regressions in our primary environment (Visual Studio on Windows). Thanks for your interest in working on portability; we hope that some of these changes can be contributed in smaller pieces in the future.

janisozaur:
I've rebased the code here to a branch '[mantissa-202110](https://github.com/janisozaur/calculator/tree/mantissa-202110)' in my fork. This is currently under testing, hopefully will submit another PR at some point.

PR #545, opened 2019-06-09
Drop __inline from conflicting declarations

janisozaur:
`mulnum` et al. are declared `extern` in `ratpak.h`, which conflicts
with the `__inline` used with them. Additionally, most similar functions
don't have such keyword applied to them.

See https://github.com/microsoft/calculator/blob/fe30c7cabc1f30c239b5a22ff6d0f7fa0fba5cd1/src/CalcManager/Ratpack/ratpak.h#L463-L464

### DIFF
--- a/src/CalcManager/Ratpack/basex.cpp
+++ b/src/CalcManager/Ratpack/basex.cpp
@@ -34,7 +34,7 @@ void _mulnumx(PNUMBER* pa, PNUMBER b);
 //
 //----------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


janisozaur:
The tests seem to be stuck waiting for agent. Can you restart them?

HowardWolosky:
> The tests seem to be stuck waiting for agent. Can you restart them?

@mcooley is currently working on getting that fixed today.  So, this should be able to get merged in later today or tomorrow. Thanks for your patience.

Issue #338, opened 2019-03-19
Add Graphing Mode


grochocki:
> This feature is in pre-production. Contribute to the in-progress spec **[here](https://github.com/Microsoft/calculator-specs/tree/graphingmode/active/GraphingMode)**.

**Problem Statement**
High school algebra is the gateway to mathematics and all other disciplines of STEM. However, algebra is the single most failed course in high school, as well as the most failed course in community college ([Washington Post](https://www.washingtonpost.com/news/answer-sheet/wp/2017/09/15/of-course-algebra-is-important-its-also-a-huge-problem/)). Graphing capabilities in their daily tools are essential for students who are beginning to explore linear algebra as early as 8th grade. Physical graphing calculators can be expensive, software solutions require licenses and configuration by school IT departments, and online solutions are not always an option. At present, Windows Calculator does not currently have the needed functionality to meet the demands of students.

**Evidence or User Insights**
Graphing calculator provides an accessible and engaging method for students and teachers to learn and teach. Research has shown that graphing calculators foster conceptual understanding of developmental math students (Hopkins & Amelia, 1998; Laughbaum, 2002, 2003; Shore & Shore, 2003) by:
* Increased speed 
* Opportunity to study rich mathematics through technology
* Make connections through different data representations
* Deeper understanding on “why”

Graphing Calculator is also the top requested feature in [Feedback Hub](https://aka.ms/AA43wie).

**Proposal**
Empower students to learn mathematics by improving conceptual understanding and attitudes towards math by adding graphing support to Windows Calculator targeting grades 8-12 students and teachers (math/physics). With graphing calculator, we can improve learning outcomes of students by increasing engagement and visualization of math equations.

**Goals**
* Provide a great baseline graphing calculator experience in Windows Calculator
* Support all [US common core math curriculum](http://www.corestandards.org/Math/), including:
  * Ability to build and interpret functions
  * Understand linear, quadratic, and exponential models
  * Trigonometric functions
  * Reason with equations and inequalities

**Non-Goals**
* Support advanced calculus or statistics functionality
* Parity with more advanced graphing solutions 

**User Cans**
* Users can enter an equation so that it can be viewed on the graph.
* Users can enter multiple equations so that they can compare plots against each other and see the interactions between the lines.
* Users can edit equations so that they can see how changes affect the plot and correct mistakes.
* Users can change the graph viewing window so that they can see different parts of the plot at different levels of detail.
[...]

Issue #453, opened 2019-04-13
Embed as UWP Control


meteorsnows:
**Problem Statement**
More UWP Control Yea.

**Evidence or User Insights**
Third Party like Telerik have it.
[Third party Telerik](https://www.telerik.com/products/wpf/calculator.aspx)

**Proposal**
Have a sample like Telerik.

**Goals**
Official UWP Control for the win. Microsoft The best.

**Non-Goals**
More Choice

**Low-Fidelity Concept**
[High Fidelity from Syncfusion](https://www.syncfusion.com/uwp-ui-controls/calculator)

rudyhuyn:
The calculator is a C++/CX application with a UWP XAML UI, so 100% of the controls you see are already UWP. But I suppose you mean UWP .Net controls (written in C# or VB), in this case it will be a little complicated.

Which control do you miss?

rudyhuyn:
This is a nice idea! You should also post your idea on the Microsoft Community Toolkit page too:
https://github.com/windows-toolkit/WindowsCommunityToolkit/issues

And here:
https://wpdev.uservoice.com/forums/110705-universal-windows-platform/category/193402-uwp-community-toolkit

Someone posted a similar idea some years ago but didn't receive enough upvotes, but Windows changed a lot since Windows 8.1 as well as usage, so you should suggest it again!

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

grochocki:
Thanks for your idea! This is really interesting, and this repo does feel like the right home for discussions on this issue, but I think we need to iron out a few more details first. For example, while we certainly _could_ do this, I think we would want to better understand how many developers would utilize this feature. Leaving this open for discussion to develop the pitch further.

mdtauk:
You could enable the Windows UI Library to take in a version of the calculator engine/core and then some common button types/templates that it can use.  I could see it as the equivalent of the Calendar pop-up control, which can be used on a **Numeric TextBox** - Perhaps a variation of it called **CalculatorTextBox**.

This Numeric TextBox could also include:

- validation for numbers;
- arrow up and down to increment the number;
- inline calculations | like the value "13 + 5" which when pressing enter changes to **18**
- extensions for value masking | like telephone numbers, IP addresses, etc

For examples of the incrementing value and inline calculations - just take a look at the Adobe Creative software suite.

Like Date Picker and Calculator controls.

[...]

PR #547, opened 2019-06-13
Create EquationTextBox control

joseartrivera:
As part of #338, this PR creates the EquationBoxControl which contains some of the styling that will be used in the EquationArea. The control is still a little rough but I wanted to commit what I have so far so others can start building on top of it. 

Manual testing, still a few bugs/kinks to work out which include:
- Existing bug that causes an equation to have a 0 inside of it
- Flickering of icons when entering analysis mode
- Still needs analysis mode visual states
- Still needs actual color picker, using stock color picker for now
- Need to figure out navigation for hover buttons

### DIFF
--- a/src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
@@ -2,11 +2,14 @@
 #include "EquationViewModel.h"
 
 using namespace Windows::UI;
+using namespace Windows::UI::Xaml;
 
 namespace CalculatorApp::ViewModel
 {
     EquationViewModel::EquationViewModel()
-        : m_LineColor{ Colors::Transparent }
+        : m_LineColor{ Colors::Transparent },
+        m_KeyGraphFeaturesVisibility{ ::Visibility::Collapsed },
+        m_Expression{ "" }
     {
     }
 }
--- a/src/CalcViewModel/GraphingCalculator/EquationViewModel.h
+++ b/src/CalcViewModel/GraphingCalculator/EquationViewModel.h
@@ -12,5 +12,6 @@ namespace CalculatorApp::ViewModel
         OBSERVABLE_OBJECT();
         OBSERVABLE_PROPERTY_RW(Platform::String^, Expression);
         OBSERVABLE_PROPERTY_RW(Windows::UI::Color, LineColor);
+        OBSERVABLE_PROPERTY_RW(Windows::UI::Xaml::Visibility, KeyGraphFeaturesVisibility);
     };
 }
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -1096,8 +1096,9 @@
                             <Grid x:Name="SearchBoxGrid">
 
                                 <Grid.Resources>
-                                    <Style x:Name="EquationTextboxButtonStyle" TargetType="Button">
-                                        <Setter Property="Foreground" Value="{ThemeResource TextControlButtonForeground}"/>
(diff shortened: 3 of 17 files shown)


mdtauk:
There may be room for some co-ordination with another control being made
### [NumberBox](https://github.com/microsoft/microsoft-ui-xaml/issues/483#issuecomment-501505054)

And another proposal which was suggested.  Similar to the Calender Date Picker - a control that when pressed would bring up a mini Calculator in a flyout.

SavoySchuler:
@joseartrivera it would be a win for us if we could help calculator reduce the volume of code it has to maintain. Our requirements are still in the refining process at the proposal/spec link @mdtauk shared. Please let me know your thoughts!

Thanks for your eagle eyes across the Microsoft repos, @mdtauk!

grochocki:
> it would be a win for us if we could help calculator reduce the volume of code it has to maintain

Thanks, @mdtauk and @SavoySchuler for the suggestion! I've been monitoring the new NumberBox control, and it is _really_ cool. In fact, we sent over a similar duplicate request we saw here in our repo (#453), since this felt like it would be better served my a platform control. I am glad to see it is progressing.

Unfortunately, I do not think NumberBox meets our needs for this feature. We are a little behind on publishing more details about this particular component of Graphing Calculator ([spec](https://github.com/Microsoft/calculator-specs/tree/graphingmode/active/GraphingMode)), but we ultimately need much more customization of the entire equation input control as well as support for variables and math symbols.

Here are some renders:
image

mdtauk:
> > it would be a win for us if we could help calculator reduce the volume of code it has to maintain
> 
> Thanks, @mdtauk and @SavoySchuler for the suggestion! I've been monitoring the new NumberBox control, and it is _really_ cool. In fact, we sent over a similar duplicate request we saw here in our repo (#453), since this felt like it would be better served my a platform control. I am glad to see it is progressing.
> 
> Unfortunately, I do not think NumberBox meets our needs for this feature. We are a little behind on publishing more details about this particular component of Graphing Calculator ([spec](https://github.com/Microsoft/calculator-specs/tree/graphingmode/active/GraphingMode)), but we ultimately need much more customization of the entire equation input control as well as support for variables and math symbols.
> 
> Here are some renders:
> image

The Slider design is getting updated soon, so hopefully Calculator will just get that for free, unless you use a custom template.

Also the new controls are likely to use a 1 epx thick border, which that EquationBox seems to use, but it is missing the 2 epx CornerRadius.

I also proposed a formal [Prefix and Suffix content property](https://github.com/microsoft/microsoft-ui-xaml/issues/784) for the TextBox control, and this EquationBox looks like it is doing something similar.  I am not sure if the proposal will be taken up, but it would be good to come up with a common design for it, which your EquationBox could define...

Comparing Similar Controls

When making designs for the NumberBox - I tried to come up with a design which could sit alongside the TextField like controls which share a similar look right now.

joseartrivera:
> There's a perf issue when adding new functions. It looks like it is re-rendering the everything that was added except the first one each time you click the add button.
> 
> Since you mentioned this is an initial check-in, I don't think it should block this review, but I wanted to let you know about the issue.

Thanks for the buddy build! I noticed this too, I went ahead and added a fix to the perf issues but I think we should consider a different approach for how/when we update the graph. This should fix most of the perf issues for now though. I also merged with your work earlier this week and made the graph initialize without an equation.

grochocki:
>The Slider design is getting updated soon, so hopefully Calculator will just get that for free, unless you use a custom template.

>Also the new controls are likely to use a 1 epx thick border, which that EquationBox seems to use, but it is missing the 2 epx CornerRadius.

Yep, we monitor design changes like this closely and will usually audit the entire app before bumping up MaxVersionTested to make sure we can account for these changes.

PR #553, opened 2019-06-16
Handle date difference for ends of months

### LINKED ISSUE #552: Date difference calculation wrong in certain cases
**Describe the bug**
When calculating the difference between dates near the ends of months the results are sometimes wrong. I.E. 31st July - 30th December 2019 results in "5 moths; 613566756 weeks; 3 days".

**Steps To Reproduce**
1. Open Calculator 
2. Switch to date calculation mode
3. Go to the difference between dates - mode
4. Enter start-date 31st of July 2019
4. Enter end-date 30th of December 2019 (might be 31th of December to cause the bug in a different time-zone)

**Expected behavior**
Display a difference that is closer to reality. What the "correct" difference is, is sort of up to debate but 613566756 weeks is definitely not it. Fixing this bug results in "4 Months; 4 Weeks; 2 Days" which is kind of silly but technically correct according to the calculation method of "31st July + 4Months = November 30th" that Windows::Globalization::Calendar appears to use.

**Screenshots**

image

 - OS Build: 10.0.18362.0
 - Architecture: X64
 - Region: de-DE
 - Dev Version Installed: True

**Additional context**
Originally reported by /u/RLWLB2985 on /r/Windows10

**Requested Assignment**
I have the minimal possible solution already implemented.

Bigpet:
The incrementally calling `Window::Globalization::Calendar::AddMonths`
resulted in a negative value for `GetDifferenceInDays` which was then
assigned to an unsigned variable `daysDiff`.

One example of the issue when running the calculator in UTC+2 was the
difference between July 31st and December 30th.

The initial guess was 4 months which then landed on November 30th.
This date was stored and then in the loop incremeted by one month.
This then landed precisely on the end date December 30th.
After the loop the final value is then used July 31st + 5 months
which results in the 31st of December.
The resulting difference of -1 days is then assigned to the unsigned
value `daysDiff`.

This commit makes the minimal changes to remedy this bug.
It makes sure to only ever call `AddMonths` with the same starting date
instead of incrementally to different dates.

This commit makes the minimal changes to remedy this bug.
It makes sure to only ever call `AddMonths` with the same starting date
instead of incrementally to different dates.

- manually tested dates near the end of month where this incremental calculation would be triggered

### DIFF
--- a/src/CalcViewModel/Common/DateCalculator.cpp
+++ b/src/CalcViewModel/Common/DateCalculator.cpp
@@ -206,7 +206,7 @@ void DateCalculationEngine::GetDateDifference(_In_ DateTime date1, _In_ DateTime
                             // pivotDate is still below the end date
                             try
                             {
-                                pivotDate = AdjustCalendarDate(pivotDate, dateUnit, 1);
+                                pivotDate = AdjustCalendarDate(tempPivotDate, dateUnit, static_cast<int>(differenceInDates[unitIndex] + 1));
                                 differenceInDates[unitIndex] += 1;
                             }
                             catch (Platform::InvalidArgumentException ^)


wilgert:
Maybe also add a Unit Test that replicates the bug? So it will never appear again!

PR #555, opened 2019-06-17
Date difference: display the difference in days (only) when we aren't able to calculate the difference in days/weeks/months..

### LINKED ISSUE #554: Prevent Date Difference to display faulty results
**Describe the bug**

When `DateCalculationEngine::GetDateDifference` isn't able to correctly calculate the number of years, months, weeks, days between two dates, the function currently returns an incorrect result, as seen with #552 (which focuses on a specific calculation error).

In order to prevent any further bugs in the future, we must add a way to indicate that the function weren't able to calculate correctly the difference and modify the ViewModel to display instead the number of days (only) between the two dates.

Note: This proposal doesn't try to fix #552 (#553 does it already), but prevent all future similar bugs.

**Steps To Reproduce**
- Open Calculator
- Switch to date calculation mode
- Go to the difference between dates - mode
- Enter start-date 31st of July 2019
- Enter end-date 30th of December 2019 (might be 31th of December to cause the bug in a different time-zone)

**Expected behavior**

**Screenshots**
image
Left: faulty result displayed
Right: display the difference in days when we aren't able to calculate the difference in year/month/week/day

 - OS Build: 10.0.18362.0
 - Architecture: X64
 - Region: en-US
 - Dev Version Installed: True

**Requested Assignment**

rudyhuyn:
## Fixes #554 

Modify how the Calculator behaves when `DateCalculationEngine::GetDateDifference` isn't able to calculate correctly the number of years, months, weeks, days between two dates.

Instead of displaying an incorrect result, the UI will now display the difference in days (only). 

- Add new constant `DateDifferenceUnknown`
- rename `TryGetDateDifference` to `TryGetDateDifference` and modify the prototype to return a boolean.
- when we aren't able to calculate correctly the result, return false
- Modify `DateCalculatorViewModel::OnInputsChanged()` to display the difference in Days (only) when we were not able to calculate the difference in days/weeks/months 

Manually.
This diff can be test with the same repro as #552 (if still not fixed).

### DIFF
--- a/src/CalcViewModel/DateCalculatorViewModel.cpp
+++ b/src/CalcViewModel/DateCalculatorViewModel.cpp
@@ -174,6 +174,7 @@ void DateCalculatorViewModel::UpdateDisplayResult()
     {
         if (m_dateDiffResultInDays == DateDifferenceUnknown)
         {
+            IsDiffInDays = false;
             StrDateDiffResultInDays = L"";
             StrDateDiffResult = L"";
         }
@@ -184,7 +185,8 @@ void DateCalculatorViewModel::UpdateDisplayResult()
             StrDateDiffResultInDays = L"";
             StrDateDiffResult = AppResourceProvider::GetInstance().GetResourceString(L"Date_SameDates");
         }
-        else if ((m_dateDiffResult.year == 0) && (m_dateDiffResult.month == 0) && (m_dateDiffResult.week == 0))
+        else if (m_dateDiffResult == DateDifferenceUnknown ||
+            (m_dateDiffResult.year == 0 && m_dateDiffResult.month == 0 && m_dateDiffResult.week == 0))
         {
             IsDiffInDays = true;
             StrDateDiffResultInDays = L"";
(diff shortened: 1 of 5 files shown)


src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -159,14 +172,21 @@ void DateCalculatorViewModel::UpdateDisplayResult()
 {
     if (m_IsDateDiffMode)
     {
-        // Are to and from dates the same
-        if (m_dateDiffResultInDays.day == 0)
+        if (m_dateDiffResultInDays == DateDifferenceUnknown)
+        {
+            IsDiffInDays = false;
+            StrDateDiffResultInDays = L"";
+            StrDateDiffResult = L"";
```
EriWong: >L"" [](start = 32, length = 3)

I wasn't able to hit this case when playing around, but it's probably worth adding a "Date calculation failed" message.  A blank result can give the user the impression that something is going on when in fact, nothing is still running.

rudyhuyn: This case _should_ never happen but we still need to manage it (in case of a future diff breaks it).

I will change the string.

Issue #268, opened 2019-03-11
Date Calculation not working currently


degant:
**Describe the bug**
Date Calculation produces different results for dates in the past vs dates in the future

**Steps To Reproduce**
1. Go to Date Calculation
2. Select current date as From
3. Select current date as To
   - Difference says 'Same dates'
4. Now change from date to one day after current date
   - Difference says '1 day'
5. Now change from date to one day before current date
   - Difference says 'Same dates'

**Expected behavior**
It should say 1 day for one day before

**Screenshots**
<img src="https://user-images.githubusercontent.com/6065982/54163952-0547b480-4418-11e9-96d3-b0dd20ef4a95.gif" width="300">

<img src="https://user-images.githubusercontent.com/6065982/54164016-48a22300-4418-11e9-804b-c66a3358d359.gif" width="300">

**Device and Application Information (please complete the following information):**
 - OS Build: 10.0.17763.0
 - Architecture: X64
 - Application Version: 10.1812.10048.0

rudyhuyn:
Duplicate of https://github.com/Microsoft/calculator/issues/178

This is an issue due to Daylight Time Saving (yesterday in the US, later this month in Europe) 

A patch is already available here, don't hesitate to review it!
https://github.com/Microsoft/calculator/pull/193

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

PR #558, opened 2019-06-18
Modify ClipTime to adjust the datetime to local time before clipping 

### LINKED ISSUE #557: Date Difference: dates returned by CalendarDatePicker and used to calculate the difference can be shifted depending on the current time and time zone.
## Describe the bug

In C++/CX, a DateTime is represented using a [`FILETIME`](https://docs.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-filetime), number of seconds since 01/01/1601 12:00am UTC. Obviously, this way of representing a date has no notion of local time/UTC time, by design, date are always UTC.

This is an issue with `CalendarDatePicker`, returning the date selected (with hours and minutes equal to the current time) in **local time**.

## What the calculator currently does
- User selects a date using the `CalendarDatePicker`
   - for example: __06/20/2019__
- `CalendarDatePicker` will "concatenate" this date with the current time (11:16pm local time)
   - _06/20/2019 11:16PM GMT-7_
- Because we use C++/CX, this date will be converted in FILETIME
   - _06/20/2019 11:16PM GMT-7_ -> _06/21/2019 6:16AM GMT_ --> _132055713600000000_
- The ViewModel will "clip" this datetime to the time to midnight.
  - _132055713600000000_ -> _132055488000000000_

The issue here is that _132055488000000000_ represents 06/21/2019, not 06/20/2019 as the user expected.

This shift shouldn't impact the result of the difference, both `FromDate` and `ToDate` being shifted the same way, but it's an issue for us developers, making bugs harder to detect/reproduce and making unit tests + code execution less deterministic.

## Expected behavior
The Calculator must compute the difference between 2 dates exactly the same way whatever the timezone, `CalendarDatePicker::Date` must be adjusted to represent the day selected at midnight UTC time.

In the example, the date returned should be **132054624000000000**.

## Screenshots

When users select 06/20/2019:

image

equivalent to _GMT: Friday, June 21, 2019 6:36:07 AM_, then "clipped" to set the time to midnight: _GMT: Friday, June **21**, 2019 12:00:00 PM_

==> should be Friday, June 20 instead.

## Device and Application Information
 - OS Build: 10.0.18362.0
 - Architecture: X64
 - Region: en-US
 - Dev Version Installed: True

## Requested Assignment

rudyhuyn:
## Fixes #557 

In some cases (depending on the current time and time zone), the calculation of the difference between two dates was based on the previous or next day of the selected one.

To fix this issue, we need to "convert" FromDate and ToDate to local time (because CalendarDatePicker uses local time to generate the datetime) before clipping the time. 

This PR will make sure that `FromDate` and `ToDate` are correctly set and really match what the user selected.

- Modify `ClipTime` and add an extra parameter to "convert" the current UTC FILETIME to Local FILETIME prior to clipping the time.

- Manually tested with different time zones, date and time.

### DIFF
--- a/src/CalcViewModel/DateCalculatorViewModel.cpp
+++ b/src/CalcViewModel/DateCalculatorViewModel.cpp
@@ -60,8 +60,7 @@ DateCalculatorViewModel::DateCalculatorViewModel()
     auto today = calendar->GetDateTime();
 
     // FromDate and ToDate should be clipped (adjusted to a consistent hour in UTC)
-    m_fromDate = ClipTime(today);
-    m_toDate = ClipTime(today);
+    m_fromDate = m_toDate = ClipTime(today, true);
 
     // StartDate should not be clipped
     m_startDate = today;
@@ -84,7 +83,7 @@ DateCalculatorViewModel::DateCalculatorViewModel()
 
     DayOfWeek trueDayOfWeek = calendar->DayOfWeek;
 
-    DateTime clippedTime = ClipTime(today);
+    DateTime clippedTime = ClipTime(today, false);
     calendar->SetDateTime(clippedTime);
     if (calendar->DayOfWeek != trueDayOfWeek)
     {
@@ -119,8 +118,8 @@ void DateCalculatorViewModel::OnInputsChanged()
 
     if (m_IsDateDiffMode)
     {
-        DateTime clippedFromDate = ClipTime(FromDate);
-        DateTime clippedToDate = ClipTime(ToDate);
+        DateTime clippedFromDate = ClipTime(FromDate, true);
+        DateTime clippedToDate = ClipTime(ToDate, true);
 
         // Calculate difference between two dates
         m_dateCalcEngine->GetDateDifference(clippedFromDate, clippedToDate, m_allDateUnitsOutputFormat, &dateDiff);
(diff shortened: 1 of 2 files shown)


src/CalcViewModel/DateCalculatorViewModel.cpp
```diff
@@ -60,8 +60,7 @@ DateCalculatorViewModel::DateCalculatorViewModel()
     auto today = calendar->GetDateTime();
 
     // FromDate and ToDate should be clipped (adjusted to a consistent hour in UTC)
```
EriWong: >(adjusted to a consistent hour in UTC) [](start = 45, length = 38)

I believe this comment is no longer correct. Mind fixing it?

EriWong:
>     m_dateResult = today;

I like your fix in that it will make debugging issues easier overall.  Does it make sense to use local times for addition and subtraction of dates as well?  I kind of go back and forth on this, as it might not because that math is easier to wrap my head around (in my mind at least).

---
Refers to: src/CalcViewModel/DateCalculatorViewModel.cpp:67 in 1e17d32. [](commit_id = 1e17d32fa6925a1642c5271dc38c599724253255, deletion_comment = False)

src/CalcViewModel/DateCalculatorViewModel.cpp
rudyhuyn: It's still the case, `FromDate` and `ToDate` must be clipped to midnight UTC time.

rudyhuyn:
> > ```
> > m_dateResult = today;
> > ```
> 
> I like your fix in that it will make debugging issues easier overall. Does it make sense to use local times for addition and subtraction of dates as well? I kind of go back and forth on this, as it might not because that math is easier to wrap my head around (in my mind at least).
> 
> Refers to: src/CalcViewModel/DateCalculatorViewModel.cpp:67 in [1e17d32](https://github.com/microsoft/calculator/commit/1e17d32fa6925a1642c5271dc38c599724253255). [](commit_id = [1e17d32](https://github.com/microsoft/calculator/commit/1e17d32fa6925a1642c5271dc38c599724253255), deletion_comment = False)

The calculation must use UTC because of DTS (see this old issue: https://github.com/microsoft/calculator/issues/268)

PR #564, opened 2019-06-24
Update .gitignore to ignore VSCode settings directories

HowardWolosky:
Some users (including this one) occasionally use VSCode for
inspecting/updating the project, so let's keep `git status`
clean by ignoring the VSCode user settings directory.

- tells `git` to Ignore the `.vscode` directory within the root directory as well as any sub-directory

- Added a `.vscode` directory to my root and a number of random sub-directories to my clone and ensured that `git status` reported that there were no new files to be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ bld/
 [Oo]bj/
 [Ll]og/
 
+# Visual Studio Code user setting files
+**/.vscode/
+
 # Visual Studio 2015 cache/options directory
 .vs/
 # Uncomment if you have tasks that create the project's static files in wwwroot


mcooley:
Have you considered adding a [global .gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) on your machine, so this works for all your git repos? The [github/gitignore repo has examples you can use.](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

If you still think it's better to take the change in this file, I'm okay with that too. (But please add to the bottom of the file so it's easier to diff with [upstream](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore).)

HowardWolosky:
> Have you considered adding a global .gitignore on your machine, so this works for all your git repos?

I hadn't.  That's a reasonable suggestion for this. We can consider revisiting this is we're finding that there are a larger number of users that are using  VSCode with this project and running into the issue.

Thanks for the suggestion.

Issue #266, opened 2019-03-11
Converter doesn't automatically flip from & to fields during conversion


degant:
**Describe the bug**
Converter doesn't easily flip the from & to fields during conversions (Temperature, Volume, Speed etc.)

**Steps To Reproduce**
<!--Steps to reproduce the behavior:
1. Launch Calculator
2. Click on Menu and open the Temperature Converter
3. It loads Celsius to Fahrenheit
4. Try to flip the conversion
5. Change first field from Celsius to Fahrenheit
6. Change second field from Fahrenheit to Celsius

**Expected behavior**
It would be nice if the converter could automatically flip the from & to values during conversion. If a Celsius-> Fahrenheit conversion is changed to Fahrenheit -> Fahrenheit conversion then it should be modify it to Fahrenheit -> Celsius similar to how most of these converters work online. In the above steps if the user shouldn't have to perform step #6.

**Screenshots**
![calculator](https://user-images.githubusercontent.com/6065982/54163204-7fc30500-4415-11e9-8292-8d31e15e7872.gif)

**Device and Application Information (please complete the following information):**
 - OS Build: 10.0.17763.0
 - Architecture: X64
 - Application Version: 10.1812.10048.0

**Additional context**
Might also be good to add a flip or exchange button to directly flip the to & from fields.

rudyhuyn:
> It would be nice if the converter could automatically flip the from & to values during conversion.
Good idea!

> Might also be good to add a flip or exchange button to directly flip the to & from fields.
This part is not necessary, users can directly click on the second field and modify the value.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

grochocki:
Thanks for the report! I really appreciate your thoroughness when filing this bug. The behavior as described is currently by design, though this is also not the first time this suggestion has come up (#172).

>>Might also be good to add a flip or exchange button to directly flip the to & from fields.

>This part is not necessary, users can directly click on the second field and modify the value.

With the current design, I agree, but I also think we can build a better user expereince in general for converters, as I think having mutliple inputs for the user to manage is not super straightforward.

For example, currency converters seem to follow a similar pattern to eachother that offer an input, and output, and a fast way to switch between the two:

**Bing**
![bing](https://user-images.githubusercontent.com/1414423/54236480-d93c3a00-44d0-11e9-9bdd-11c0d0dc0616.jpg)

**Google**
[...]

PR #566, opened 2019-06-24
Ensure consistent margins/padding with currency converter keypad

### LINKED ISSUE #87: Uneven NumberPad Margin in Converter Mode
**Describe the bug**
Margin around the number pad should be of equal value. 

**Steps To Reproduce**
1. Go to any converter 
2. Decrease window width until the vertical layout VisualState is triggered

** Screenshot **
Observe the following layout

image

**Expected behavior**
Padding should be equal 

**Recommended Solution**
There are two options:

1. Decrease Right and Left margin to match the bottom (this would emulate the padding used on the main calculator interface)
2. Increase the bottom margin to match the sides.

 - OS Build: 18346
 - Architecture: x64

HowardWolosky:
- Simplifies the layout of `ConverterNumPad` to match that  seen of `NumpadPanel` in Calculator.xaml.
   - They now both use the same static margin ('`3,0,3,3`)
   - Removed two "gutter" columns that had been used for dynamic margins.

> The `GutterLeft` and `GutterRight` columns were left untouched on purpose.
> They are only activated when in a "landscape" width/height orientation, used to ensure that the app display is evenly split 50/50 between the input fields and the number pad.
> `GutterLeft` gives a margin to the input field column, while `GutterRight` is used to ensure that the number pad column has an equal width available to it (as the numberpad is displayed in a column span that merges its column with that of `GutterRight`.

- Validated with a "screen ruler" that the left/right/bottom margins were the same within Volume.

Volume (before and after)

- Validated with a "screen ruler" that the left/top/bottom in Volume were identical to those seen in Standard mode.
Existing standard vs new Volume

- Verified the margins in "wide" mode as well
Volume (wide) (before and after)

### DIFF
--- a/src/Calculator/Views/UnitConverter.xaml
+++ b/src/Calculator/Views/UnitConverter.xaml
@@ -321,8 +321,6 @@
                                                      Threshold="1"/>
                     </VisualState.StateTriggers>
                     <VisualState.Setters>
-                        <Setter Target="GutterTop.Height" Value="0.5*"/>
-                        <Setter Target="GutterBottom.Height" Value="0.5*"/>
                         <Setter Target="GutterLeft.Width" Value="48"/>
                         <Setter Target="GutterRight.Width" Value="48"/>
                         <Setter Target="ColumnLeft.Width" Value="1*"/>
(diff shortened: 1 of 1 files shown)


mdtauk:
These converter pages should reuse the same Number Pad, and screen layout.  Perhaps there should be a template page included so future additional pages will be consistent.

grochocki:
> These converter pages should reuse the same Number Pad, and screen layout.

Visually, it is the same Number Pad, no? I agree this should be one component, if it is not already.

IRT overall page layout, I think we need to consider a broader redesign of the converter interface. I touch on that here a bit [here](https://github.com/microsoft/calculator/issues/266#issuecomment-472184587), but I think there might be an opportunity to collapse converters so instead of navigating to an entirely different page, you can change between units on one page.

PR #569, opened 2019-06-25
edit instance condition

RaisingRay:
using a conditional operator instead of a basic condition is a good practice and the code looks neat.

and i am new  here so please reply, that would make my day even i am sure i didn't follow your instructions properly.

### DIFF
--- a/src/CalculatorUITestFramework/WinAppDriver.cs
+++ b/src/CalculatorUITestFramework/WinAppDriver.cs
@@ -18,11 +18,7 @@ namespace CalculatorUITestFramework
         {
             get
             {
-                if (instance == null)
-                {
-                    instance = new WinAppDriver();
-                }
-                return instance;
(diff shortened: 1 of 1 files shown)


mcooley:
Hi @RaisingRay,

Thanks for your interest in contributing!

Unfortunately we can't accept this change because it doesn't work correctly--it never sets the "instance" field so ```new WinAppDriver()``` is called every time. Also, as mentioned in our [guidelines](https://github.com/microsoft/calculator/blob/master/CONTRIBUTING.md#contributions-we-accept) we do not accept changes only to code formatting and style, such as switching a conditional block to a ternary.

I think a better way to improve this code would be to use a static initializer as mentioned in [this comment](https://github.com/microsoft/calculator/pull/501#discussion_r297421142)--you might consider testing and then submitting that change instead.

I will close this PR for now, but we're glad you're here, and we hope you enjoy continuing to code with Calculator!

PR #572, opened 2019-06-27
Updated Calculator diagnostic data collection per the specification

### LINKED ISSUE #525: Refresh Calculator Diagnostic Data
To make decisions about what user experiences to focus on, we need to understand which parts of calculator are used most often and whether they're working correctly. We want to update the app-specific diagnostic information Calculator is collecting to answer these questions:

1. Is Calculator performing?
2. Which Calculator mode is the best default experience?
3. Can users discover all the calculators and converters in the app?
4. What changes can we make to the keyboard layout to optimize for the most used operators?
5. Are History and Memory features used by the users?
6. Does Calculator need to support multi-instance?
7. Where is the product non-fatally failing (error/exception)?

sanderl:
- Removed unneeded diagnostic events and code
- Added and consolidated events into the events defined in the spec

- Tested manually to ensure the diagnostic events were logged as expected


src/Calculator/AboutFlyout.xaml.cpp
```diff
@@ -64,5 +64,4 @@ void AboutFlyout::SetDefaultFocus()
 
 void CalculatorApp::AboutFlyout::UserControl_Loaded(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
EriWong: We should probably remove the entire event handler if we aren't doing anything with it.

src/Calculator/Views/Calculator.xaml.cpp
```diff
@@ -702,12 +696,11 @@ void Calculator::OnMemoryAccessKeyInvoked(_In_ UIElement ^ sender, _In_ AccessKe
...
-    if (DockPivot->SelectedIndex == 0)
-    {
-        TraceLogger::GetInstance().LogHistoryBodyOpened();
-    }
-    else
-    {
-        TraceLogger::GetInstance().LogMemoryBodyOpened();
-    }
+}
+
```
EriWong: remove empty event handler?

README.md
```diff
@@ -48,10 +48,10 @@ We also welcome [issues submitted on GitHub](https://github.com/Microsoft/calcul
 ## Roadmap
 For information regarding Windows Calculator plans and release schedule, please see the [Windows Calculator Roadmap](docs/Roadmap.md).
 
-## Data / Telemetry
+## Data / Diagnostics
```
mcooley: >Data / Diagnostics [](start = 3, length = 18)

nit: "Diagnostic Data" would be a better heading

README.md
```diff
@@ -48,10 +48,10 @@ We also welcome [issues submitted on GitHub](https://github.com/Microsoft/calcul
...
 This project collects usage data and sends it to Microsoft to help improve our products and services.
 Read our [privacy statement](https://go.microsoft.com/fwlink/?LinkId=521839) to learn more.
-Telemetry is disabled in development builds by default, and can be enabled with the `SEND_TELEMETRY`
+Diagnostics is disabled in development builds by default, and can be enabled with the `SEND_DIAGNOSTICS`
```
mcooley: >Diagnostics [](start = 0, length = 11)

"Diagnostic data is disabled..."

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +121,215 @@ namespace CalculatorApp
...
-            LogTelemetryEvent(EVENT_NAME_BITLENGTH_BUTTON_USED, fields);
-
-            bitLengthButtonLoggedInSession = true;
-        }
-    }
 
-    void TraceLogger::LogRadixButtonUsed(int windowId)
-    {
-        if (radixButtonUsage.find(windowId) == radixButtonUsage.end())
+        // Writer lock for the static resources
```
mcooley: >Writer lock for the static resources [](start = 11, length = 36)

Comment should probably be updated here and on line 106--windowIdLog and buttonLog are not "static resources".

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +121,215 @@ namespace CalculatorApp
...
-                fields.AddUInt32(L"WindowId", windowId);
-                LogTelemetryEvent(EVENT_NAME_FUNCTION_USAGE, fields);
-            }
-        }
-    }
-
-    void TraceLogger::LogHypButtonUsed(int windowId)
-    {
-        if (!isHypButtonLogged)
+        for (auto i : buttonLog)
```
mcooley: >buttonLog [](start = 22, length = 9)

You need to take a lock here to synchronize access to ButtonLog

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +121,215 @@ namespace CalculatorApp
...
-            LogTelemetryEvent(EVENT_NAME_HYP_BUTTON_USED, fields);
-
-            isHypButtonLogged = true;
+            fields.AddInt32(L"CalcMode", NavCategory::Serialize(i.mode));
+            fields.AddUInt32(L"ButtonId", i.buttonId);
+            fields.AddString(L"ButtonName", i.buttonName.data());
+            fields.AddUInt32(L"ViewModeId", NavCategory::Serialize(i.mode));
+            fields.AddUInt32(L"UsageCount", i.count);
+            fields.AddUInt64(PDT_PRIVACY_DATA_TAG, PDT_PRODUCT_AND_SERVICE_USAGE);
+            LogLevel2Event(EVENT_NAME_BUTTON_USAGE, fields);
```
mcooley: >LogLevel2Event(EVENT_NAME_BUTTON_USAGE, fields); [](start = 12, length = 48)

It would be better to log a single event with many fields (B5_Count, B5_Mode, etc.) rather than writing dozens of events in a loop. That way we get atomic behavior--either we log usage for all the buttons or none at all.

src/Calculator/Common/AppLifecycleLogger.cpp
```diff
@@ -35,7 +35,7 @@ namespace CalculatorApp
     AppLifecycleLogger::AppLifecycleLogger()
         : m_appLifecycleProvider(
             L"Microsoft.Windows.AppLifeCycle",
-            LoggingChannelOptions(GUID{ 0x4f50731a, 0x89cf, 0x4782, 0xb3, 0xe0, 0xdc, 0xe8, 0xc9, 0x4, 0x76, 0xba }), // Microsoft Telemetry group
+            LoggingChannelOptions(GUID{ 0x4f50731a, 0x89cf, 0x4782, 0xb3, 0xe0, 0xdc, 0xe8, 0xc9, 0x4, 0x76, 0xba }), // Microsoft Diagnostics group
```
mcooley: >Microsoft Diagnostics group [](start = 121, length = 27)

I appreciate why you've made this change elsewhere, but in this particular case the new comment really isn't correct. I think we should revert this one. (Same feedback applies in TraceLogger.cpp)

src/Calculator/Views/DateCalculator.xaml.h
```diff
@@ -50,5 +50,6 @@ namespace CalculatorApp
         void RaiseLiveRegionChangedAutomationEvent(_In_ bool isDateDiffMode);
 
         Windows::Foundation::EventRegistrationToken m_dateCalcOptionChangedEventToken;
+        void OnVisualStateChanged(Platform::Object ^ sender, Windows::UI::Xaml::VisualStateChangedEventArgs ^ e);
```
mcooley: > void OnVisualStateChanged(Platform::Object ^ sender, Windows::UI::Xaml::VisualStateChangedEventArgs ^ e); [](start = 7, length = 106)

nit: this method declaration should go with the others, above the fields

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +121,215 @@ namespace CalculatorApp
...
 
-        if (NavCategory::IsValidViewMode(fromMode) && NavCategory::IsValidViewMode(toMode))
+        if (NavCategory::IsValidViewMode(mode))
         {
             LoggingFields fields{};
-            fields.AddString(L"FromMode", NavCategory::GetFriendlyName(fromMode)->Data());
-            fields.AddString(L"ToMode", NavCategory::GetFriendlyName(toMode)->Data());
-            fields.AddInt32(L"WindowId", windowId);
-            LogMeasureEvent(EVENT_NAME_MODE_CHANGE_BEGIN, fields);
+            // cast mode to an int for diagnostics
```
mcooley: >// cast mode to an int for diagnostics [](start = 12, length = 38)

remove this comment (is it correct)?

src/CalcViewModel/Common/TraceLogger.h
```diff
@@ -13,19 +13,19 @@ static const int maxFunctionSize = (int)CalculationManager::Command::CommandBINE
...
 {
-    struct FuncLog
+    struct ButtonLog
     {
     public:
         int count;
-        std::wstring funcName;
-        FuncLog()
+        int buttonId;
+        std::wstring buttonName;
```
mcooley: buttonId and buttonName are redundant pieces of information. Rather than storing both of them in memory, let's just get the buttonName string at the time we actually need it, in LogButtonUsage.

Then you could define an equality operator in this class to simplify the find_if in UpdateButtonUsage.

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +121,215 @@ namespace CalculatorApp
...
-            radixButtonUsage[windowId]++;
-        }
-        if ((radixButtonUsage[windowId] == 2) && !radixButtonLoggedInSession)
-        {
-            LoggingFields fields{};
-            fields.AddUInt32(L"WindowId", windowId);
-            LogTelemetryEvent(EVENT_NAME_RADIX_BUTTON_USED, fields);
-
-            radixButtonLoggedInSession = true;
+            NumbersAndOperatorsEnum button = safe_cast<NumbersAndOperatorsEnum>(buttonId);
```
mcooley: >safe_cast<NumbersAndOperatorsEnum>(buttonId); [](start = 45, length = 45)

safe_cast is a CX-specific weird thing--in this case you want static_cast

Also, can we just use NumbersAndOperatorsEnum everywhere until the point in LogButtonUsage where we actually need to stick this in a tracelogging field

src/Calculator/Common/AppLifecycleLogger.cpp
sanderl: I was asked to remove all instance of "telemetry" to be consistent with the rest of Microsoft use of diagnostics. Should I just remove the comment?

src/CalcViewModel/Common/TraceLogger.cpp
sanderl: Right now, we are only logging an event for each unique button and mode combination used in the session. Are you suggesting that we dynamically add fields to log all of the buttons in 1 event?

mcooley: Yes, exactly that.

src/Calculator/Common/AppLifecycleLogger.cpp
mcooley: Removing the comment would be OK.

abhiram2466:
Can you please tell how are you implementing "ButtonUsageInSession" Event?

src/CalcViewModel/Common/TraceLogger.cpp
sanderl: I updated the LogButtonUsage method to write the ButtonLog array into a string that uses | and , to separate the fields and button info. So it will write a string like "Mode|Button1|Count,Mode|Button2|count, ..."

sanderl:
> Can you please tell how are you implementing "ButtonUsageInSession" Event?

ButtonUsageInSession event will be fired periodically throughout the app. It will send the whole ButtonLog array as a string once the array reaches a size of 10. It also sends the array when the app suspends. Originally I was just sending all of the data on suspend, however there was a potential issue where the data would not be sent on a crash or if the process was killed. Sending the button usage data in batches, mitigates this issue so that even if there is data lost it will only be the last 10 button/mode combinations.

src/CalcViewModel/Common/CalculatorButtonUser.h
```diff
@@ -161,55 +161,14 @@ public
...
-        Sinh = (int)CM::Command::CommandSINH,
-        Cosh = (int)CM::Command::CommandCOSH,
-        Tanh = (int)CM::Command::CommandTANH,
-        InvSinh = (int)CM::Command::CommandASINH,
-        InvCosh = (int)CM::Command::CommandACOSH,
-        InvTanh = (int)CM::Command::CommandATANH,
-        Cube = (int)CM::Command::CommandCUB,
-        DMS = (int)CM::Command::CommandDMS,
+        // Enum values below are used for Tracelogging and do not map to the Calculator engine
+        MemoryAdd,
```
mcooley: We should set these to well-known integers--otherwise someone might accidentally change them in the future.

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +122,242 @@ namespace CalculatorApp
...
+        fields.AddString(L"CalcMode", NavCategory::GetFriendlyName(mode)->Data());
+        fields.AddInt32(L"MemoryListSize", memoryListSize);
+        fields.AddInt32(L"MemoryItemIndex", loadedIndex);
         fields.AddUInt64(PDT_PRIVACY_DATA_TAG, PDT_PRODUCT_AND_SERVICE_USAGE);
-        LogTelemetryEvent(EVENT_NAME_INVALID_PASTED_INPUT_OCCURRED, fields);
+        LogLevel2Event(EVENT_NAME_MEMORY_ITEM_LOAD, fields);
     }
-
-    void TraceLogger::LogValidInputPasted(ViewMode mode) const
+    void TraceLogger::LogError(ViewMode mode, wstring_view functionName, wstring_view errorString)
```
mcooley: nit: space above

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -236,782 +122,242 @@ namespace CalculatorApp
...
+        fields.AddGuid(L"SessionGuid", sessionGuid);
+        fields.AddString(L"CalcMode", NavCategory::GetFriendlyName(mode)->Data());
+        fields.AddString(L"FunctionName", functionName);
+        fields.AddString(L"Message", errorString);
+        fields.AddUInt64(PDT_PRIVACY_DATA_TAG, PDT_PRODUCT_AND_SERVICE_USAGE);
+        LogLevel2Event(EVENT_NAME_EXCEPTION, fields);
     }
-
-    void TraceLogger::LogStandardException(wstring_view functionName, const exception& e) const
+    void TraceLogger::LogStandardException(ViewMode mode, wstring_view functionName, const exception& e) const
```
mcooley: nit: space above

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -299,7 +301,7 @@ namespace CalculatorApp
 
     void TraceLogger::LogButtonUsage()
     {
-        if (!GetTraceLoggingProviderEnabled())
+        if (!GetTraceLoggingProviderEnabled() || buttonLog.size() == 0)
```
mcooley: This check needs to come after you've taken the lock.

PR #576, opened 2019-07-06
Removed need for trimIdx and fDigitsFound, resulting in less code.

pi1024e:
Having an extra integer variable to hold onto an i or i-1 value is inefficient, for "if (m_token.at(i) != chZero)", the code under "fDigitsFound" will always run, so it makes sense to put the fDigitsFound code under the if statement, which can return from the function entirely instead of breaking.

- Manual Testing verified the code having identical behavior as before, with no side effects.


src/CalcManager/ExpressionCommand.cpp
```diff
@@ -256,38 +256,30 @@ const wstring& COpndCommand::GetToken(wchar_t decimalSymbol)
...
 
     // Remove zeros
-    bool fDigitsFound = false;
-    int trimIdx = 0;
     for (unsigned int i = 0; i < m_token.size(); i++)
     {
         if (m_token.at(i) != chZero)
         {
-            if (m_token.at(i) == decimalSymbol)
+            if (m_token.at(i) == decimalSymbol)
```
HowardWolosky: nit: you added a trailing space

```suggestion
            if (m_token.at(i) == decimalSymbol)
```

src/CalcManager/ExpressionCommand.cpp
```diff
@@ -256,38 +256,30 @@ const wstring& COpndCommand::GetToken(wchar_t decimalSymbol)
...
         }
     }
-    else
-    {
-        m_token.clear();
-        m_token.append(&chZero);
-    }
+     m_token.clear();
+     m_token.append(&chZero);
+
```
HowardWolosky: nit: you added an extra, unnecessary line

```suggestion
```

src/CalcManager/ExpressionCommand.cpp
```diff
@@ -256,38 +256,30 @@ const wstring& COpndCommand::GetToken(wchar_t decimalSymbol)
...
+            }
+            return m_token;
         }
     }
-    else
-    {
-        m_token.clear();
-        m_token.append(&chZero);
-    }
+     m_token.clear();
```
HowardWolosky: nit: the indentation for these lines is off after the refactor

```suggestion
    m_token.clear();
```

src/CalcManager/ExpressionCommand.cpp
```diff
@@ -256,38 +256,30 @@ const wstring& COpndCommand::GetToken(wchar_t decimalSymbol)
...
+     m_token.append(&chZero);
```
HowardWolosky: nit: the indentation for these lines is off after the refactor

```suggestion
    m_token.append(&chZero);
```

pi1024e:
Made those formatting changes. Thank you!

HowardWolosky:
Thanks @pi1024e.  This update brought in a bunch of additional changes across ten more files, unrelated to the topic of the initial PR.  Was that intentional?

pi1024e:
That was unintentional, but this actually applies the same types of code reduction changes I made in the first revision. If you like I can resend the pull request with only the first file changed.

pi1024e:
All formatting changes applied!

pi1024e:
Is this OK or should I only commit some of the changed files as per the original PR? Please let me know. Thank you!

HowardWolosky:
>Is this OK or should I only commit some of the changed files as per the original PR? Please let me know. Thank you!

Let's stick with the content of your original PR.  You're welcome to submit a separate PR for the other changes, and those will be evaluated/reviewed separately.  Thanks!

pi1024e:
All requested changes are done! This should be good to go!

HowardWolosky:
You still have whitespace changes in two unrelated files which you need to revert first. Thanks.

pi1024e:
Fixed!

PR #579, opened 2019-07-10
Always-on-Top mode implemented

### LINKED ISSUE #78: Add the ability for users to pin Calculator on top of other windows
**Problem Statement**
When attempting to use Calculator with other application, such as a spreadsheet, Calculator will lose focus when you click on the other app (e.g. to scroll/copy). Users have requested the ability to pin Calculator on top of other windows to facilitate ongoing use with other apps. 

**Evidence or User Insights**
This is our second-highest customer feedback request in Feedback Hub with some users reporting their only alternative is to pay for an app which is sticky.

**Proposal**
Add the ability for users to pin Calculator on top of other windows. 

**Goals**
- Users can easily pin/unpin the Calculator window on top
- Users can easily continue with other tasks while Calculator is pinned
- Users have access to full Calculator functionality, but can always perform at least basic calculations when in the more compact view (smaller than currently allowed minimum window sizing) 

**Non-Goals**
- Parity with the full version of the app​​ when in smallest of layout states

**Low-Fidelity Concept**
Here is a quick concept showing what the smallest of modes might look like. There are still several open questions to address for this layout state as well as determining the best way to enter/exit this mode:

image

greedyAI:
Implemented the 2nd most user requested new feature: the ability for users to pin Calculator on top of other windows.

- A `Button` has been added to `TitleBar` and styled identical to the other `TitleBar` buttons. This button only appears in Standard calculator mode. Clicking on this button calls `ViewModePreferences` API to switch the window into `CompactOverlay` mode (and back, when exiting from Always-on-Top mode).
- The `ViewModel` is updated accordingly when the button is pressed. These updates propagate dependency properties and observable variables back to the UI to show/hide and enable/disable certain UI elements (eg. buttons) and keyboard shortcuts.
- The screen size is stored in `LocalSettings` whenever the user resizes the screen in Always-on-Top mode. This is used for restoring the last used screen size when the user enters Always-on-Top mode.
- `AdaptiveTriggers` and `SizedChanged` event handlers are used to implement a responsive UI: certain buttons are hidden when the screen shrinks and the font sizes vary based on the screen size as well.
- The main expression in the Always-on-Top mode combines an `ItemsControl` and a `TextBlock` to render the desired mathematical result/expression. This scrollable expression shows the full mathematical expression when edited. When the user presses the "=" button, it shows the final computed result.
- The resulting functionality matches that as specified in the PM spec [here](https://github.com/microsoft/calculator-specs/blob/master/active/Always-on-top/README.md).

- Manual testing.

Fixes #78


src/Calculator/Controls/OverflowTextBlock.cpp
```diff
@@ -32,6 +32,7 @@ void OverflowTextBlock::OnApplyTemplate()
 {
     UnregisterEventHandlers();
 
+    // NEED TO ADD ANOTHER MEMBER VARIABLE TO KEEP TRACK OF MAIN RESULT SO THAT WE CAN USE ITS WIDTH IN SCROLLING CALCULATIONS
```
sanderl: >// NEED TO ADD ANOTHER MEMBER VARIABLE TO KEEP TRACK OF MAIN RESULT SO THAT WE CAN USE ITS WIDTH IN SCROLLING CALCULATIONS [](start = 4, length = 122)

Do you still need to do this? If not, remove the comment.

src/Calculator/Views/Calculator.xaml
```diff
@@ -94,6 +114,79 @@
...
+                            </ScrollViewer>
+                            <Button x:Name="ScrollLeft"
+                                    x:Uid="scrollLeft"
+                                    Grid.Column="0"
+                                    Margin="-4,3,-4,0"
+                                    Style="{StaticResource ScrollButtonStyle}"
+                                    HorizontalContentAlignment="Center">
+                                <FontIcon FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                                          FontSize="28"
+                                          Glyph="&#xE26C;"/>
```
sanderl: Does this not exist in the CalcMDL2.ttf?

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -292,12 +314,25 @@ void StandardCalculatorViewModel::SetExpressionDisplay(
     m_commands = commands;
     if (!IsEditingEnabled)
     {
+        unsigned int nTokens = 0;
+        tokens->GetSize(&nTokens);
+        DoesAlwaysOnTopResultShowTokens = !IsInError;
+        if (IsEngineRecording || nTokens == 0 || IsInError)
+        {
+            DoesAlwaysOnTopResultConcatenate = true;
+            CalculationAlwaysOnTopResultAutomationName = GetCalculatorExpressionAutomationName() + " " + CalculationResultAutomationName;
```
joshkoon: Does this read out correctly when using a right-to-left language?

src/Calculator/Calculator.vcxproj
```diff
@@ -221,6 +221,11 @@
   </ItemDefinitionGroup>
   <PropertyGroup>
     <AppVersion Condition="'$(AppVersion)' == ''">0.0.0.0</AppVersion>
+    <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
+    <AppxAutoIncrementPackageRevision>True</AppxAutoIncrementPackageRevision>
+    <AppxBundlePlatforms>x86|x64|arm</AppxBundlePlatforms>
+    <AppInstallerUpdateFrequency>1</AppInstallerUpdateFrequency>
+    <AppInstallerCheckForUpdateFrequency>OnApplicationRun</AppInstallerCheckForUpdateFrequency>
```
joshkoon: Remove these - they're artifacts from building the app locally and shouldn't be checked in.

src/Calculator/Resources/en-US/CEngineStrings.resw
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
joshkoon: What changed here? Can you please revert if there is no content change?

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -128,6 +136,7 @@ void MainPage::OnAppPropertyChanged(_In_ Platform::Object ^ sender, _In_ Windows
         if (newValue == ViewMode::Standard)
         {
             EnsureCalculator();
+            EnsureAlwaysOnTopCalculator();
```
joshkoon: We shouldn't need to call this every time we call `EnsureCalculator()`. We should only load Always On Top when that button is clicked. Otherwise, we're doing a lot of extra work for no reason.

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -167,4 +175,88 @@ namespace CalculatorApp
...
+            auto changeModeTask = Concurrency::create_task(result);
+            changeModeTask.then([this](bool success) {
+                auto avm = safe_cast<CalculatorApp::ViewModel::ApplicationViewModel ^>(this->DataContext);
+                avm->CalculatorViewModel->IsAlwaysOnTop = success;
+                if (success)
+                {
+                    AlwaysOnTopButton->HorizontalAlignment = ::HorizontalAlignment::Left;
+                    AlwaysOnTopButton->Content = L"\uEE47";
+                    AppName->Text = "";
+                    ToolTipService::SetToolTip(AlwaysOnTopButton, "Exit always-on-top");
```
joshkoon: Strings need to be checked into the .resw file so they can be localized.

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -167,4 +175,88 @@ namespace CalculatorApp
...
+                }
+                else
+                {
+                    AlwaysOnTopButton->HorizontalAlignment = ::HorizontalAlignment::Right;
+                    AlwaysOnTopButton->Content = L"\uEE49";
+                    AppName->Text = AppResourceProvider::GetInstance().GetResourceString(L"AppName");
+                    ToolTipService::SetToolTip(AlwaysOnTopButton, "Always on top");
```
joshkoon: Same here - please add strings to the .resw file for localization

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -209,6 +212,25 @@ void StandardCalculatorViewModel::SetPrimaryDisplay(_In_ wstring const& displayS
...
+    AreAlwaysOnTopResultsUpdated = false;
+    if (!IsEditingEnabled && m_tokens != nullptr)
+    {
+        unsigned int nTokens = 0;
+        m_tokens->GetSize(&nTokens);
+        DoesAlwaysOnTopResultShowTokens = !IsInError;
+        if (IsEngineRecording || nTokens == 0 || IsInError)
+        {
+            DoesAlwaysOnTopResultConcatenate = true;
+            CalculationAlwaysOnTopResultAutomationName = GetCalculatorExpressionAutomationName() + " " + CalculationResultAutomationName;
```
joshkoon: Similar to my other comment - concatenating strings usually breaks right-to-left languages. To achieve this result, we should create a localizable formatting string in the .resw file.

src/Calculator/Views/Calculator.xaml.h
```diff
@@ -100,6 +103,18 @@ public
...
+        bool ClearMemoryButtonErrorPrevState;
+        bool MemRecallErrorPrevState;
+        bool MemPlusErrorPrevState;
+        bool MemMinusErrorPrevState;
+        bool MemButtonErrorPrevState; 
+        bool ClearMemoryButtonEnabled = true;
+        bool MemRecallEnabled = true;
+        bool MemPlusEnabled = true;
+        bool MemMinusEnabled = true;
+        bool MemButtonEnabled = true;
```
joshkoon: A couple comments here:
For consistency, private member variables should be prefixed with `m_`, and preferably begin with lowercase letters, similar to the flyout automation name variables below. I realize we haven't been perfectly consistent with this (such as the m_IsLastFlyout* variables above), but we should try to move towards consistency when possible.

Second, this change is adding 10 member variables just to track the enabled/disabled state of the various memory buttons for always-on-top mode. Can we find a way to reduce this or remove them altogether?

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -309,6 +330,32 @@ void MainPage::EnsureCalculator()
     }
 }
 
+void MainPage::EnsureAlwaysOnTopCalculator()
+{
```
joshkoon: Why are you creating a separate calculator just for always-on-top mode? This creates a lot of additional (and duplicate) elements in the UI tree, which has a high potential for slowing down the app startup time and increasing the memory usage.

Can we examine if there's a way we can implement always-on-top using the existing UI tree and managing visual states, instead?

src/Calculator/Views/TitleBar.xaml
```diff
@@ -3,28 +3,53 @@
...
+                ClickMode="Release"
+                Width="46"
+                Height="32"
+                HorizontalAlignment="Right"
+                HorizontalContentAlignment="Center"
+                Background="Transparent"
+                FontWeight="Thin"
+                FontSize="14"
+                Visibility="{x:Bind Model.IsNormalMode, Converter={StaticResource BooleanToVisibilityConverter}, Mode=OneWay}"
+                ToolTipService.ToolTip="Always on top"/>
```
joshkoon: Tooltips should be localized using strings in our resource files

src/Calculator/Controls/OverflowTextBlock.cpp
greedyAI: Resolved.

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -79,7 +86,8 @@ namespace CalculatorApp
 
     void TitleBar::SetTitleBarVisibility()
     {
-        this->LayoutRoot->Visibility = m_coreTitleBar->IsVisible && !m_accessibilitySettings->HighContrast ? ::Visibility::Visible : ::Visibility::Collapsed;
+        // this->BackgroundElement->Visibility = m_coreTitleBar->IsVisible && !m_accessibilitySettings->HighContrast ? ::Visibility::Visible : ::Visibility::Collapsed;
+        this->LayoutRoot->Visibility = m_coreTitleBar->IsVisible ? ::Visibility::Visible : ::Visibility::Collapsed;
```
rudyhuyn: Warning, this diff will change how the title bar looks when high-contrast is on (different background color), is it intended?

src/Calculator/App.xaml
```diff
@@ -131,6 +131,27 @@
             <!-- Standard Operators Standard/Scientific in Fill/Full -->
             <x:Double x:Key="CalcOperatorCaptionSize">15</x:Double>
 
+            <Style x:Key="AlwaysOnTopButtonStyle" TargetType="Button">
```
rudyhuyn: Never used?

src/Calculator/Controls/OverflowTextBlock.cpp
```diff
@@ -64,6 +64,12 @@ void OverflowTextBlock::OnApplyTemplate()
         m_itemsControl = safe_cast<ItemsControl ^>(uiElement);
     }
 
+    uiElement = GetTemplateChild("EditableToken");
```
rudyhuyn: Wrong PR?

src/Calculator/Views/Calculator.xaml.cpp
```diff
@@ -60,6 +61,8 @@ Calculator::Calculator()
     auto resLoader = AppResourceProvider::GetInstance();
     CopyMenuItem->Text = resLoader.GetResourceString(L"copyMenuItem");
     PasteMenuItem->Text = resLoader.GetResourceString(L"pasteMenuItem");
+
+    this->SizeChanged += ref new SizeChangedEventHandler(this, &Calculator::UpdateOverflowTextBlock);
```
rudyhuyn: Use a more explicit function name: `Calculator_SizeChanged`?

src/Calculator/Views/TitleBar.xaml
```diff
@@ -3,28 +3,53 @@
...
                    TextAlignment="Left"
-                   TextTrimming="CharacterEllipsis"/>
+                   TextTrimming="CharacterEllipsis"
+                   Visibility="{x:Bind Model.CalculatorViewModel.IsAlwaysOnTop, Converter={StaticResource BooleanToVisibilityNegationConverter}, Mode=OneWay}"/>
+        </Grid>
+        <Button x:Name="AlwaysOnTopButton" Grid.Row="0"
+                FontFamily="{StaticResource CalculatorFontFamily}"
+                Style="{ThemeResource CommandBarFlyoutEllipsisButtonStyle}"
+                Content="&#xEE49;"
+                Click="AlwaysOnTopButtonClick"
```
rudyhuyn: NIT: `AlwaysOnTopButtonClicked` or `AlwaysOnTopButton_Click`

It looks like the first format is more used in the app.

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -167,4 +175,88 @@ namespace CalculatorApp
...
+        if (ApplicationView::GetForCurrentView()->GetForCurrentView()->ViewMode == ApplicationViewMode::CompactOverlay)
+        {
+            avm->Mode = m_beforeAlwaysOnTopMode;
+            if (avm->Mode == CalculatorApp::Common::ViewMode::Standard)
+            {
+                avm->CalculatorViewModel->AreHistoryShortcutsEnabled = true;
+                avm->CalculatorViewModel->HistoryVM->AreHistoryShortcutsEnabled = true;
+            }
+            IAsyncOperation<bool> ^ result = ApplicationView::GetForCurrentView()->TryEnterViewModeAsync(ApplicationViewMode::Default);
+            auto changeModeTask = Concurrency::create_task(result);
```
rudyhuyn: you can simplify your code and use `co_await` instead

src/Calculator/Views/TitleBar.xaml.cpp
```diff
@@ -167,4 +175,88 @@ namespace CalculatorApp
...
+        else
+        {
+            m_beforeAlwaysOnTopMode = avm->Mode;
+            avm->Mode = CalculatorApp::Common::ViewMode::Standard;
+            avm->CalculatorViewModel->AreHistoryShortcutsEnabled = false;
+            avm->CalculatorViewModel->HistoryVM->AreHistoryShortcutsEnabled = false;
+
+            Windows::Storage::ApplicationDataContainer ^ localSettings = Windows::Storage::ApplicationData::Current->LocalSettings;
+            ViewModePreferences ^ compactOptions = ViewModePreferences::CreateDefault(ApplicationViewMode::CompactOverlay);
+            if (!localSettings->Values->GetView()->HasKey("calculatorAlwaysOnTopLaunched"))
```
rudyhuyn: You should probably create some constant strings for all the keys you use ("calculatorAlwaysOnTopLaunched", "calculatorAlwaysOnTopLastWidth", ...)

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -200,3 +202,57 @@ void ApplicationViewModel::SetMenuCategories()
...
+void ApplicationViewModel::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    HandleAlwaysOnTopButtonClick(sender, e);
+}
+
+Concurrency::task<void> ApplicationViewModel::HandleAlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    if (ApplicationView::GetForCurrentView()->GetForCurrentView()->ViewMode == ApplicationViewMode::CompactOverlay)
+    {
+        CalculatorViewModel->AreHistoryShortcutsEnabled = true;
```
joseartrivera: If TryEnterViewModeAsync returns false, we still disable these shortcuts, which doesn't seem like correct behavior.

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -200,3 +202,57 @@ void ApplicationViewModel::SetMenuCategories()
...
+        CalculatorViewModel->HistoryVM->AreHistoryShortcutsEnabled = true;
+
+        bool success = co_await ApplicationView::GetForCurrentView()->TryEnterViewModeAsync(ApplicationViewMode::Default);
+        CalculatorViewModel->IsAlwaysOnTop = !success;
+        IsAlwaysOnTop = !success;
+    }
+    else
+    {
+        CalculatorViewModel->AreHistoryShortcutsEnabled = false;
```
joseartrivera: Same comment as above.

src/CalcViewModel/ApplicationViewModel.cpp
joseartrivera: Could you do something similar to this property for memory shortcuts instead of trying to keep them disabled in AoT mode?

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -260,6 +271,15 @@ void MainPage::OnPageLoaded(_In_ Object ^, _In_ RoutedEventArgs ^ args)
         }));
 }
 
+void MainPage::OnHighContrastChanged(_In_ AccessibilitySettings ^ /*sender*/, _In_ Object ^ /*args*/)
+{
+    if (Model->IsAlwaysOnTop && this->ActualHeight < 394)
+    {
+        // Sets to default always-on-top size to force re-layout
+        ApplicationView::GetForCurrentView()->TryResizeView(Size(320, 394));
```
joseartrivera: Why is this needed?

src/Calculator/Views/TitleBar.xaml
```diff
@@ -3,28 +3,54 @@
...
+                Height="32"
+                HorizontalAlignment="Left"
+                HorizontalContentAlignment="Center"
+                Background="Transparent"
+                FontWeight="Thin"
+                FontSize="14"
+                AllowFocusWhenDisabled="False"
+                Visibility="{x:Bind ApplicationViewModel.IsAlwaysOnTop, Mode=OneWay}"
+                AutomationProperties.AutomationId="AlwaysOnTopButton"
+                AutomationProperties.Name="Back to full view"/>
```
joseartrivera: This string should be localized.

src/Calculator/Views/Calculator.xaml
```diff
@@ -94,6 +114,111 @@
...
+                                <ColumnDefinition Width="Auto"/>
+                            </Grid.ColumnDefinitions>
+                            <ScrollViewer x:Name="ExpressionContainer"
+                                          Grid.Column="0"
+                                          Grid.ColumnSpan="3"
+                                          Padding="0,0,0,0"
+                                          VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
+                                          Style="{StaticResource ResultsScrollerSnapped}"
+                                          AutomationProperties.AccessibilityView="Raw">
+                                <StackPanel Orientation="Horizontal"
```
joseartrivera: I don't think you need this StackPanel since it only contains one item.

src/Calculator/Views/Calculator.xaml
```diff
@@ -94,6 +114,111 @@
...
+                                            Grid.Row="2"
+                                            HorizontalAlignment="Right">
+                                    <TextBlock x:Name="EditableToken"
```
joseartrivera: Since you are no longer showing the expression tokens you should enable the copy/paste menu. Also why not just use the CalculationResult control here directly?

src/Calculator/Package.appxmanifest
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10" xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest" xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10" xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5" IgnorableNamespaces="uap uap5 mp">
-    <Identity Name="Microsoft.WindowsCalculator.Dev" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.0.0.0" />
+    <Identity Name="Microsoft.WindowsCalculator.Dev" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.0.7.0" />
```
sanderl: Is there a reason the version is incremented? It should probably be set back to 0.0.0.0

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -1469,6 +1481,22 @@
     <value>Categories DropDown</value>
     <comment>Screen reader prompt for the Categories dropdown field.</comment>
   </data>
+  <data name="NormalAlwaysOnTopButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
```
sanderl: >NormalAlwaysOnTopButton [](start = 14, length = 23)

Can you keep the names consistent with the above resource strings? Maybe EnterAlwaysOnTopButton and ExitAlwaysOnTopButton?

src/Calculator/Views/Calculator.xaml
```diff
@@ -94,6 +114,110 @@
...
+            <Setter Property="ColumnWidth" Value="14"/>
+            <Setter Property="ColumnHeight" Value="14"/>
+            <Setter Property="ScrollButtonFontSize" Value="12"/>
+            <Setter Property="Template">
+                <Setter.Value>
+                    <ControlTemplate TargetType="controls:OverflowTextBlock">
+                        <Grid x:Name="TokenContainer" Background="{TemplateBinding Background}">
+                            <Grid.Resources>
+                                <ResourceDictionary>
+                                    <ResourceDictionary.ThemeDictionaries>
```
sanderl: >ResourceDictionary.ThemeDictionaries> [](start = 37, length = 37)

Do you need a dark theme here?

src/Calculator/Views/UnitConverter.xaml
```diff
@@ -712,7 +712,7 @@
                                        Style="{StaticResource SymbolOperatorButtonStyle}"
                                        FontSize="16"
                                        ButtonId="Negate"
-                                       Content="&#xe94d;"
+                                       Content="&#xF898;"
```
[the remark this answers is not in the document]
greedyAI: A new font file was created for this feature to add the two symbols for this mode. Part of the file was updated for other symbols (eg. square root) as well.

src/Calculator/Views/TitleBar.xaml.cpp
greedyAI: Correct.

src/Calculator/Controls/OverflowTextBlock.cpp
greedyAI: This is part of the styling for Always-on-Top.

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -392,13 +435,26 @@ String ^ StandardCalculatorViewModel::GetCalculatorExpressionAutomationName()
...
+}
+
 void StandardCalculatorViewModel::SetMemorizedNumbers(const vector<wstring>& newMemorizedNumbers)
 {
     const auto& localizer = LocalizationSettings::GetInstance();
     if (newMemorizedNumbers.size() == 0) // Memory has been cleared
     {
         MemorizedNumbers->Clear();
         IsMemoryEmpty = true;
+        IsMemoryEmptyAlwaysOnTop = IsMemoryEmpty || IsAlwaysOnTop;
```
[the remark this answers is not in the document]
greedyAI: This combines existing visibility logic for the MemoryButton with extra logic necessary for the Always-on-Top mode.

src/Calculator/Views/Calculator.xaml
```diff
@@ -365,7 +457,7 @@
                         <AdaptiveTrigger MinWindowHeight="0" MinWindowWidth="0"/>
                     </VisualState.StateTriggers>
                     <VisualState.Setters>
-                        <Setter Target="RowMemoryControls.MinHeight" Value="24"/>
+                        <Setter Target="RowMemoryControls.MinHeight" Value="0"/>
```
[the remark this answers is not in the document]
greedyAI: The app, when not in AoT, is limited by a MinWidth and MinHeight, which will force RowMemoryControls's MinHeight to be at least 24 anyways.

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -115,6 +115,14 @@ void MainPage::WindowSizeChanged(_In_ Platform::Object ^ /*sender*/, _In_ Window
 {
     // We don't use layout aware page's view states, we have our own
     UpdateViewState();
+    if (m_model->CalculatorViewModel->IsAlwaysOnTop)
+    {
+        Windows::Storage::ApplicationDataContainer ^ localSettings = Windows::Storage::ApplicationData::Current->LocalSettings;
+        // double width = safe_cast<float>(this->ActualWidth);
+        // double height = safe_cast<float>(this->ActualHeight);
+        localSettings->Values->Insert(App::WidthLocalSettings, this->ActualWidth);
+        localSettings->Values->Insert(App::HeightLocalSettings, this->ActualHeight);
```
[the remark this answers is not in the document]
greedyAI: Suspend may not always be called when the app exits. For example, it may be forcefully exited by the system.

src/Calculator/Views/MainPage.xaml.cpp
greedyAI: Resolved in person.

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -204,3 +206,55 @@ void ApplicationViewModel::SetMenuCategories()
...
 }
+
+void ApplicationViewModel::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    HandleAlwaysOnTopButtonClick(sender, e);
+}
+
+Concurrency::task<void> ApplicationViewModel::HandleAlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    if (ApplicationView::GetForCurrentView()->GetForCurrentView()->ViewMode == ApplicationViewMode::CompactOverlay)
```
joseartrivera: Do you need to call GetForCurrentView() twice here?

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -204,3 +206,55 @@ void ApplicationViewModel::SetMenuCategories()
     // property setter logic.
     Categories = NavCategoryGroup::CreateMenuOptions();
 }
+
+void ApplicationViewModel::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    HandleAlwaysOnTopButtonClick(sender, e);
+}
+
+Concurrency::task<void> ApplicationViewModel::HandleAlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
joseartrivera: Remove Concurrency::task<void> and replace with task<void> add the Concurrency namespace to the top of the file.

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -204,3 +206,55 @@ void ApplicationViewModel::SetMenuCategories()
...
+    {
+        bool success = co_await ApplicationView::GetForCurrentView()->TryEnterViewModeAsync(ApplicationViewMode::Default);
+        CalculatorViewModel->AreHistoryShortcutsEnabled = success;
+        CalculatorViewModel->HistoryVM->AreHistoryShortcutsEnabled = success;
+        CalculatorViewModel->IsAlwaysOnTop = !success;
+        IsAlwaysOnTop = !success;
+    }
+    else
+    {
+        Windows::Storage::ApplicationDataContainer ^ localSettings = Windows::Storage::ApplicationData::Current->LocalSettings;
```
joseartrivera: Same comment here and throughout all of the files. in .cpp files the style is to add using namespace Windows::Storage; at the top instead of doing this. Header files should have the fully qualified name though.

src/Calculator/Views/CalculatorStandardOperators.xaml.cpp
```diff
@@ -21,6 +21,7 @@ using namespace Windows::UI::Xaml::Data;
 using namespace Windows::UI::Xaml::Input;
 using namespace Windows::UI::Xaml::Media;
 using namespace Windows::UI::Xaml::Navigation;
+using namespace Windows::UI::ViewManagement;
```
joseartrivera: This can be removed right?

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -115,6 +116,12 @@ void MainPage::WindowSizeChanged(_In_ Platform::Object ^ /*sender*/, _In_ Window
 {
     // We don't use layout aware page's view states, we have our own
     UpdateViewState();
+    if (m_model->IsAlwaysOnTop)
+    {
+        Windows::Storage::ApplicationDataContainer ^ localSettings = Windows::Storage::ApplicationData::Current->LocalSettings;
```
joseartrivera: This should go in OnSuspend

src/Calculator/Views/TitleBar.xaml.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "Views/TitleBar.g.h"
+#include <CalcViewModel\ApplicationViewModel.h>
```
joseartrivera: Change to #include "CalcViewModel\ApplicationViewModel.h" for consistency

src/CalcViewModel/StandardCalculatorViewModel.cpp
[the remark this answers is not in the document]
joseartrivera: Where is it used? I only see it being set, never used anywhere?

src/Calculator/Views/MainPage.xaml.cpp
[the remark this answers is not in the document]
joseartrivera: Most apps save their state during suspend. You can also read docs here:
https://docs.microsoft.com/en-us/windows/uwp/launch-resume/suspend-an-app#save-application-data-before-suspension

Unless there is something special about this setting, I think suspend should cover most cases we need. If the system forcefully causes the app to shutdown then I think I would rather  lose the setting data rather than degrading performance on resize. @mcooley thoughts on this?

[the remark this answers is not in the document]
joseartrivera: Most apps save their state during suspend. You can also read docs here:
https://docs.microsoft.com/en-us/windows/uwp/launch-resume/suspend-an-app#save-application-data-before-suspension

Unless there is something special about this setting, I think suspend should cover most cases we need. If the system forcefully causes the app to shutdown then I think I would rather  lose the setting data rather than degrading performance on resize. @mcooley thoughts on this?

src/Calculator/Views/MainPage.xaml.cpp
```diff
@@ -527,3 +545,18 @@ void MainPage::OnNavItemInvoked(MUXC::NavigationView ^ /*sender*/, _In_ MUXC::Na
 {
     NavView->IsPaneOpen = false;
 }
+
+void MainPage::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    Model->AlwaysOnTopButtonClick(sender, e, 0, 0);
```
joshkoon: You might consider having the `width` and `height` params be optional with 0 as the default value, simplifying the call.

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -204,3 +207,59 @@ void ApplicationViewModel::SetMenuCategories()
...
+{
+    HandleAlwaysOnTopButtonClick(sender, e, width, height);
+}
+
+task<void> ApplicationViewModel::HandleAlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e, float width, float height)
+{
+    if (ApplicationView::GetForCurrentView()->ViewMode == ApplicationViewMode::CompactOverlay)
+    {
+        ApplicationDataContainer ^ localSettings = ApplicationData::Current->LocalSettings;
+        localSettings->Values->Insert(WidthLocalSettings, width);
```
joshkoon: Unsure of the desired behavior here: if there are multiple calculator windows open, and you move them between AoT and normal modes, the windows will always each "remember" their normal window size.

However, the AoT size is being recorded and used as an app-wide setting, so when a window is transitioned to AoT mode, it will pick up whatever size was last recorded when a window exited AoT.

To see what I mean:

1. Launch two Calc windows.
2. Put window A into AoT and resize to the smallest size.
3. Put window B into AoT. Observe it still goes to the default size (okay).
4. Take window B out of Aot. It goes back to its original size (okay).
5. Take window A out of AoT, then put it back into AoT. It goes back to small size (okay)
6. Put window B into AoT. It goes to the small size, as well, since that was the size last "exited" from (wrong?)

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -399,7 +405,7 @@ void StandardCalculatorViewModel::SetMemorizedNumbers(const vector<wstring>& new
     if (newMemorizedNumbers.size() == 0) // Memory has been cleared
     {
         MemorizedNumbers->Clear();
-        IsMemoryEmpty = true;
+        IsMemoryEmpty = true || IsAlwaysOnTop;
```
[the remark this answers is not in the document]
joseartrivera: Good catch, I believe this is the intent, I think IsMemoryEmpty could have a better name though to make it clear. Maybe something like IsMemoryEnabled.

src/CalcViewModel/ApplicationViewModel.cpp
grochocki: FWIW, this is the behavior of Calculator in general (i.e., if you have multiple windows open, we'll restore the last closed window size next time you reopen the app). I agree that there is some behavior to figure out here worth a discussion, but I'd like to see if this becomes a real problem before complicating this logic anymore. @greedyAI, can you open up an issue after this PR is merged in to track that follow up discussion?

src/Calculator/Views/MainPage.xaml.cpp
greedyAI: WinRT does not allow optional parameters for public functions. See [here](https://social.msdn.microsoft.com/Forums/en-US/4800e079-fb2e-4488-a76d-64b3a04bda76/optional-parameters-for-winrt-component)

rudyhuyn: I'm always surprised by this limitation.

But you can write:
`void ApplicationViewModel::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
{
    HandleAlwaysOnTopButtonClick(sender, e, 0, 0);
}`

or 

`void ApplicationViewModel::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
{
    AlwaysOnTopButtonClick(sender, e, 0, 0);
}`

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -204,3 +207,59 @@ void ApplicationViewModel::SetMenuCategories()
     // property setter logic.
     Categories = NavCategoryGroup::CreateMenuOptions();
 }
+
+void ApplicationViewModel::AlwaysOnTopButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e, float width, float height)
```
rudyhuyn: I would suggest to better separate View and ViewModel:
- Don't pass `sender` or `RoutedEventArgs` (these 2 items are related to the View)
- Use a name describing what the method does and not what is the reason for the call. Something like `SwitchViewModel()` for example

src/CalcViewModel/ApplicationViewModel.h
```diff
@@ -64,6 +67,48 @@ namespace CalculatorApp
...
+                bool get()
+                {
+                    return m_isAlwaysOnTop;
+                }
+                void set(bool value)
+                {
+                    if (m_isAlwaysOnTop != value)
+                    {
+                        m_isAlwaysOnTop = value;
+                        RaisePropertyChanged(L"IsAlwaysOnTop");
```
rudyhuyn: NIT: `RaisePropertyChanged(nameof(IsAlwaysOnTop));`

This way PropertyChanged will still work even if you refactor/rename this property!

src/CalcViewModel/ApplicationViewModel.cpp
rudyhuyn: @joseartrivera, what's your opinion?

joseartrivera: Agreed, looks like the last commit updated the args, just needs a reasonable name. Something like EnterAlwaysOnTop().

PR #581, opened 2019-07-10
Add variable editing

joseartrivera:
As part of #338, this PR adds the UI and backend for editing variables that exist for each equation. Variables can be modified with a slider and textbox. Min/Max/Step can be modified by selecting the gear icon for each variable (temporary icon).

The entry point to the UI is temporary as that is still being finalized.

Manual testing


src/Calculator/App.xaml
```diff
@@ -17,6 +17,7 @@
                     <x:Double x:Key="HighContrastStrokeThickness">0</x:Double>
                     <Color x:Key="AltHighColor">#FF000000</Color>
                     <Color x:Key="ChromeMediumLowColor">#FF2B2B2B</Color>
+                    <Color x:Key="AppControlFocusedVariableForegroundText">#FFFFFFFF</Color>
```
sanderl: >AppControlFocusedVariableForegroundText [](start = 34, length = 39)

Does this need to be added to Dark and High Contrast ThemeDictionaries?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -39,6 +46,175 @@
...
+                                        <ResourceDictionary x:Key="HighContrast">
+                                            <SolidColorBrush x:Key="TextControlBackgroundFocused" Color="{ThemeResource SystemColorButtonFaceColor}"/>
+                                            <SolidColorBrush x:Key="TextControlBackgroundPointerOver" Color="{StaticResource SystemColorButtonFaceColor}"/>
+                                            <SolidColorBrush x:Key="TextControlForegroundFocused" Color="{ThemeResource SystemColorButtonTextColor}"/>
+                                            <x:Double x:Key="TextControlThemeMinWidth">32</x:Double>
+                                            <x:Double x:Key="TextControlThemeMinHeight">16</x:Double>
+                                        </ResourceDictionary>
+                                    </ResourceDictionary.ThemeDictionaries>
+                                </ResourceDictionary>
+                            </ListView.Resources>
```
sanderl: Any reason you don't want these to be reusable by other controls in the future?

src/CalcViewModel/Common/Utils.h
```diff
@@ -32,6 +32,15 @@
     }}\
     } private: t m_##n; public:
 
+#define OBSERVABLE_PROPERTY_RW_ALWAYS_NOTIFY(t, n)\
+    property t n {\
+    t get() { return m_##n; }\
+    void set(t value) {\
+    m_##n = value;\
+    RaisePropertyChanged(L#n);\
```
rudyhuyn: What are we trying to fix with it this macro? Why is it important to launch `PropertyChanged` when the value didn't change?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -49,3 +54,67 @@ void GraphingCalculator::ViewModel::set(GraphingCalculatorViewModel^ vm)
...
+{
+    GraphingControl->SetVariable(args.variableName, args.newValue);
+}
+
+
+void GraphingCalculator::SubmitTextbox(TextBox^ sender)
+{
+    auto variableViewModel = static_cast<VariableViewModel^>(sender->DataContext);
+
+    if (sender->Name == "Value")
```
rudyhuyn: to be sure to not break this function if the XAML is modified:

`if(sender == Value)`
{
...
}`

The names of these TextBox should probably be changed to remove some confusions and conflicts, I don't expect this->Value to be a TextBox for example: 
proposal: `ValueTextBox`, `MinTextBox`, `MaxTextBox`, `StepTextBox`

src/Calculator/App.xaml
joseartrivera: Actually should have been deleted, gone now.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
joseartrivera: These are actually not new resources. It is [lightweight styling](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/xaml-styles#lightweight-styling) to avoid completely retemplating these controls.

src/CalcViewModel/Common/Utils.h
joseartrivera: I added this in because it handles a lot of work that I would need to do manually otherwise when binding a TextBox Text property to this property. For example, if a user types invalid input, this will let me set the Value to the previous Value and still fire the event, causing the binding to update, resetting the double in the textbox to the previous value with correct formatting.

I could just manually set the TextBox text in SubmitTextbox instead, but this would require extra work to also format the double -> string conversion to remove trailing zeroes. I noticed you added a utility in one of your recent PRs that lets us remove trailing zeroes very easily from a string. I added a comment to go ahead and remove this work around and use your util function once we re-merge with master and pick up that change.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: Done!

PR #603, opened 2019-07-26
Fixed issue where UI Responsive events were not fired

### LINKED ISSUE #602: Missing UI_Responsive Event in Latest Calculator Build
**Describe the bug**

The logic for updating the WindowIdLog in the TraceLogger code was removed, so no windowIds are ever added to the map. This prevents the UIResponsive events from firing after the app launches because it checks if the current window is in the log before firing the event.
**Steps To Reproduce**
Used internal diagnostics tools to repro this issue.

**Expected behavior**
UIResponsive events are fired after the app lauches.

 - OS Build:
 - Architecture:
 - Application Version: 10.1906.25.0
 - Region:
 - Dev Version Installed:

**Additional context**

**Requested Assignment**

sanderl:
- Updated the LogWindowCreated method to take in the windowId and update the WindowIdLog
- Renamed UpdateWindowIdLog to IsWindowIdInLog so that it reflects what the method is actually doing.

- Manually using internal diagnostics tools to ensure the UI Responsive events are firing now.

### DIFF
--- a/src/CalcViewModel/Common/TraceLogger.cpp
+++ b/src/CalcViewModel/Common/TraceLogger.cpp
@@ -104,24 +104,17 @@ namespace CalculatorApp
 #pragma endregion
 
     // return true if windowId is logged once else return false
-    bool TraceLogger::UpdateWindowIdLog(int windowId)
+    bool TraceLogger::IsWindowIdInLog(int windowId)
     {
         // Writer lock for the windowIdLog resource
         reader_writer_lock::scoped_lock lock(s_traceLoggerLock);
 
-        if (windowIdLog.find(windowId) == windowIdLog.end())
+        if (find(windowIdLog.begin(), windowIdLog.end(), windowId) == windowIdLog.end())
         {
             return false;
         }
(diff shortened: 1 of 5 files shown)


src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -139,8 +132,14 @@ namespace CalculatorApp
         LogLevel2Event(EVENT_NAME_VISUAL_STATE_CHANGED, fields);
     }
 
-    void TraceLogger::LogWindowCreated(ViewMode mode) const
+    void TraceLogger::LogWindowCreated(ViewMode mode, int windowId)
     {
+        // store windowId in windowIdLog which says we have logged mode for the present windowId.
+        if (find(windowIdLog.begin(), windowIdLog.end(), windowId) == windowIdLog.end())
```
EriWong: >find(windowIdLog.begin(), windowIdLog.end(), windowId) == windowIdLog.end() [](start = 12, length = 75)

call IsWindowIdInLog here?

PR #619, opened 2019-08-06
Do not collapse Update Rates button in a transient state

### LINKED ISSUE #352: Focus unexpectedly shifts to “Clear Entry” button on activating “Update rates” link
**Repro Steps**
* Launch Calculator app.
* Navigate to "Open navigation" button and activate it.
* Navigate to list items and select Currency Converter.
* Navigate to update rates link and activate it.
* Observe the issue.

**Actual Result**
Focus unexpectedly shifts to “Clear Entry” button on activating “Update rates” link.

**Expected Result**
Focus should remain on “Update rate” link itself on activating “Update rates” link.

**Test Environment**
App Name: Calculator 
App Version: 10.1902.16.0

zherman:
In this change I am fixing an issue where focus moves from the "Update Rates" button to the "CE" button without being initiated by the user explicitly. The issue is that when "Update Rates" is pressed the button becomes collapsed. When the button becomes collapsed the XAML focus manager then migrates focus to the next logical thing (which happens to be the "CE" button). In this change I modified the "Update Rates" button to not collapse when being pressed. For completeness I added a storyboard animation to fade in everything that was binding to IsCurrencyLoadingVisible so that it matches the animations of the Timestamp. The animations are only used if the user settings enable animations.

Now that the "Update Rates" button is always visible there is a possibility that the user will click it numerous times before the first update actually completes. My change safeguards this scenario by adding a logical check in the button's click handler so that work is only done if we are not already updating the rates.

Manual testing


zherman:
Thanks @rudyhuyn for the quick feedback!

I think there might be a misunderstanding of my change. The "Update Rates" button never becomes collapsed/hidden/invisible. My change makes it such that the button is always visible. That is why I safeguarded against spamming the button quickly. The elements that I added the bindings to can't get focus and are not picked up by narrator, so there should be no issue with having them hidden during the reload.

Picture is after I press the button before it finishes updating the rate
calc-while-updating

With regards to the binding/storyboard combo, I didnt know that could race! Ill switch that over to a VisualState shared for the 2 elements.

rudyhuyn:
> Thanks @rudyhuyn for the quick feedback!
> 
> I think there might be a misunderstanding of my change. The "Update Rates" button never becomes collapsed/hidden/invisible. My change makes it such that the button is always visible. That is why I safeguarded against spamming the button quickly. The elements that I added the bindings to can't get focus and are not picked up by narrator, so there should be no issue with having them hidden during the reload.
> 
> Picture is after I press the button before it finishes updating the rate
> calc-while-updating
> 
> With regards to the binding/storyboard combo, I didnt know that could race! Ill switch that over to a VisualState shared for the 2 elements.

You are right! I missed this point. 👍

 A Visual State will be great

PR #623, opened 2019-08-06
Remove unused expression item template selector

rudyhuyn:
- Code cleaning:
  - remove unused `DataTemplate` and `ExpressionItemTemplateSelector` from `Calculator.xaml`
  - Replace incorrect `ThemeResource` references by `StaticResource` when they don't depend on the currently active theme.

- Manually

### DIFF
--- a/src/Calculator/Views/Calculator.xaml
+++ b/src/Calculator/Views/Calculator.xaml
@@ -34,26 +34,6 @@
                        Text="{x:Bind Token, Mode=OneWay}"/>
         </DataTemplate>
 
-        <DataTemplate x:Key="AlwaysOnTopOperand" x:DataType="common:DisplayExpressionToken">
-            <TextBlock Margin="2,0,0,0"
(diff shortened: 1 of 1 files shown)


rudyhuyn:
Related questions: any reasons why the result is displayed using an extra `OverflowTextBlock` control in the Always-On-Top mode instead of the existing `CalculationResult` control? (to support Always-On-Top, we only need to set MinFontSize and MaxFontSize and modify the font size of ScrollLeft and ScrollRight according to the control font size, it would simplify a lot the support IMO).

sanderl:
> Related questions: Why is the result displayed using an extra `OverflowTextBlock` control in the Always-On-Top mode instead of the existing `CalculationResult` control? (It should work if set MinFontSize and modify the font size of ScrollLeft and ScrollRight according to the control font size.)

The CalculationResult control has 2 lines of text to show whole calculation, not just the current input. For Always on Top there is no room for the smaller line above the current input, so the control was not used.

Issue #560, opened 2019-06-19
Develop a minimal repro for coroutine issues and remove workarounds


mcooley:
In [several](https://github.com/microsoft/calculator/pull/548#issuecomment-503312276) [places](https://github.com/microsoft/calculator/blob/9f01c8168b921a82ada80be24203de7729b1b6fa/src/CalcViewModel/Common/ConversionResultTaskHelper.cpp#L32) [in](https://github.com/microsoft/calculator/blob/fe30c7cabc1f30c239b5a22ff6d0f7fa0fba5cd1/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp#L186) Calculator, we disable optimizations to work around bugs in the compiler related to coroutines. Some of these workarounds might no longer be needed (since some bugs have been [fixed](https://developercommunity.visualstudio.com/content/problem/263131/co-await-optimization-bugs-in-visual-studio-157-x6.html)) while others might be hiding still-unresolved issues that need to be fixed in the compiler.

For every instance where we've disabled optimizations (starting with the [one](https://github.com/microsoft/calculator/pull/548#issuecomment-503312276) which seems to be new in VS2019) we should [produce a minimal repro case](https://docs.microsoft.com/en-us/cpp/overview/how-to-report-a-problem-with-the-visual-cpp-toolset?view=vs-2019) which we can send to the C++ team.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

kaina404:
嗯

PR #626, opened 2019-08-06
Fixes Always-on-Top crashes in release

greedyAI:
- Always-on-Top currently crashes in release due to a compiler optimization bug related to co_await's. I've fixed this by turning off compiler optimization for parts of the code that uses co_await's.

- Manual testing.

Related to Issue #560

### DIFF
--- a/src/CalcViewModel/ApplicationViewModel.cpp
+++ b/src/CalcViewModel/ApplicationViewModel.cpp
@@ -213,6 +213,7 @@ void ApplicationViewModel::ToggleAlwaysOnTop(float width, float height)
     HandleToggleAlwaysOnTop(width, height);
 }
 
+#pragma optimize("", off)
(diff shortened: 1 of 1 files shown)


src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -257,7 +258,8 @@ task<void> ApplicationViewModel::HandleToggleAlwaysOnTop(float width, float heig
         IsAlwaysOnTop = success;
     }
     SetDisplayNormalAlwaysOnTopOption();
-}
+};
```
HowardWolosky: No need for the `;`
```suggestion
}
```

joseartrivera:
Closing to restart build

src/CalcViewModel/ApplicationViewModel.cpp
joseartrivera: I believe this is needed for the compiler to recognize this as a function.

PR #628, opened 2019-08-07
Add copy button to history menu item context menu

### LINKED ISSUE #429: Copy values directly from history to clipboard using right-click
**Problem Statement**

The pain point is that in order to copy a previously-calculated number to the clipboard, you have to save the current number to memory, click on the history item, click copy on the display, then re-insert the item from memory. This is slow and tedious, and also erases the last item in the history as a side effect of restoring the value. Instead, I propose that items can be copied from the history directly using a similar interface as they are copied from the output from a normal calculation.

**Evidence or User Insights**

We should do this because it will allow users to find previous calculations and save them, or combine them together. macOS's paper-tape is essentially a text field which allows copying numbers directly to the clipboard. Also, the TI-83 series has the capability to add historical items to its own "clipboard" by setting a variable to it. This is similar to the memory function, but it allows at least 26 different variables to be used, vaguely similar to how a clipboard works.

**Proposal**

This solution will solve the problem by allowing right-clicks to be used to copy the value to the clipboard. This will reduce the number of clicks from seven to two.

**Goals**

The user will be able to copy any item from the history to their clipboard.

**Non-Goals**

We should not support pasting values into history because this would make calculator more like a spreadsheet.

**Low-Fidelity Concept**

image

zherman:
Added a Copy button to the context menu for history menu items located above the delete button in the menu. Copy only copies the result and not the entire content of the history item (equation and result).

Manual Testing

### DIFF
--- a/src/Calculator/Views/HistoryList.xaml
+++ b/src/Calculator/Views/HistoryList.xaml
@@ -50,8 +50,7 @@
             </muxc:SwipeItems>
 
             <MenuFlyout x:Name="HistoryContextMenu">
-                <MenuFlyoutItem x:Name="CopyHistoryMenuItem"
-                                x:Uid="CopyHistoryMenuItem"
+                <MenuFlyoutItem x:Uid="CopyHistoryMenuItem"
                                 Click="OnCopyMenuItemClicked"
                                 Icon="Copy"/>
                 <MenuFlyoutItem x:Uid="DeleteHistoryMenuItem"
--- a/src/Calculator/Views/HistoryList.xaml
+++ b/src/Calculator/Views/HistoryList.xaml
@@ -50,6 +50,10 @@
             </muxc:SwipeItems>
 
             <MenuFlyout x:Name="HistoryContextMenu">
+                <MenuFlyoutItem x:Name="CopyHistoryMenuItem"
+                                x:Uid="CopyHistoryMenuItem"
+                                Click="OnCopyMenuItemClicked"
+                                Icon="Copy"/>
                 <MenuFlyoutItem x:Uid="DeleteHistoryMenuItem"
                                 Click="OnDeleteMenuItemClicked"
                                 Icon="Delete"/>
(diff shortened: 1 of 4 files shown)


src/Calculator/Views/HistoryList.xaml
```diff
@@ -50,6 +50,10 @@
             </muxc:SwipeItems>
 
             <MenuFlyout x:Name="HistoryContextMenu">
+                <MenuFlyoutItem x:Name="CopyHistoryMenuItem"
```
HowardWolosky: No need for `x:Name` since we won't be programatically accessing this item with the change in localization.

src/Calculator/Views/HistoryList.xaml.cpp
```diff
@@ -37,6 +38,8 @@ HistoryList::HistoryList()
 {
     InitializeComponent();
 
+    CopyHistoryMenuItem->Text = AppResourceProvider::GetInstance().GetResourceString(L"copyMenuItem");
```
HowardWolosky: There are two reasons why we don't want to do this as written:

1. For the most part, the rest of the project avoids directly setting UI strings this way in favor of setting values directly on the relevant property within the resw file itself.  Here's how it's being done for the `Delete` menu item:

https://github.com/microsoft/calculator/blob/24d997dd00534938cb880722108a0f85ae4c3611/src/Calculator/Resources/en-US/Resources.resw#L1120-L1123

2. This implementation is re-using a localization string that was designed for a different context (in this case, `copyMenuItem` is a string that was designed for the context of copying a result from a unit conversion).  The general philosophy is to not reuse localized strings in different contexts, because you never know when the string will be updated to be better for one context, making the other context worse/wrong.  Instead, a new localized string should be introduced for each context, and intelligence in the tools can reuse the localization that has already happened if the new string is identical to a previously localized string.

So, instead, please remove this line, and instead create a _new_ localized string that sets the value on the menu directly via property access like above:

      <data name="CopyHistoryMenuItem.Text" xml:space="preserve">
        <value>Copy</value>
        <comment>Text string for the Calculator Copy option in the History list context menu</comment>
      </data>

zherman:
Thanks @HowardWolosky for the feedback. I just pushed an update addressing the localization.

zherman:
No worries about the delay. I’m still learning the github interface and didn’t realize right away that I could notify people when I address feedback.

PR #632, opened 2019-08-07
Fix regression bug with application title no longer grayed out.

### LINKED ISSUE #631: The app title is no longer grayed out when the application loses focus
**Describe the bug**
A regression seems to have been introduced, the behavior added with #442 doesn't work anymore.

**Steps To Reproduce**
- Open the calculator
- Focus another windows
- Observe that the title is no more gray

**Expected behavior**
- The title should be gray

**Screenshots**
image
Left: public release, Right: current dev version

 - OS Build: 10.0.18362.0
 - Architecture: X64
 - Application Version: 10.1906.53.0
 - Region: en-US
 - Dev Version Installed: True

rudyhuyn:
- Move back the VisualStateManager node to the root XAML element to fix visual states of the titlebar.

- Manually

### DIFF
--- a/src/Calculator/Views/TitleBar.xaml
+++ b/src/Calculator/Views/TitleBar.xaml
@@ -11,19 +11,19 @@
     <Grid x:Name="LayoutRoot"
           Height="32"
           HorizontalAlignment="Stretch">
+        <VisualStateManager.VisualStateGroups>
+            <VisualStateGroup x:Name="WindowFocusStates">
+                <VisualState x:Name="WindowFocused"/>
+                <VisualState x:Name="WindowNotFocused">
+                    <VisualState.Setters>
+                        <Setter Target="AppName.Foreground" Value="{ThemeResource SystemControlForegroundChromeDisabledLowBrush}"/>
+                    </VisualState.Setters>
+                </VisualState>
+            </VisualStateGroup>
+        </VisualStateManager.VisualStateGroups>
         <Grid x:Name="BackgroundElement"
               Background="Transparent"
               Height="32">
-            <VisualStateManager.VisualStateGroups>
-                <VisualStateGroup x:Name="WindowFocusStates">
-                    <VisualState x:Name="WindowFocused"/>
-                    <VisualState x:Name="WindowNotFocused">
-                        <VisualState.Setters>
-                            <Setter Target="AppName.Foreground" Value="{ThemeResource SystemControlForegroundChromeDisabledLowBrush}"/>
-                        </VisualState.Setters>
-                    </VisualState>
-                </VisualStateGroup>
-            </VisualStateManager.VisualStateGroups>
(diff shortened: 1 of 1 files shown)


rudyhuyn:
> Any chance that you can explain for myself (and the rest of our readers here)?

Of course!

The attached property `VisualStateManager.VisualStateGroups` can only be attached to **the root XAML node of the control's content** to work with VisualStateManager::GoToState(...),  else it won't work (hard to detect because it won't crash the app or display an error/warning at compilation time).

The previous change added an additional grid (`BackgroundElement`) moving by accident the VisualStateManager out of the root node and making `VisualStateManager::GoToState` ineffective.
image.

HowardWolosky:
Thanks @rudyhuyn.

> The attached property `VisualStateManager.VisualStateGroups` can only be attached to **the root
> XAML node of the control's content** to work with VisualStateManager::GoToState(...),  else it won't 
> work (hard to detect because it won't crash the app or display an error/warning at compilation
> time).

That was what I figured based on your change.  Thanks for the confirmation.  I then found corresponding documentation to support that statement as well:

https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.VisualStateManager#remarks

>Control authors or app developers add VisualStateGroup object elements to the root element of a
> control template definition in XAML, using the VisualStateManager.VisualStateGroups attached
> property.

But then I'm trying to understand other places where it appears to _not_ be on the root XAML node of the control.  Take for example,, the `TitleContentControl`:

https://github.com/microsoft/calculator/blob/24d997dd00534938cb880722108a0f85ae4c3611/src/Calculator/Views/Calculator.xaml#L928-L944

Here we have a `Grid` which contains a `Border`, and the `VisualStateMananger.VisualStateGroups` is on the `Border` as opposed to the `Grid` element which should in theory be considered the root node.

I think your change here is completely right, no doubt.  I'm just trying to wrap my head around some inconsistent usage (as far as I can tell) of VSM then within the project (although, as far as I can tell, this is the only other example in the project besides the one in this PR where VSM isn't attached to the root node).

Thoughts?

rudyhuyn:
> Here we have a Grid which contains a Border, and the VisualStateMananger.VisualStateGroups is on the Border as opposed to the Grid element which should in theory be considered the root node.

Even if the Border is a X(A)ML descendant of the Grid, they are not related and are part of 2 different XAML tree structure because the ContentTemplate has its own XAML tree structure.

The Grid is a content of the Page, and is part of its visual tree.
The Border isn't part of the visual tree but part of a template definition describing a Button and is the root node of this definition. (defined in Grid Resources). 

So the definition is still correct: the VisualStateManager is attached to the root element of a control template.

image

HowardWolosky:
Ah.  Good catch.  I completely missed that there was a `ControlTemplate` embedded within the `ContentControl` there.  Thanks for looking into all of this.

Issue #587, opened 2019-07-17
In small layouts, history button overlaps with window management buttons in title bar


MicrosoftIssueBot:
**Describe the bug**
Notice how sometimes the history button will render overlapped with content in the title bar.

**Steps To Reproduce**
This bug was filed from another source and did not have detailed repro steps, but anecdotally, I have seen this behavior. It would be great if someone could help consistently repro this.

**Expected behavior**
History button is positioned below title bar:
![image](https://user-images.githubusercontent.com/1414423/61338118-e5bd0000-a7ec-11e9-89a4-3811e66469e4.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/1414423/61337980-5a436f00-a7ec-11e9-9d25-f847ffe5249d.png)

**Device and Application Information**
 - OS Build: 10.0.18942.0
 - Architecture: X64
 - Application Version: 10.1906.25.0
 - Region: en-US
 - Dev Version Installed: True

**Requested Assignment**
I'm just reporting this problem.  I don't want to fix it.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

zherman:
I think https://github.com/microsoft/calculator/commit/bd04c92c1cd4c9b0e35684a838e7f363c8a7a439#diff-62ce6ee1acd8a14dee00a6f580125a25 might address this issue. The CustomTitleBar XAML element having a default Visibility of Collapsed I think was the root cause of the bug. The commit I found seems to have removed that assignment in MainPage.Xaml

rudyhuyn:
Is there a way to know the version number of the application used by the user who reported the problem? (I suppose the one in the original message is not the user's one)

grochocki:
> Is there a way to know the version number of the application used by the user who reported the problem?

I tried to reach out to the original bug filer, but have not heard back.

rudyhuyn:
@grochocki: This issue can be closed!

grochocki:
Addressed as part of #634.

PR #634, opened 2019-08-07
General improvement of the title bar for Always-On-Top.

rudyhuyn:
Code cleaning + Make sure issues similar to #587 won't happen with AOT.

- Make TitleBar generic and remove ApplicationViewModel dependency, so other apps can use this control
- Force the calculation of the title bar visibility when we enter and leave AOT.
- Use Visual States instead of bindings to manage the 2 states of the title bar.
- (very) small performance improvement: deferred loading of `ExitAlwaysOnTopButton`
- Make sure `ExitAlwaysOnTopButton` has the same height than the title bar (the height was fixed)
- Remove the converter and properties not used by `ExitAlwaysOnTopButton` (FontWeight, AllowFocusWhenDisabled, Grid.Row)

In ApplicationViewModel:
- Use macro to declare `IsAlwaysOnTop`
- Make `IsAlwaysOnTop` and `DisplayNormalAlwaysOnTopOption` read-only

- Manually


src/Calculator/Views/TitleBar.xaml
```diff
@@ -32,25 +37,22 @@
...
-                Click="AlwaysOnTopButtonClick"
+                Click="AlwaysOnTopButton_Click"
                 Width="46"
-                Height="32"
+                Height="Auto"
+                Visibility="Collapsed"
                 HorizontalAlignment="Left"
                 HorizontalContentAlignment="Center"
                 Background="Transparent"
-                FontWeight="Thin"
```
sanderl: The button looks bold now, I think you still want the FontWeight so that the style of the two AOT buttons match.

sanderl:
It also looks like there is a conflict in the branch. Can you take a look at that too?

src/Calculator/Views/TitleBar.xaml
rudyhuyn: My fault! Fixed!

sanderl:
Thanks for improvements to AOT.

PR #635, opened 2019-08-07
Fix titlebar changing color based off window focus state

zherman:
## Fixes #.
#631 

The VisualStateManager tag associated with the visual states used for the titlebar got moved into a difference scope as part of the Always-on-Top feature by mistake. This caused the states to not function properly. Moving the XAML elements back to their original scope fixes the issue. The titlebar is still not visible when in Always-on-Top mode.

Manual Testing


zherman:
Didnt realize there was a fix in PR already

rudyhuyn:
> Didnt realize there was a fix in PR already

No prob! I applaud and appreciate your efforts, reactivity and efficiency, you did a great job!

PR #637, opened 2019-08-08
Adds zoom buttons to graph controls

greedyAI:
- Zoom-in and zoom-out buttons have been added. When clicked or pressed, they call `GraphingControl`'s `ScaleRange` function. These buttons take advantage of `RepeatButton`'s API to allow for hold-down repeat click behavior.
- Zoom reset button has been added. When clicked or pressed, it calls `GraphingControl`'s `ResetGrid` function.
- Appropriate keyboard shortcuts and light-mode styling for these buttons have also been added.

- Manual testing.

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -16,7 +16,6 @@
             <Setter Property="Width" Value="36"/>
             <Setter Property="Height" Value="36"/>
             <Setter Property="CornerRadius" Value="18"/>
-            <Setter Property="Background" Value="White"/>
             <Setter Property="BorderThickness" Value="1"/>
             <Setter Property="FontSize" Value="12"/>
             <Setter Property="Delay" Value="500"/>
@@ -26,16 +25,8 @@
             <Setter Property="Width" Value="36"/>
             <Setter Property="Height" Value="36"/>
             <Setter Property="CornerRadius" Value="18"/>
-            <Setter Property="Background" Value="White"/>
             <Setter Property="BorderThickness" Value="1"/>
             <Setter Property="FontSize" Value="12"/>
-        </Style>
-        <Style x:Key="ActionButtonStyle" TargetType="Button">
-            <Setter Property="Width" Value="36"/>
-            <Setter Property="Height" Value="36"/>
-            <Setter Property="Background" Value="White"/>
-            <Setter Property="BorderThickness" Value="0"/>
-            <Setter Property="FontSize" Value="18"/>
         </Style>
         <converters:BooleanToVisibilityConverter x:Name="BooleanToVisibilityConverter"/>
         <converters:BooleanToVisibilityNegationConverter x:Name="BooleanToVisibilityNegationConverter"/>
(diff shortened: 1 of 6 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -30,6 +52,30 @@
         <Grid x:Name="LeftGrid"
               Grid.Row="1"
               Grid.Column="0">
+            <Grid.Resources>
```
rudyhuyn: Unfortunately, XAML doesn't provide a elegant way to include resources in Style, so we need to be hacky :(

But I would recommend to limit the scope of these brushes a little more to be sure to not affect Buttons and RepeatButtons in the Grapher control and all future controls in this Grid.

Two proposals:
- Limit the scope to the buttons:
    - create a ResourceDictionary file with these brushes 
    - write: `<RepeatButton><RepeatButton.Resources><ResourceDictionary Source="ZoomButtonResourceDictionary.xaml" /></RepeatButton.Resources>`
- Create a new Grid hosting the buttons and move these resources there

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -215,6 +265,124 @@
...
+                                    <ColumnDefinition Width="Auto"/>
+                                    <ColumnDefinition Width="Auto"/>
+                                </Grid.ColumnDefinitions>
+                                <Button Margin="0,0,4,0"
+                                        Grid.Column="0"
+                                        Background="Transparent"
+                                        Content="Degrees"/>
+                                <Button Margin="4,0,0,0"
+                                        Grid.Column="1"
+                                        Background="Transparent"
```
sanderl: Is this part of a different PR? I'm not seeing a flyout for the zoom buttons.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -215,6 +262,34 @@
                     </Flyout>
                 </Button.Flyout>
             </Button>
+
+            <RepeatButton Margin="0,0,12,12"
+                          x:Uid="zoomOutButton"
+                          FontFamily="{StaticResource SymbolThemeFontFamily}"
+                          Content="&#xE738;"
+                          RequestedTheme="Light"
```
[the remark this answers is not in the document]
sanderl: Tooltips and narrator should be supported as well.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -12,6 +12,31 @@
...
+            <Setter Property="Height" Value="36"/>
+            <Setter Property="CornerRadius" Value="18"/>
+            <Setter Property="Background" Value="White"/>
+            <Setter Property="BorderThickness" Value="1"/>
+            <Setter Property="FontSize" Value="12"/>
+        </Style>
+        <Style x:Key="ActionButtonStyle" TargetType="Button">
+            <Setter Property="Width" Value="36"/>
+            <Setter Property="Height" Value="36"/>
+            <Setter Property="Background" Value="White"/>
```
sanderl: The buttons do not resize with the window size of calculator. I think there should be some visual states added to keep the buttons in proportion to the window size.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -215,6 +265,124 @@
...
+                          Style="{ThemeResource ZoomRepeatButtonStyle}"
+                          VerticalAlignment="Bottom"
+                          HorizontalAlignment="Right">  
+            </RepeatButton>
+
+            <Button Margin="0,0,12,108"
+                    x:Uid="zoomResetButton"
+                    FontFamily="{StaticResource SymbolThemeFontFamily}"
+                    Content="&#xE895;"
+                    RequestedTheme="Light"
```
sanderl: The reset button should be the last button in the set, right now it is first.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
joseartrivera: Steph and I discussed offline. We still need to add visual states to make either the graph or equation area take the full screen when in a small window size. We can handle resizing the buttons at that point with those new visual states if they still need adjustment. That can be done in another PR.

We also discussed that this PR should add in the high contrast/narrator pieces mentioned in other comments.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -53,7 +102,8 @@
                     FontFamily="{StaticResource SymbolThemeFontFamily}"
                     FontSize="18"
                     Content="&#xE70F;"
-                    RequestedTheme="Light">
+                    RequestedTheme="Light"
+                    Background="White">
```
joseartrivera: Why is this white now?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -118,3 +118,25 @@ void GraphingCalculator::TextBoxGotFocus(TextBox^ sender, RoutedEventArgs^ e)
 {
     sender->SelectAll();
 }
+
+void GraphingCalculator::ZoomInButtonClick(Object ^ /* parameter */)
+{
+    constexpr double scrollDamper = 0.15;
+    double scale = 1.0 + (abs(zoomButtonDelta) / WHEEL_DELTA) * scrollDamper;
```
joseartrivera: Since in this case, zoomButtonDelta never changes, you should just do this calculation once and hard code it here. Or make this a constexpr so that the value is computed at compile time.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -118,3 +118,25 @@ void GraphingCalculator::TextBoxGotFocus(TextBox^ sender, RoutedEventArgs^ e)
...
+    scale = 1.0 / scale;
+
+    GraphingControl->ScaleRange(0, 0, scale);
+}
+
+void GraphingCalculator::ZoomOutButtonClick(Object ^ /* parameter */)
+{
+    constexpr double scrollDamper = 0.15;
```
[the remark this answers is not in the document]
joseartrivera: You could also keep the ScaleRange function protected in grapher.cpp and instead create a public function that is just simply called ZoomIn/ZoomOut.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -215,6 +265,124 @@
...
+                            <TextBlock Text="Units"/>
+                            <Grid HorizontalAlignment="Stretch">
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="Auto"/>
+                                    <ColumnDefinition Width="Auto"/>
+                                </Grid.ColumnDefinitions>
+                                <Button Margin="0,0,4,0"
+                                        Grid.Column="0"
+                                        Background="Transparent"
+                                        Content="Degrees"/>
```
joseartrivera: This and the rest of the strings should be localized.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -30,6 +55,30 @@
         <Grid x:Name="LeftGrid"
               Grid.Row="1"
               Grid.Column="0">
+            <Grid.Resources>
+                <SolidColorBrush x:Key="RepeatButtonBackgroundPointerOver" Color="White"/>
```
joseartrivera: Should this really be white or transparent?

Issue #385, opened 2019-03-26
Narrator reads information of "0th" bit button incorrectly


MicrosoftIssueBot:
**Repro Steps**
* Launch "Calculator" app. Launch narrator.
* Navigate to Menu and select "Programmer".
* Navigate to “Bit toggling keypad” radio button and select it.
* Navigate to "0th" bit button and observe the narrator announcement. 

**Actual Result**
Narrator reads "0 and th" separately for 0th bit button. 

**Expected Result**
Narrator should read "0th" bit button on/off.

**Test Environment**
OS Version: Windows 10 Build 18351.1
App Name: Calculator 
App Version: 10.1902.16.0

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

brulecap:
New to the calculator community and would love to contribute any way I can.

I verified that this is not only a problem with Narrator but also with the NVDA screen reader. Both pronounce "0th" as "0" and "th" separately. Other screen readers?

I then went down a bit of a rabbit hole when considering the propriety/impropriety of using the ordinal term zeroth(0th). It seems the 1st bit would be bit position 0, but considering that we are performing 0 based numbering, I grudgingly accepted the propriety of using zeroth(0th) to denote the first bit.

2 solutions I came up with:

1)Change the name="0" data value from 0th( or language-specific equivalents) to 0(non-language-specific) in all resources.resw locale files.
	The result would be that Narrator would read "0 bit value 0" instead of "0 th bit value 0". All others would stay the same. i.e. "32nd bit value 0".

2)Change the string returned from BitFlipAutomationNameConverter.
	Example: Instead of returning "1st bit value 1" have it return "Bit 1 value 1".
We could use the name contained in "parameter" in the parameter list of BitFlipAutomationNameConverter to achieve this. Should we? The right way would be to change all name data values from their ordinal to cardinal representation in all resources.resw locale files? i.e: value "1" instead of "1st".

I welcome any feedback. I would love to tackle this once a proper resolution is decided upon by the community.

danbelcher-MSFT:
What about "low bit"/"high bit"?

grochocki:
How about just ["least significant bit"](https://en.wikipedia.org/wiki/Bit_numbering)?

brulecap:
[...]

PR #640, opened 2019-08-14
Optimize BitFlipPanel to suppress flicker when users switch between bit lengths

### LINKED ISSUE #408: The programmer tab's byte view flickers when switching between byte, dword, and qword only when standard view has not been used
**Describe the bug**

When switching between byte, qword, and dword on the programmers view, the zeros (that are normally dimmed) will flicker (going to grey, then to white, then to grey within a few milliseconds after the button was clicked once) and then go back to grey when the modes are switched. This issue does not occur after other calculations are performed on the standard view (e.g. adding `1+1` on the standard view), then going back to the programmers view.

**Steps To Reproduce**

1. Open Calculator, maximize window to fill entire screen
2. Click on hamburger menu, then click on Programmer
3. Click on the icon on the left-hand side that has four dots arranged in a number pad to go to byte view
4. Click on "MS", then set value to 128 by clicking on the appropriate binary numbers. Perform steps 5-8 quickly in a loop.
5. Click on "Qword"
6. Click "Dword"
7. Click "Word"
8. Click "Byte". The numbers will flicker.
9. Click on "MS"
10. Click Qword, Dword, Word again quickly, and notice that the number do not flicker just this once.
11. Click on Qword, Dword, Word again quickly and notice that the numbers do flicker now
12. Switch to standard mode. Type in 56 then add 111 to it. Hold down enter for five seconds.
13. Click on MS
14. Switch back to Programmer mode
15. Go back into the byte view
16. When clicking on Word, Dword, Word again (after entering 128), the numbers will not flicker no matter how many times it was clicked

**Expected behavior**

The numbers should switch between white to grey, with no other in-between transitions.

**Screenshots**

This is what I call the byte view:

image

This is the standard view after performing a long sample calculation then clicking MS (but going back to it from programmers view)

image

 - OS Build: 10.0.17134.0
 - Architecture: X64
 - Application Version: 10.1812.10048.0 (latest version after re-installing it and checking Windows Store for updates)

**Additional context**

rudyhuyn:
- Create new type `BitLength` and use only one way to represent bit lengths in `CopyPasteManager`, `StandardCalculatorViewModel` (instead of 4 booleans + 1 enum)
- Optimize how we convert the binary number into a list of booleans:
   - Before:
       - In VM: call GetStringForDisplay, GroupDigitsPerRadix (add spaces), localizer.LocalizeDisplayValue (localize the string to use the localized '0'/'1' characters)
       - In View: remove LTR, RTL, spaces, etc... , remove characters added by LocalizeDisplayValue, add padding ('0'), compare each characters with the local '0'
    - Now:
        - In VM: call GetStringForDisplay, convert the string in vector<bool>
        - in View: nothing
 - remove the very heavy `BitFlipAutomationNameConverter` converter and replace it by a more optimized way to generate the string.
- The view model is now responsible for sending commands to CalculatorManager instead of the view.
- optimize `CalculatorProgrammerBitFlipPanel` 
  - remove half of native bindings
  - modify how the application manages the visibility of this control

- Tested in all modes

### Result
- CPU usage decreased when users enter a new number or change bit length
- No more flickering


src/CalcManager/CalculatorManager.cpp
```diff
@@ -613,9 +613,9 @@ namespace CalculationManager
         }
     }
 
-    wstring CalculatorManager::GetResultForRadix(uint32_t radix, int32_t precision)
+    wstring CalculatorManager::GetResultForRadix(uint32_t radix, int32_t precision, bool groupDigitsPerRadix)
```
EriWong: >wstring CalculatorManager::GetResultForRadix(uint32_t radix, int32_t precision, bool groupDigitsPerRadix) [](start = 4, length = 105)

It looks like we don't have tests that covered this before...  Can you add a test or two to sanity check this method(as well as your addition to it)?

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -65,7 +65,8 @@ void CopyPasteManager::CopyToClipboard(String ^ stringToCopy)
     Clipboard::SetContent(dataPackage);
 }
 
-task<String ^> CopyPasteManager::GetStringToPaste(ViewMode mode, CategoryGroupType modeType, int programmerNumberBase, int bitLengthType)
+task<String ^>
+CopyPasteManager::GetStringToPaste(ViewMode mode, CategoryGroupType modeType, int programmerNumberBase, BitLength bitLengthType)
```
EriWong: nit: is the line break intentional?

src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp
```diff
@@ -185,35 +174,51 @@ void CalculatorProgrammerBitFlipPanel::UpdateCheckedStates()
...
-    wstring displayValue = Model->BinaryDisplayValue->Data();
-    for (const wchar_t& c : displayValue)
+    m_updatingCheckedStates = true;
+    unsigned int bitIndex = 0;
+    auto it = m_flipButtons.begin();
+    for (bool val : Model->BinaryDigits)
     {
-        if (find(begin(unwantedChars), end(unwantedChars), c) == unwantedChars.end())
+        // Highest bit (64) is at index 0 in bit string.
+        // To get bit 0, grab from opposite end of string.
```
EriWong: I think this comment makes more sense in StandardCalculatorViewModel.cpp where you actually reverse the string.  I was a little confused when reading this as to why you were using reverse iterators until I saw this comment here.

src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp
```diff
@@ -185,35 +174,51 @@ void CalculatorProgrammerBitFlipPanel::UpdateCheckedStates()
...
-    m_updatingCheckedStates = false;
+String ^ CalculatorProgrammerBitFlipPanel::GenerateAutomationPropertiesName(int position, bool value) const
+{
+    auto resourceLoader = AppResourceProvider::GetInstance();
+
+    String ^ indexName = resourceLoader.GetResourceString(ref new Platform::String(to_wstring(position).c_str()));
+    String ^ bitName = resourceLoader.GetResourceString(L"BitAutomationName");
+    String ^ valueName = resourceLoader.GetResourceString(L"ValueAutomationName");
+
+    return indexName + bitName + valueName
```
EriWong: >indexName + bitName + valueName [](start = 11, length = 31)

I think automation name gets localized.  As such, I don't think you want to be concatenating in this way.

EriWong: Looks like the converter was doing this the same way before your change.  We can probably leave this as is and track this as a separate issue if you don't fix this here.

rudyhuyn: Yes. But you are right, we can improve this part. I will update the PR tonight

rudyhuyn:
Note about the new strings:
- The new template is designed to be compatible with the change needed for #385 
- `0` (zero) and `1` (one) don't need to be localized
- trailing space removed from resources `1`, `2`, `3`, it should not be part of the localization.

src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp
```diff
@@ -185,35 +175,49 @@ void CalculatorProgrammerBitFlipPanel::UpdateCheckedStates()
...
+    for (bool val : Model->BinaryDigits)
     {
-        if (find(begin(unwantedChars), end(unwantedChars), c) == unwantedChars.end())
+        auto checkbox = *it;
+        if (forceUpdate || checkbox->IsChecked->Value != val)
         {
-            stream << c;
+            checkbox->IsChecked = val;
+            // Only generate the string when the value changed
+            checkbox->SetValue(AutomationProperties::NameProperty, GenerateAutomationPropertiesName(bitIndex, val));
```
EriWong: When testing out your fix for localization, I found I wasn't hitting this codepath when buttons are getting updated.  As a result, I'm not seeing automation name change when a bit gets flipped.    It looks like checkbox->IsChecked->Value has already been updated to val at this point, so the check is always passing.

rudyhuyn: You are right. It's now fixed!

rudyhuyn:
> Looks like you have a conflict. I can accept this PR once you've resolved the merge. Thanks!

Done!

PR #641, opened 2019-08-16
Remove ProgModeRadixChange unload

joseartrivera:
# Description of the changes:
Looks like by the time the OnUnloaded event is called Model can already be null in cases where one of the Calculator windows is being closed. This PR uses a dependency property instead of an event to fix the crash.

Manual testing


joseartrivera:
> As we discussed, I think you might benefit from switching to registering/unregistering when the [DataContext changes](https://stackoverflow.com/questions/45323145/remove-event-from-viewmodel-when-the-user-control-is-unloading)

The DataContextChanged event doesn't seem to fire on app shutdown when the DataContext is set to nullptr. Also it doesn't give us access to the old value to unregister the event. https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.datacontextchangedeventargs#properties

I have updated the code to remove the unload code and revert the behavior.

joseartrivera:
Updated the design to no longer use events. Now Numberpad has a dependency property that takes the radix mode. This new property is bound to the view model's radix mode property.

src/Calculator/Views/NumberPad.xaml.cpp
```diff
@@ -85,6 +85,7 @@ void NumberPad::ProgModeRadixChange()
     }
 }
 
+
```
HowardWolosky: extra linefeed unnecessary here..

PR #652, opened 2019-08-30
Fix scroll buttons in CalculationResult

### LINKED ISSUE #650: CalculationResult doesn't correctly update the status of scroll buttons
## Describe the bug
The `CalculationResult` control has some issues with scroll buttons:
- when using a mouse, scroll buttons are only updated after the mouse has left the control
- buttons are never updated with touch events.
- buttons are not updated when users scroll

ℹ _Issue visible in the public and dev version of the Calculator_

## Steps To Reproduce
- horizontally resize the window to the minimum 
- switch to the Programmer mode
- enter the number ‭999999999999999999‬ (decimal mode)
- Click on "BIN" to switch to binary mode
- try one of the 3 actions listed above

## Expected behavior
Visibility of scroll buttons should be automatically updated

## Screenshots

4FF2F4BD-35E5-488A-8F04-376FBF11F586

## Device and Application Information
 - OS Build: 10.0.18362.0
 - Architecture: X64
 - Application Version: 10.1906.53.0
 - Region: en-US
 - Dev Version Installed: True

## Requested Assignment

If possible, I would like to fix this. 🖐

rudyhuyn:
- Check ViewChanged events and verify if we need to hide/show `ScrollLeft`/`ScrollRight`
- To improve the reliability, I've added a sufficient safety margin (4 pixels) to guarantee the detection (it happens sometimes when users scroll to the maximum, HorizontalOffset is a little low by a few sub-pixels and when users scroll again to the right, no events are launched)

Code quality:
- unregister `SizeChanged`, `ScrollLeft::Click` and `ScrollRight::Click` events (were missing)
- make `ScrollLeft` and `ScrollRight` independent (templates in the app always use both, but it's cleaner this way).
- rename `TextContainerSizeChanged` to `OnTextContainerSizeChanged` to be consistent with other events.
- support <kbd>&larr;</kbd> and <kbd>&rarr;</kbd> keys even if `ScrollLeft` and `ScrollRight` are not part of the template.

- Manually

### DIFF
--- a/src/Calculator/Controls/CalculationResult.cpp
+++ b/src/Calculator/Controls/CalculationResult.cpp
@@ -46,6 +46,10 @@ DEPENDENCY_PROPERTY_INITIALIZATION(CalculationResult, DisplayStringExpression);
 #define WIDTHTOFONTOFFSET 3
 #define WIDTHCUTOFF 50
 #define FONTTOLERANCE 0.001
+
+// We need a safety margin to guarantee we correctly always show/hide ScrollLeft and ScrollRight buttons when necessary.
+// In rare cases, ScrollViewer::HorizontalOffset is a little low by a few (sub)pixels when users scroll to one of the extremity
+// and no events are launched when they scroll again in the same direction
 #define SCROLL_BUTTONS_APPROXIMATION_RANGE 4
 
 StringReference CalculationResult::s_FocusedState(L"Focused");
--- a/src/Calculator/Controls/CalculationResult.cpp
+++ b/src/Calculator/Controls/CalculationResult.cpp
@@ -68,22 +68,37 @@ Platform::String ^ CalculationResult::GetRawDisplayValue()
 
 void CalculationResult::OnApplyTemplate()
 {
-    assert((m_scrollLeft == nullptr && m_scrollRight == nullptr) || (m_scrollLeft != nullptr && m_scrollRight != nullptr));
     if (m_textContainer)
     {
         if (m_textContainerLayoutChangedToken.Value != 0)
         {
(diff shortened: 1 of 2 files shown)


src/Calculator/Controls/CalculationResult.cpp
```diff
@@ -322,48 +329,24 @@ void CalculationResult::OnScrollClick(Object ^ sender, RoutedEventArgs ^ /*e*/)
...
-    {
-        ShowHideScrollButtons(::Visibility::Collapsed, ::Visibility::Visible);
+        return;
     }
-    else // Width is larger than the container and right most part of the number is shown. Should be able to scroll left.
+
+    if (m_scrollLeft != nullptr)
     {
-        ShowHideScrollButtons(::Visibility::Visible, ::Visibility::Collapsed);
+        m_scrollLeft->Visibility = m_textContainer->HorizontalOffset > SCROLL_BUTTONS_APPROXIMATION_RANGE ? ::Visibility::Visible : ::Visibility::Collapsed;
```
EriWong: >SCROLL_BUTTONS_APPROXIMATION_RANGE  [](start = 71, length = 35)

It's probably worth taking your pr comment and putting it in here as an actual comment.  From the code itself, it isn't immediately obvious why this buffer is here.

rudyhuyn: Comment added just before the definition of SCROLL_BUTTONS_APPROXIMATION_RANGE!

Aneeshsundaresh:
https://github.com/microsoft/calculator/labels/needs%20author%20feedback

PR #662, opened 2019-09-06
Update Calc Engine for new functions needed for keyboard refresh

### LINKED ISSUE #428: Update Calculator keyboard for greater consistency and extensibility

joseartrivera:
## Fixes #428

Spec for this change can be found here:
[Update Calculator keyboard for greater consistency and extensibility](https://github.com/microsoft/calculator-specs/blob/master/active/KeyboardRefresh/README.md)

This PR includes only the Calc Engine changes for the keyboard refresh spec. The UI portion will be a separate PR.
- Updated Calc Engine to support new functionality (sec, csc, cot, hyperbolic, inverse, floor, ceil, rand, log base x, new bit shifts, nand, xor.
- Exposed property to know when to pivot on the Clear Entry button vs Clear button with the IsInputEmpty boolean 

- Manual testing
- Unit tests added


src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -714,6 +733,47 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
         HandleErrorCommand(wParam);
         break;
 
+    case IDC_RAND:
+        if (!m_fIntegerMode)
+        {
+            CheckAndAddLastBinOpToHistory(); // rand is like entering the number
+
+            random_device rd;
+            mt19937 gen(rd());
```
rudyhuyn: NIT: We should create only 1 instance of `mt19937 gen` (first time we use it) instead of creating a new generator on each call to ensure that the distribution of results is done as expected.

src/CalcManager/Header Files/CalcEngine.h
```diff
@@ -147,6 +152,8 @@ class CCalcEngine
     NUM_WIDTH m_numwidth;                    // one of qword, dword, word or byte mode.
     int32_t m_dwWordBitWidth;                // # of bits in currently selected word size
 
+    uint64_t m_carryBit;
```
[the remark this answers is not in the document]
joseartrivera: It was easier to leave it at uint64_t since we use OR operations to set the msb in calculations and the msb and lsb are currently also handled as uint64_t

src/CalcManager/Header Files/CCommand.h
```diff
@@ -136,7 +137,39 @@
...
+#define IDC_UNARYEXTENDEDLAST  IDC_RORC
+
+#define IDC_LASTCONTROL IDC_CEIL
+
+#define IDC_BINARYEXTENDEDFIRST 500
+#define IDC_LOGBASEX            500 //logx(y)
+#define IDC_NAND                501
+#define IDC_NOR                 502
+
+#define IDC_LSHFL               504
```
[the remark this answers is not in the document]
joseartrivera: I believe we can, I can revert if it turns out I am wrong in the UI PR

src/CalcManager/Header Files/CCommand.h
```diff
@@ -136,7 +137,39 @@
 #define IDC_INV 146
 #define IDC_SET_RESULT 147
 
-#define IDC_LASTCONTROL IDC_SET_RESULT
+#define IDC_STRING_MAPPED_VALUES  400
+#define IDC_UNARYEXTENDEDFIRST  IDC_STRING_MAPPED_VALUES
```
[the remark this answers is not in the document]
joseartrivera: I don't think we persistent store these. But I noticed there are many places in code that indirectly rely on a lot of these values being in a particular order (Such as the IDC_BINEDITSTART + 63 as you pointed out earlier) so I would not want to do such a large change in this PR.

src/CalcManager/CEngine/scicomm.cpp
joseartrivera: Moved the initialization to CalcEngine.h. I couldn't figure out how to check if it had already been initialized since I wanted to avoid initializing it on start up.

I tried this code in CalcEngine.h but kept getting the same rand number repeatedly. Any ideas?

```
    bool m_isRandomGeneratorSeeded;
    std::mt19937 m_randomGeneratorEngine;

    std::mt19937 GetRandomGeneratorEngine()
    {
        if (!m_isRandomGeneratorSeeded)
        {
            std::random_device rd;
            m_randomGeneratorEngine = std::mt19937(rd());
            m_isRandomGeneratorSeeded = true;
        }
        return m_randomGeneratorEngine;
    }
```

rudyhuyn: I don't have this issue, here is my code:

**RandomGenerator.h**
```h
#pragma once
#include <random>

class RandomGenerator
{
public:
	double GenerateNumber();
private:
	std::unique_ptr<std::mt19937> m_randomGeneratorEngine;
	std::unique_ptr<std::uniform_real_distribution<>> m_distr;
};
```

**RandomGenerator.cpp**
```cpp
#include "pch.h"
#include "RandomGenerator.h"

using namespace std;

double RandomGenerator::GenerateNumber()
{
	if (m_randomGeneratorEngine == nullptr)
	{
		random_device rd;
		m_randomGeneratorEngine = std::make_unique<std::mt19937>(rd());
		m_distr = std::make_unique<std::uniform_real_distribution<>>(0, 1);
	}
	return (*m_distr.get())(*m_randomGeneratorEngine.get());
}
```



Sample app here: https://1drv.ms/u/s!AkrL-mAvq0QFs8obJ3vwbChR5LvFZA?e=dmBLED

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -16,9 +16,11 @@
 #include <string>
 #include "Header Files/CalcEngine.h"
 #include "Header Files/CalcUtils.h"
+#include "NumberFormattingUtils.h"
 
 using namespace std;
 using namespace CalcEngine;
+using namespace CalcManager::NumberFormattingUtils;
```
joshkoon: Generally, we should maintain the CalcManager as a layer above the CalcEngine, and keep the engine from taking dependencies on the manager.

src/CalcManager/Command.h
```diff
@@ -151,6 +152,34 @@ namespace CalculationManager
         CommandINV = 146,
         CommandSET_RESULT = 147,
 
+        CommandSEC = 400,
```
sanderl: Can you make sure these new commands are hooked up to the ButtonUsage diagnostics?

joseartrivera: These commands will not hook up to diagnostics until the UI portion of this changes comes in since the diagnostics is hooked up to the actual UI button rather than the calc engine commands.

src/CalcManager/CEngine/scicomm.cpp
joseartrivera: Thanks, this works!

PR #669, opened 2019-09-11
CalculationResult: keep focus on Scroll buttons

### LINKED ISSUE #651: CalculationResult should keep the focus on scroll buttons when their visibility changes
## Problem Statement
This is hard to test because of #650, but contrary to `OverflowTextBlock`, `CalculationResult` does nothing to keep the focus when the button `ScrollRight` has the focus and becomes collapsed.

See code of `OverflowTextBlock` here: https://github.com/microsoft/calculator/blob/796d171960a7351bb2c36cac5cec22d3e74ec0de/src/Calculator/Controls/OverflowTextBlock.cpp#L201

## Proposal
`CalculationResult` should have the same behavior than `OverflowTextBlock` and give the focus to `ScrollLeft` when `ScrollRight` has it and becomes collapsed.

## Goals
Get consistent behavior between controls and improve UX.

## Requested Assignment
If possible, I would like to implement this because already done 😉

rudyhuyn:
- Modify CalculationResult::UpdateScrollButtons to switch the focus from ScrollLeft to ScrollRight when one of them is focused but will be collapsed.

- manually

### How to test it
- Reduce window size to the minimum
- Switch to Programmer mode
- Select BIN
- Type 1111111111...1111111

### DIFF
--- a/src/Calculator/Controls/CalculationResult.cpp
+++ b/src/Calculator/Controls/CalculationResult.cpp
@@ -340,17 +340,38 @@ void CalculationResult::UpdateScrollButtons()
         return;
     }
 
+    bool shouldTryFocusScrollRight = false;
     if (m_scrollLeft != nullptr)
     {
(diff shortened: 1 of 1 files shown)


rudyhuyn:
rebased!

PR #672, opened 2019-09-12
Switch to RichEdit math mode in EquationTextBox

joseartrivera:
- Switch to using math mode in the rich edit control in EquationTextBox
- Switch GraphControl to use MathML instead of plain text

- Manual testing

### DIFF
--- a/src/GraphControl/Control/Equation.cpp
+++ b/src/GraphControl/Control/Equation.cpp
@@ -9,6 +9,8 @@ using namespace Windows::UI::Xaml;
 
 namespace GraphControl
 {
+    static constexpr wstring_view s_mathPrefix = L"mml:";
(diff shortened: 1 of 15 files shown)


src/GraphControl/Control/Equation.cpp
```diff
@@ -9,6 +9,8 @@ using namespace Windows::UI::Xaml;
 
 namespace GraphControl
 {
+    static constexpr wstring_view s_mathPrefix = L"mml:";
```
sanderl:
```suggestion
    // Remove mml: formatting specific to RichEditBox control, which is not understood by the graph engine.
    static constexpr wstring_view s_mathPrefix = L"mml:";
```

PR #675, opened 2019-09-18
Modify how narrator names the first and last bits of a number in Bit Flip

### LINKED ISSUE #385: Narrator reads information of "0th" bit button incorrectly

rudyhuyn:
- replace "0th bit"  by "least significant bit"
- add support of  "most significant bit" and update texts when BitLength changes.
  - Note: to keep CPU usage low, we only update texts of the previous checkbox associated to the MSB and the new one.

- Manually

### DIFF
--- a/src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp
+++ b/src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp
@@ -69,7 +69,8 @@ void CalculatorProgrammerBitFlipPanel::OnPropertyChanged(Object ^ sender, Proper
     {
         UpdateCheckedStates(false);
     }
-    else if (e->PropertyName == StandardCalculatorViewModel::IsBitFlipCheckedPropertyName
+    else if (
+        e->PropertyName == StandardCalculatorViewModel::IsBitFlipCheckedPropertyName
         || e->PropertyName == StandardCalculatorViewModel::IsProgrammerPropertyName)
     {
         if (Model->IsBitFlipChecked && Model->IsProgrammer)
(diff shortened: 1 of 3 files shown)


rudyhuyn:
(Let me know if you also want to use "most significant bit")

grochocki:
If it is not too much trouble :)

rudyhuyn:
> If it is not too much trouble :)

Will do it! It's a little more complicated than LSB (but still simple) because we need to update the text of the previous MSB and the new one when we change the bit length.

src/Calculator/Views/CalculatorProgrammerBitFlipPanel.xaml.cpp
```diff
@@ -190,10 +193,13 @@ void CalculatorProgrammerBitFlipPanel::UpdateCheckedStates(bool updateAutomation
...
     auto it = m_flipButtons.begin();
     int index = 0;
+    bool mustUpdateTextOfMostSignificantDigits = m_currentValueBitLength != Model->ValueBitLength;
+    int previousMSDPosition = GetIndexOfLastBit(m_currentValueBitLength);
+    int newMSDPosition = GetIndexOfLastBit(Model->ValueBitLength);
     for (bool val : Model->BinaryDigits)
     {
         FlipButtons ^ flipButton = *it;
-        if (updateAutomationPropertiesNames)
+        if (updateAutomationPropertiesNames || (mustUpdateTextOfMostSignificantDigits && (index == previousMSDPosition || index == newMSDPosition)))
```
joseartrivera: I am unsure if it is related to this PR, but I get the incorrect value being read out from Narrator in this scenario:

1. Set size to "BYTE"
2. Toggle the most significant bit to 1
3. Set size to "QWORD"
4. All of the previously disabled bits are now set to 1 **(Root cause of the bug?)**
5. Narrator reads them all as 0 instead of 1 as you tab to each one

Let me know if I should just write this up as a separate issue.

rudyhuyn: This is the expected behavior. 
If the most significant bit == 1, it means the number is negative, and based on Two's complement, it's how you should represent a negative number: https://en.wikipedia.org/wiki/Two%27s_complement

**-2**
Byte: ‭11111110‬
QWORD: ‭1111111111111111111111111111111111111111111111111111111111111110‬

What is not normal is that the narrator says "0" instead of "1", it is not related to this PR, but I will include the fix to it

rudyhuyn: Fixed!

PR #681, opened 2019-09-25
Update unit test signature

DavidShoe:
Update unit test signature so that VS 2019 will build the UT project.  The signature check is enforced in VS 2019.


DavidShoe:
@sanderl Here is the signature change.

PR #687, opened 2019-09-30
Add E-notation support to CopyPasteManager for standard mode

### LINKED ISSUE #683: Clarify whether the application should allow users to paste numbers in scientific notation when using Standard mode or not
## Describe the bug
It is not clear if the calculator should allow users to paste expressions with scientific notations (E notation, for example 12e+10) when the calculator is in Standard mode.

Some unit tests and methods show that we should support it:
https://github.com/microsoft/calculator/blob/41e2e975918c6838947e29a046778bd4d57feeee/src/CalculatorUnitTests/CopyPasteManagerTest.cpp#L133

https://github.com/microsoft/calculator/blob/41e2e975918c6838947e29a046778bd4d57feeee/src/CalcViewModel/Common/CopyPasteManager.cpp#L197

while other methods like `CopyPasteManager::ExpressionRegExMatch` don't accept them when the calculator is in Standard mode, making it impossible to paste a number in scientific notation.

### Why we should allow it
- the standard mode can display number with this notation, for example if they calculate `999999999999*999999999999`, so we should probably allow users to copy the result and paste it later.

### Why we should not allow it
- This mode doesn't have a `exp` button, meaning the only way to enter a number in scientific mode will be via copy/paste (and via the history).

## Steps To Reproduce
- launch the Calculator
- paste `12e+10`
- the calculator will display 'INVALID INPUT'

## Expected behavior

Currently unknown

## Device and Application Information**
 - OS Build: 10.0.18363.0
 - Architecture: X64
 - Application Version: 10.1907.24.0
 - Region: en-US
 - Dev Version Installed: True

## Requested Assignment

## Bonus

If we allow users to do it, we should also decide if we want to allow them to copy these type of numbers with unit converters.

rudyhuyn:
- modify `CopyPasteManager::standardModePatterns` to support E-notation

#### Bonus:
- merge the 2 regular expressions from ` CopyPasteManager::scientificModePatterns` to improve performance
- Fix `unitConverterPatterns` (the regular expression accepted empty strings) and use `c_signedDecFloat` instead of the custom expression.

- Manually

### DIFF
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -26,7 +26,7 @@ static const wstring c_wspc = L"[\\s\\x85]*";
 static const wstring c_wspcLParens = c_wspc + L"[(]*" + c_wspc;
 static const wstring c_wspcLParenSigned = c_wspc + L"([-+]?[(])*" + c_wspc;
 static const wstring c_wspcRParens = c_wspc + L"[)]*" + c_wspc;
-static const wstring c_signedDecFloat = L"[-+]?\\d*(\\d|[.])\\d*";
+static const wstring c_signedDecFloat = L"[-+]?\\d+(\\.\\d*)?|\\.\\d+";
 static const wstring c_optionalENotation = L"(e[+-]?\\d+)?";
 
 // Programmer Mode Integer patterns
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -27,6 +27,7 @@ static const wstring c_wspcLParens = c_wspc + L"[(]*" + c_wspc;
 static const wstring c_wspcLParenSigned = c_wspc + L"([-+]?[(])*" + c_wspc;
 static const wstring c_wspcRParens = c_wspc + L"[)]*" + c_wspc;
 static const wstring c_signedDecFloat = L"[-+]?\\d*(\\d|[.])\\d*";
+static const wstring c_optionalENotation = L"(e[+-]?\\d+)?";
 
 // Programmer Mode Integer patterns
 // Support digit separators ` (WinDbg/MASM), ' (C++), and _ (C# and other languages)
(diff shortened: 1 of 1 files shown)


HowardWolosky:
I'm inclined to agree with you @rudyhuyn  -- there should be at least one digit. @grochocki do you have thoughts here?

grochocki:
> Should we only accept expression with minimum 1 digit? (.1, 1., 1.1...)

Technically, sure, though it is kind of a no-op, right? If you are proposing we throw up "Invalid Input", I would say that unless it is causing any _problems_, we should continue to support it.

rudyhuyn:
if someone pastes expressions that seem to come from the emoji picker like `8*.*8`, `-.-`,  `-( . )-( . )-`, I would expect an `Invalid Input`. 

The current regular expression is a little too lax, it's something we can improve if you want.

grochocki:
> I would expect an Invalid Input.

For those pasted expressions (and the many many other combinations), I agree. I suppose I was over-indexing on the special case of pasting in just a decimal point.

rudyhuyn:
PR updated with a new regex to parse numbers

src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -37,11 +38,9 @@ static const wstring c_binProgrammerChars = L"[0-1]+((_|'|`)[0-1]+)*";
 static const wstring c_uIntSuffixes = L"[uU]?[lL]{0,2}";
 
 // RegEx Patterns used by various modes
-static const array<wregex, 1> standardModePatterns = { wregex(c_wspc + c_signedDecFloat + c_wspc) };
-static const array<wregex, 2> scientificModePatterns = {
-    wregex(L"(" + c_wspc + L"[-+]?)|(" + c_wspcLParenSigned + L")" + c_signedDecFloat + c_wspcRParens),
-    wregex(L"(" + c_wspc + L"[-+]?)|(" + c_wspcLParenSigned + L")" + c_signedDecFloat + L"e[+-]?\\d+" + c_wspcRParens)
-};
+static const array<wregex, 1> standardModePatterns = { wregex(c_wspc + c_signedDecFloat + c_optionalENotation + c_wspc) };
```
EriWong: >c_signedDecFloat + c_optionalENotation [](start = 71, length = 38)

Does this work?  I believe as written this only catches the alternate half of c_signedDecFloat.  ie: This will match ".5e3" but not "0.5e3".  I think you need parens around c_signedDecFloat

rudyhuyn: Good catch! the second commit to exclude `.` (see conversations above) broke everything 🤦‍♂️
 
It's now fixed!

rudyhuyn:
Parentheses added to fix issues when we concatenate regex strings.

PR #698, opened 2019-10-04
Fix the focus when right-clicking CalculationResult's TextBlock

### LINKED ISSUE #642: Right-click on numbers fails to highlight after 'selecting all' from context menu
**Describe the bug**
When trying to "select all" when right-clicking on the main display output, you are not able to actually select all text.

**Steps To Reproduce**
1. Enter in a number in Calculator
2. Right-click on the number and click "select all"

**Expected behavior**
Display text appears selected.

**Screenshots**
Could not capture a good screenshot.

 - OS Build: 10.0.18942.0
 - Architecture: X64
 - Application Version: 10.1907.22.0
 - Region: en-US
 - Dev Version Installed: True

**Additional context**
Right-clicking again selects all text, so it appears this bug might only affect visuals.

**Requested Assignment**

praveensvsrk:
- Focused the TextBlock element when the event source is the TextBlock itself. This fixes the resultant number not gaining focus when a right-click is performed.

- Tested Manually


rudyhuyn:
This change can prevent `CalculationResult::OnGotFocus` to be called, which means that the control can't update its visual state.

**BUT** this is not a problem because none of the CalculationResult templates used by the app contains a visual state called "Focused", so your change is perfect! 

I will submit another PR to remove `OnGotFocus` and `OnLostFocus`.

Good job!! 👍

src/Calculator/Controls/CalculationResult.cpp
```diff
@@ -377,7 +377,16 @@ void CalculationResult::OnTapped(TappedRoutedEventArgs ^ e)
 
 void CalculationResult::OnRightTapped(RightTappedRoutedEventArgs ^ e)
 {
-    this->Focus(::FocusState::Programmatic);
+    auto requestedElement = e->OriginalSource;
+
+    if (requestedElement->Equals(safe_cast<Object ^>(m_textBlock)))
```
joseartrivera: use dynamic_cast instead since m_textBlock could technically be nullptr if a style is applied to this control that doesn't; include it.

praveensvsrk: Makes sense! I will do the change.

PR #700, opened 2019-10-04
remove the useless code related to the management of the focus in CalculationResult

### LINKED ISSUE #699: Code cleanup: remove the code related to the management of the focus in CalculationResult
**Describe the code cleanup**
`CalculationResult` implements 3 methods to manage the focus states of the control: 
- OnLostFocus
- OnGotFocus
- OnRightTapped

But none of the templates used by `CalculationResult` contains the visual states "Focused"/"Unfocused", meaning that these 3 methods are meaningless.

 - OS Build: 10.0.18363.0
 - Architecture: X64
 - Region: en-US
 - Dev Version Installed: True

**Requested Assignment**

rudyhuyn:
- remove OnLostFocus, OnFocus and OnRightTapped
- remove `CalculationResult::s_FocusedState` and `CalculationResult::s_UnfocusedState`
- stop forcing the focus when users tap the control

- manually


mcooley:
I added back the explicit call to Focus() in OnTapped() since it is a change in behavior from the current app. We can remove it separately from this code cleanup if it's the right thing to do.

PR #702, opened 2019-10-04
Active tracing

DavidShoe:
-
-
-

-
-
-


sanderl:
> > Should tracing only be disabled using the toggle button? Right now it is also disabled when clicking in the equation edit box.
> 
> You can break out of active tracing by doing other actions like adding a new equation or editing an existing one.

When I used it, I thought it was weird that only clicking in the EquationEditBox would toggle off focus and not clicking the gray area below the EquationEditBox. I think it would be best to just use the toggle button so that a stray click cannot accidentally turn it off.

sanderl:
I'm going to approve this in because it fixes the toggle button issue. I think we should file a bug to track if a click outside the graph area should turn off tracing.

PR #704, opened 2019-10-07
Key graph features

sanderl:
- This is a working implementation of Key Graph Features for Graphing Calculator. It is not complete.
- Remaining work: Hooking up the KGF EquationEditBox, Error Handling, Rich Edit Styling, Fit and Finish


src/Calculator/Calculator.vcxproj
```diff
@@ -224,6 +223,12 @@
   </ItemDefinitionGroup>
   <PropertyGroup>
     <AppVersion Condition="'$(AppVersion)' == ''">0.0.0.0</AppVersion>
+    <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
```
sanderl: I will remove this section.

src/Calculator/Calculator.vcxproj.filters
```diff
@@ -1482,5 +1487,21 @@
   <ItemGroup>
     <CopyFileToFolders Include="$(GraphingImplDll)" />
     <CopyFileToFolders Include="$(GraphingEngineDll)" />
+    <CopyFileToFolders Include="$(GraphingImplDll)" />
```
sanderl: removing

src/Calculator/Package.appxmanifest
```diff
@@ -1,14 +1,14 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10" xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest" xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10" xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5" IgnorableNamespaces="uap uap5 mp">
-    <Identity Name="Microsoft.WindowsCalculator.Graphing" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.0.0.0" />
+    <Identity Name="Microsoft.WindowsCalculator.Graphing" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.0.5.0" />
```
sanderl: I'll revert this

src/Calculator/Package.appxmanifest
```diff
@@ -1,14 +1,14 @@
...
     <mp:PhoneIdentity PhoneProductId="b58171c6-c70c-4266-a2e8-8f9c994f4456" PhonePublisherId="95d94207-0c7c-47ed-82db-d75c81153c35" />
     <Properties>
         <DisplayName>ms-resource:DevAppStoreName</DisplayName>
         <PublisherDisplayName>Microsoft Corporation</PublisherDisplayName>
         <Logo>Assets\CalculatorStoreLogo.png</Logo>
     </Properties>
     <Dependencies>
-        <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17133.0" MaxVersionTested="10.0.18362.0" />
+        <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.18362.0" MaxVersionTested="10.0.18362.0" />
```
sanderl: reverting

src/GraphControl/GraphControl.vcxproj.filters
```diff
@@ -74,4 +74,10 @@
   <ItemGroup>
     <None Include="$(ProjectDir)\GraphingImplOverrides.props" />
   </ItemGroup>
+  <ItemGroup>
+    <CopyFileToFolders Include="$(GraphingImplDll)" />
+    <CopyFileToFolders Include="$(GraphingEngineDll)" />
+    <CopyFileToFolders Include="$(GraphingImplDll)" />
```
sanderl: will remove dupes

src/MockGraphingImpl/MockGraphingImpl.vcxproj
```diff
@@ -42,8 +42,8 @@
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
-    <WindowsTargetPlatformVersion>10.0.18970.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformMinVersion>10.0.18362.0</WindowsTargetPlatformMinVersion>
+    <WindowsTargetPlatformVersion>10.0.18980.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.18980.0</WindowsTargetPlatformMinVersion>
```
sanderl: will revert min version

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +436,108 @@ namespace GraphControl
...
+            }
+        }
+
+        return nullptr;
+    }
+
+    void Grapher::UpdateKeyGraphFeatures()
+    {
+        auto equations = GetValidEquations();
+        for (auto equation : GetValidEquations())
```
joseartrivera: Don't we only need to do this for the equation that had it's button pressed?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +436,108 @@ namespace GraphControl
...
+        {
+            if (auto graph = GetGraph(equation))
+            {
+                if (auto analyzer = graph->GetAnalyzer())
+                {
+                    bool result;
+                    analyzer->CanFunctionAnalysisBePerformed(result);
+                    if (result)
+                    {
+                        m_solver->FormatOptions().SetFormatType(FormatType::MathML);
```
joseartrivera: This is the same solver we use for the main graph right? We should just set this on initialization instead of in here.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +436,108 @@ namespace GraphControl
...
+                    else
+                    {
+                        equation->AnalysisNotSupported = true;
+                    }
+                }
+            }
+        }
+    }
+
+    Windows::Foundation::Collections::IObservableVector<Platform::String ^> ^ Grapher::ConvertWStringVector(vector<wstring> inVector)
```
joseartrivera: You can remove namespaces in the cpp files to make it cleaner, here and below/above

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.cpp
```diff
@@ -0,0 +1,282 @@
...
+            PeriodicityTextBlock->Text = m_resourceLoader->GetString(L"Periodicity");
+            MinimaTextBlock->Text = m_resourceLoader->GetString(L"Minima");
+            MaximaTextBlock->Text = m_resourceLoader->GetString(L"Maxima");
+            InflectionPointsTextBlock->Text = m_resourceLoader->GetString(L"InflectionPoints");
+            VerticalAsymptotesTextBlock->Text = m_resourceLoader->GetString(L"VerticalAsymptotes");
+            HorizontalAsymptotesTextBlock->Text = m_resourceLoader->GetString(L"HorizontalAsymptotes");
+            ObliqueAsymptotesTextBlock->Text = m_resourceLoader->GetString(L"ObliqueAsymptotes");
+            MonotonicityTextBlock->Text = m_resourceLoader->GetString(L"Monotonicity");
+
+            Domain = ViewModel->Domain;
```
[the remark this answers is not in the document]
sanderl: No I didn't directly bind the viewmodel to the UI so that I could substitute a the value with a string if nothing was returned for that property.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.cpp
```diff
@@ -0,0 +1,282 @@
...
+                Monotonicity += m_resourceLoader->GetString(L"KGFMonotonicityUnknown");
+                break;
+            case '1':
+                Monotonicity += m_resourceLoader->GetString(L"KGFMonotonicityIncreasing");
+                break;
+            case '2':
+                Monotonicity += m_resourceLoader->GetString(L"KGFMonotonicityDecreasing");
+                break;
+            case '3':
+                Monotonicity += m_resourceLoader->GetString(L"KGFMonotonicityConstant");
```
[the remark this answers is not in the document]
sanderl: This is basically building in a string that has 2 parts. This first is a mathml string, which I cannot localize and the second is just a direction. Output looks like [0, 1] increasing. However, it could also be Increasing [0, 1], so I don't think we need to change the order in this case.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.cpp
```diff
@@ -0,0 +1,282 @@
...
+String ^ KeyGraphFeaturesPanel::ConvertVectorToString(Windows::Foundation::Collections::IObservableVector<String ^> ^ inVector)
+{
+    String ^ outString = ref new String();
+    int vectorSize = inVector->Size;
+    for (int i = 0; i < vectorSize; i++)
+    {
+        outString += inVector->GetAt(i);
+        if (i < vectorSize - 1)
+        {
+            outString += "\n";
```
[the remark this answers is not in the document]
sanderl: In this case since I am using a new line, it will be ok. Using a ListView with MathRichEditBoxes will be much more expensive than using the new line character.

src/GraphControl/Control/Grapher.cpp
sanderl: hmm...moving it unset the value. I'll dig in.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.cpp
```diff
@@ -0,0 +1,282 @@
...
+                                                                        : m_resourceLoader->GetString(L"KGFObliqueAsymptotesNone");
+            SetMonotoncityStringProperty();
+            SetTooComplexFeaturesErrorProperty();
+        }
+    }
+}
+
+void KeyGraphFeaturesPanel::SetParityStringProperty()
+{
+    if (ViewModel->Parity)
```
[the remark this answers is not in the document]
sanderl: Good catch, I also fixed this for Periodicity and Monotonicity.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
```diff
@@ -0,0 +1,96 @@
...
+             mc:Ignorable="d">
+
+
+    <UserControl.Resources>
+        <Style x:Name="KGF_RichEditBoxStyle" TargetType="RichEditBox">
+            <Setter Property="Margin" Value="0,8,3,0"/>
+            <Setter Property="FontSize" Value="14"/>
+            <Setter Property="BorderThickness" Value="0"/>
+            <Setter Property="IsReadOnly" Value="True"/>
+            <Setter Property="Background" Value="Transparent"/>
```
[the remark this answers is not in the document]
sanderl: I tested it and it looked find to me.

src/GraphControl/Control/Grapher.cpp
sanderl: Yes, I'll refactor it to only do analysis when the equation button is clicked after we refactor the equation class.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.h
```diff
@@ -0,0 +1,85 @@
...
+    };
+
+public
+    ref class KeyGraphFeaturesPanel sealed : public Windows::UI::Xaml::Data::INotifyPropertyChanged
+    {
+    public:
+        KeyGraphFeaturesPanel();
+
+        OBSERVABLE_OBJECT_CALLBACK(OnPropertyChanged);
+        OBSERVABLE_PROPERTY_R(Platform::String ^, Domain);
```
joseartrivera: All of these observable properties already exist in the EquationViewModel. This control should only take a ViewModel and bind to the properties on the view model. The ViewModel can handle the logic to return an error message or not, the control should just cleanly bind to each property.

src/GraphControl/Control/Equation.h
```diff
@@ -57,17 +65,330 @@ namespace GraphControl
             }
             void set(Windows::UI::Xaml::Media::SolidColorBrush^ value)
             {
-                if (value == nullptr || LineColor == nullptr || (value->Color.A != LineColor->Color.A) || (value->Color.R != LineColor->Color.R) || (value->Color.G != LineColor->Color.G)
```
joseartrivera: You should add this logic back

src/GraphControl/Control/Grapher.cpp
```diff
@@ -1,5 +1,6 @@
 #include "pch.h"
 #include "Grapher.h"
+#include <IBitmap.h>
```
joseartrivera:
```suggestion
#include "IBitmap.h"
```

src/GraphControl/Control/Grapher.cpp
```diff
@@ -404,6 +408,7 @@ namespace GraphControl
                         m_renderMain->Graph = m_graph;
 
                         UpdateVariables();
+                        UpdateKeyGraphFeatures();
```
joseartrivera: Could we delay this until the user clicks the button? I know we can't detect the specific equation yet, but it would be good to minimize the performance hit when we initially display the graph as much as possible.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +439,105 @@ namespace GraphControl
...
+                return graph;
+            }
+        }
+
+        return nullptr;
+    }
+
+    void Grapher::UpdateKeyGraphFeatures()
+    {
+        auto equations = GetValidEquations();
```
joseartrivera: You are calling GetValidEquations twice

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +439,105 @@ namespace GraphControl
...
+
+                                equation->Minima = ConvertWStringVector(functionAnalysisData.Minima);
+                                equation->Maxima = ConvertWStringVector(functionAnalysisData.Maxima);
+                                equation->InflectionPoints = ConvertWStringVector(functionAnalysisData.InflectionPoints);
+                                equation->Monotonicity = ConvertWStringIntMap(functionAnalysisData.MonotoneIntervals);
+                                equation->VerticalAsymptotes = ConvertWStringVector(functionAnalysisData.VerticalAsymptotes);
+                                equation->HorizontalAsymptotes = ConvertWStringVector(functionAnalysisData.HorizontalAsymptotes);
+                                equation->ObliqueAsymptotes = ConvertWStringVector(functionAnalysisData.ObliqueAsymptotes);
+                                equation->TooComplexFeatures = functionAnalysisData.TooComplexFeatures;
+                                equation->AnalysisError = (int)AnalysisErrorType::NoError;
```
joseartrivera:
```suggestion
                                equation->AnalysisError = static_cast<int>(AnalysisErrorType::NoError);
```

Use static_cast/dynamic_cast instead here and in the rest of the code

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +439,105 @@ namespace GraphControl
...
+                        {
+                            Graphing::IGraphFunctionAnalysisData functionAnalysisData = m_solver->Analyze(analyzer.get());
+                            {
+                                equation->XIntercept = ref new String(functionAnalysisData.Zeros.c_str());
+                                equation->YIntercept = ref new String(functionAnalysisData.YIntercept.c_str());
+                                equation->Domain = ref new String(functionAnalysisData.Domain.c_str());
+                                equation->Range = ref new String(functionAnalysisData.Range.c_str());
+                                equation->Parity = functionAnalysisData.Parity;
+
+                                // There is only ever one item in Periodicity. Map is being used because it works with DependencyProperty.
```
joseartrivera: Could we simplify this by just having a property for the periodicity and the direction?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +439,105 @@ namespace GraphControl
...
+                    }
+                }
+            }
+
+            equation->AnalysisError = (int)AnalysisErrorType::AnalysisCouldNotBePerformed;
+        }
+    }
+    IObservableVector<String ^> ^ Grapher::ConvertWStringVector(vector<wstring> inVector)
+    {
+        Vector<Platform::String ^> ^ outVector = ref new Vector<String ^>();
```
joseartrivera:
```suggestion
        Vector<String ^> ^ outVector = ref new Vector<String ^>();
```

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +439,105 @@ namespace GraphControl
...
+        ;
```
joseartrivera: nit: extra semicolon

src/GraphControl/Control/Grapher.cpp
```diff
@@ -433,6 +439,105 @@ namespace GraphControl
...
+        for (auto v : inVector)
+        {
+            outVector->Append(ref new String(v.c_str()));
+        }
+
+        return outVector;
+    }
+    IObservableMap<String ^, String ^> ^ Grapher::ConvertWStringIntMap(map<wstring, int> inMap)
```
joseartrivera: nit: add space between functions

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -391,21 +421,27 @@
                 <RowDefinition Height="5*"/>
                 <RowDefinition Height="3*"/>
             </Grid.RowDefinitions>
-
-            <Grid>
+            <local:KeyGraphFeaturesPanel x:Name="KeyGraphFeaturesControl"
```
joseartrivera: Since this is pretty expensive with all of the richeditboxes, we should consider setting [x:load = false](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/x-load-attribute)

src/GraphingInterfaces/IGraph.h
```diff
@@ -16,12 +17,14 @@ namespace Graphing
...
+        virtual std::vector<std::shared_ptr<Graphing::IVariable>> GetVariables() = 0;
 
         virtual void SetArgValue(std::wstring variableName, double value) = 0;
 
-        virtual std::shared_ptr< Renderer::IGraphRenderer > GetRenderer() const = 0;
+        virtual std::shared_ptr<Renderer::IGraphRenderer> GetRenderer() const = 0;
 
         virtual bool TryResetSelection() = 0;
+
+		virtual std::shared_ptr< Graphing::Analyzer::IGraphAnalyzer > GetAnalyzer() const = 0;
```
joseartrivera: nit: whitespace

src/Calculator/Views/Calculator.xaml
```diff
@@ -680,8 +695,8 @@
                   x:Uid="MemoryPanel"
                   Grid.Row="4"
                   Margin="3,0,3,0"
-                  Visibility="{x:Bind Model.IsAlwaysOnTop, Converter={StaticResource BooleanToVisibilityNegationConverter}, Mode=OneWay}"
-                  AutomationProperties.HeadingLevel="Level1">
+                  AutomationProperties.HeadingLevel="Level1"
+                  Visibility="{x:Bind Model.IsAlwaysOnTop, Converter={StaticResource BooleanToVisibilityNegationConverter}, Mode=OneWay}">
                 <Grid.ColumnDefinitions>
                     <ColumnDefinition Width="1*" MaxWidth="80"/>
                     <ColumnDefinition Width="1*" MaxWidth="80"/>
```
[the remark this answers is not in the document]
sanderl: Just xaml styler.

src/Calculator/App.xaml
```diff
@@ -1618,13 +1618,245 @@
...
+                                    <Button.Resources>
+                                        <SolidColorBrush x:Name="ButtonBackgroundPointerOver"
+                                                         Opacity="0.5"
+                                                         Color="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=EquationColor.Color}"/>
+                                        <SolidColorBrush x:Name="ButtonForegroundPointerOver" Color="{ThemeResource SystemChromeWhiteColor}"/>
+                                        <SolidColorBrush x:Name="ButtonBorderBrushPointerOver" Color="Transparent"/>
+                                    </Button.Resources>
+
+                                </Button>
+                                <Border x:Name="EquationBoxBorder"
```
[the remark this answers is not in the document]
sanderl: No it doesn't I removed it.

src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void MathRichEditBox::SetMathTextProperty(String ^ newValue)
+{
+    bool readOnlyState = this->IsReadOnly;
+    this->IsReadOnly = false;
+
+    // Check if this is a MathML string. If it is, then use SetMath so the math text displays properly.
+    wstring value = newValue->Data();
+    if (value.find(L"<mml:math") == 0 || value.find(L"<math") == 0)
```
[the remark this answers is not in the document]
sanderl: yes it would, i can look for the full header tag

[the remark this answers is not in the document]
sanderl: That doesn't really fix the issue, let me think on how to solve this problem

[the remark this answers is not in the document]
joseartrivera: That would have the same problem I think, does the caller know what type it would be? They could set it. I think a better design might be to just use a TextBlock if it is plain text anyway, would be lighter weight too.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.cpp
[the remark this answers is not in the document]
sanderl: Since we can show the information directly from the engine or a string provided by the UI, I decided not to directly bind to the xaml so that I could swap out the data more easily. Domain and Range or the only two that do not have an error string associated with them, so I decided not to bind them to the xaml directly just to be consistent with the other controls.

The xaml binding is to the properties in the KeyGraphFeaturesPanel.xaml.h file.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.h
sanderl: I can move the logic. and then just bind directly.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml.cpp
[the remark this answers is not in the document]
sanderl: Just saw your next comment, I'll move the logic to the viewmodel and bind directly. Thanks for the suggestion.

src/GraphControl/Control/Grapher.cpp
sanderl: Yes, but it would be best to do this after the code is refactored to use an equation model. I haven't noticed much of a performance hit in release with the code this way.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
```diff
@@ -0,0 +1,160 @@
...
+            </StackPanel>
+        </DataTemplate>
+
+        <ts:KeyGraphFeaturesTemplateSelector x:Key="KGFTemplateSelector"
+                                             GridTemplate="{StaticResource KGFGridDataTemplate}"
+                                             RichEditTemplate="{StaticResource KGFRichEditDataTemplate}"
+                                             TextBlockTemplate="{StaticResource KGFTextBlockDataTemplate}"/>
+
+        <converters:BooleanToVisibilityNegationConverter x:Name="BooleanToVisibilityNegationConverter"/>
+        <converters:ItemSizeToVisibilityConverter x:Name="ItemSizeToVisibilityConverter"/>
```
joseartrivera: I think this is unused

src/Calculator/App.xaml
```diff
@@ -1618,13 +1615,209 @@
...
+                                        Foreground="{StaticResource SystemChromeWhiteColor}">
+                                    <Button.Content>
+                                        <StackPanel Orientation="Horizontal">
+                                            <FontIcon FontFamily="{ThemeResource SymbolThemeFontFamily}" Glyph="&#xE72B;"/>
+                                            <TextBlock FontFamily="{TemplateBinding FontFamily}" Text=" ƒₓ"/>
+                                        </StackPanel>
+                                    </Button.Content>
+                                    <Button.Resources>
+                                        <SolidColorBrush x:Name="ButtonBackgroundPointerOver"
+                                                         Opacity="0.4"
```
joseartrivera: Opacity of 0.4 feels a little too washed out, how does 0.6 look?

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -3527,8 +3527,163 @@
     <value>Max</value>
     <comment>Label text for the max text box</comment>
   </data>
-  <data name="LineColorText.Text" xml:space="preserve">
```
joseartrivera: Add this back

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -1,14 +1,316 @@
...
+        , m_AnalysisError{ 0 }
+        , m_AnalysisErrorVisible{ false }
+        , m_KeyGraphFeaturesItems{ ref new Vector<KeyGraphFeaturesItem ^>() }
+        , m_resourceLoader{ Windows::ApplicationModel::Resources::ResourceLoader::GetForCurrentView() }
+    {
+    }
+
+    void EquationViewModel::OnPropertyChanged(String ^ propertyName)
+    {
+        if (propertyName == L"IsAnalysisUpdated" && IsAnalysisUpdated)
```
joseartrivera: I think it would be cleaner to simplify this if statement to if (propertyName == L"IsAnalysisUpdated") and call a method called OnIsAnalysisUpdatedChanged()

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -1,14 +1,316 @@
...
+
+    void EquationViewModel::SetParityStringProperty()
     {
+        KeyGraphFeaturesItem ^ parityItem = ref new KeyGraphFeaturesItem();
+        parityItem->Title = m_resourceLoader->GetString(L"Parity");
+        switch (Parity)
+        {
+        case 0:
+            parityItem->DisplayItems->Append(m_resourceLoader->GetString(L"KGFParityUnknown"));
+            parityItem->IsText = true;
```
joseartrivera: You can factor out the partyItem->IsText = true; to outside the switch statement.

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -1,14 +1,316 @@
...
+            break;
+        default:
+            parityItem->DisplayItems->Append(m_resourceLoader->GetString(L"KGFParityUnknown"));
+            parityItem->IsText = true;
+        }
+
+        KeyGraphFeaturesItems->Append(parityItem);
+    }
+    void EquationViewModel::SetPeriodicityStringProperty()
```
joseartrivera: nit: spacing

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -1,14 +1,316 @@
...
+        }
+
+        Platform::String ^ separator = ref new String(LocalizationSettings::GetInstance().GetListSeparator().c_str());
+
+        wstring error;
+        if ((TooComplexFeatures & KeyGraphFeaturesFlag::Domain) == KeyGraphFeaturesFlag::Domain)
+        {
+            error.append((m_resourceLoader->GetString(L"Domain") + separator + L" ")->Data());
+        }
+        else if ((TooComplexFeatures & KeyGraphFeaturesFlag::Range) == KeyGraphFeaturesFlag::Range)
```
joseartrivera: Shouldn't all of these be just if statements instead of else if? Otherwise you only ever capture one feature that is too complex. Also out of curiosity what type of equation is too complex? I only ever get the unsupported error message.

src/CalcViewModel/GraphingCalculator/EquationViewModel.h
```diff
@@ -4,14 +4,93 @@
...
-        OBSERVABLE_PROPERTY_RW(Windows::UI::Xaml::Visibility, KeyGraphFeaturesVisibility);
+
+        // Key Graph Features
+        OBSERVABLE_PROPERTY_RW(bool, IsAnalysisUpdated);
+        OBSERVABLE_PROPERTY_RW(Platform::String ^, Domain);
+        OBSERVABLE_PROPERTY_RW(Platform::String ^, Range);
+        OBSERVABLE_PROPERTY_RW(Platform::String ^, XIntercept);
+        OBSERVABLE_PROPERTY_RW(Platform::String ^, YIntercept);
+        OBSERVABLE_PROPERTY_RW(int, Parity);
+        // OBSERVABLE_PROPERTY_R(Platform::String ^, ParityString);
```
joseartrivera: You can remove the commented out code

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
```diff
@@ -0,0 +1,160 @@
...
+                    <ItemsControl.ItemTemplate>
+                        <DataTemplate x:DataType="x:String">
+                            <TextBlock Style="{StaticResource KGF_TextBlockStyle}" Text="{x:Bind}"/>
+                        </DataTemplate>
+                    </ItemsControl.ItemTemplate>
+                </ItemsControl>
+            </StackPanel>
+        </DataTemplate>
+
+        <ts:KeyGraphFeaturesTemplateSelector x:Key="KGFTemplateSelector"
```
joseartrivera: Great job with this! Much better UI design 🎉

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
sanderl: good catch.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -392,20 +424,29 @@
                 <RowDefinition Height="3*"/>
             </Grid.RowDefinitions>
 
-            <Grid>
+            <!-- Ideally the KeyGraphFeaturesPanel should be a frame so that navigation to and from the panel could be handled nicely -->
+            <local:KeyGraphFeaturesPanel x:Name="KeyGraphFeaturesControl"
```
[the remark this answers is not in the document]
sanderl: I'm going to try to repro these MathRichEditBox issues in a sample app to try to determine if this is a MathRichEditBox issue or a RichEditBox issue.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
```diff
@@ -0,0 +1,160 @@
+<UserControl x:Class="CalculatorApp.KeyGraphFeaturesPanel"
```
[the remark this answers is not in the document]
sanderl: @grochocki, the comps show the error text to look the same as the expressions, should these look different?

src/Calculator/App.xaml
```diff
@@ -1152,260 +1152,258 @@
...
+                                           FontStyle="Normal"
+                                           AutomationProperties.AccessibilityView="Raw"
+                                           Text="{TemplateBinding Content}"/>
+                            </Grid>
+                        </ControlTemplate>
+                    </Setter.Value>
+                </Setter>
+            </Style>
+
+            <Style x:Key="MathRichEdit_EquationTextBoxStyle" TargetType="Controls:MathRichEditBox">
```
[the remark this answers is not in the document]
sanderl: I confirmed with Eric this is a RichEdit bug.

src/Calculator/App.xaml
```diff
@@ -1152,260 +1152,258 @@
...
+                                                <ObjectAnimationUsingKeyFrames Storyboard.TargetName="PlaceholderTextContentPresenter" Storyboard.TargetProperty="Foreground">
+                                                    <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource TextControlPlaceholderForegroundDisabled}"/>
+                                                </ObjectAnimationUsingKeyFrames>
+                                            </Storyboard>
+                                        </VisualState>
+                                        <VisualState x:Name="Normal"/>
 
-                                                        <VisualStateManager.VisualStateGroups>
-                                                            <VisualStateGroup x:Name="CommonStates">
+                                        <VisualState x:Name="PointerOver">
```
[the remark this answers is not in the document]
sanderl: There is a SelectionHighlightColorWhenNotFocused property that is causing this to select. Unfortunately setting it to Transparent does not appear to be working. Also trying to deselect when the REB loses focuses doesn't work all the time either. I'm going to file a bug to fix this later.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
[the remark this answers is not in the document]
sanderl: I believe this is related to the SelectionWhenNotFocused highlighting from the above comment. I'll make sure to note this behavior in the bug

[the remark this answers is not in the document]
sanderl: @grochocki, do you want to disable the button or show an error string in this case?

src/Calculator/App.xaml
```diff
@@ -1618,13 +1618,245 @@
...
+                <Setter Property="IsTabStop" Value="False"/>
+                <Setter Property="Typography.StylisticSet20" Value="True"/>
+                <Setter Property="Template">
+                    <Setter.Value>
+                        <ControlTemplate TargetType="Controls:EquationTextBox">
+                            <Grid>
+
+                                <Grid.Resources>
+
+                                    <Style x:Key="EquationTextBoxStyle" TargetType="Controls:MathRichEditBox">
```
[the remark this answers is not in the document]
sanderl: I've moved all of the KGF specific styles to the KeyGraphFeaturesPanel.xaml since it is not expected they will be used elsewhere.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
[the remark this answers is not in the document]
sanderl: I talked with Dave and he said not to change the way the strings look at this time.

src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -0,0 +1,66 @@
...
+using namespace Windows::Foundation::Collections;
+
+DEPENDENCY_PROPERTY_INITIALIZATION(MathRichEditBox, MathText);
+
+MathRichEditBox::MathRichEditBox()
+{
+    static LimitedAccessFeatureStatus m_lafResultStatus;
+    String ^ packageName = Package::Current->Id->Name;
+
+    if (packageName == L"Microsoft.WindowsCalculator.Dev")
```
joseartrivera: We should remove these from running code and use ifdefs instead

src/GraphControl/Control/Grapher.cpp
```diff
@@ -73,6 +83,8 @@ namespace GraphControl
         auto cw = CoreWindow::GetForCurrentThread();
         cw->KeyDown += ref new TypedEventHandler<CoreWindow ^, KeyEventArgs ^>(this, &Grapher::OnCoreKeyDown);
         cw->KeyUp += ref new TypedEventHandler<CoreWindow ^, KeyEventArgs ^>(this, &Grapher::OnCoreKeyUp);
+
+                auto& formatOptions = m_solver->FormatOptions();
```
joseartrivera: Remove this

src/GraphControl/Control/Grapher.cpp
```diff
@@ -370,14 +382,15 @@ namespace GraphControl
 
     void Grapher::UpdateGraph()
     {
+        auto& formatOptions = m_solver->FormatOptions();
```
joseartrivera: left over debug code? Here and on line 414?

PR #710, opened 2019-10-11
fix 505

### LINKED ISSUE #505: Update CALC_E_* to no longer use macros
**Problem Statement**
One of the goals for this project appears to be converting legacy C style code to modern C++.  Avoiding macros unless completely necessary has been a C++ best practice for quite some time.

**Proposal**
With more modern features such as constexpr values, you should be able to replace all CALC_E_* macros with constexpr values.  Preferably, you could replace them with an enum class, but that is riskier and would require much more code refactoring.

**Goals**
Modernize the code base and make it easier to maintain.

Grommers00:
- Changed all the E_CODE_* Variables located in CalcErr.h from Macros to static const expr (inline is implicite)

- Manual-Adhoc testing

### DIFF
--- a/src/CalcManager/Ratpack/CalcErr.h
+++ b/src/CalcManager/Ratpack/CalcErr.h
@@ -41,8 +41,8 @@ typedef int32_t ResultCode;
 // CALC_E_DIVIDEBYZERO
 //
 // The current operation would require a divide by zero to complete
-
 static constexpr uint32_t CALC_E_DIVIDEBYZERO = (uint32_t)0x80000000;
+
 // CALC_E_DOMAIN
 //
 // The given input is not within the domain of this function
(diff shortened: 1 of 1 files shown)


src/CalcManager/Ratpack/CalcErr.h
```diff
@@ -41,45 +41,45 @@ typedef int32_t ResultCode;
 // CALC_E_DIVIDEBYZERO
 //
 // The current operation would require a divide by zero to complete
-#define CALC_E_DIVIDEBYZERO ((uint32_t)0x80000000)
 
+static constexpr uint32_t CALC_E_DIVIDEBYZERO = (uint32_t)0x80000000;
```
HowardWolosky: Looks like the line feed separation got mixed up for this one.  The const should be connected to the comment like the others.

Grommers00:
I have updated the code to remove the white line space. 

Is there anything else I need to do?

PR #715, opened 2019-10-14
Add debug-generated files to .gitignore

Howard-C:
Those files are generated during debugging (actually cmake, but they're generated under "x64-Debug" folder) using the latest version of Visual Studio, so they should be added to .gitignore.


mcooley:
Hi @Howard-C, I haven't seen Visual Studio generate these files. Calculator builds with MSBuild, not CMake. Can you share more specifics about what you're doing which causes these files to be generated?

Howard-C:
I literally did nothing, I opened the repo in Visual Studio and CMake automatically began. Then these files were generated.

PR #723, opened 2019-10-21
Updating CalculatorHistory to use a default destructor.

### LINKED ISSUE #719: CalculatorHistory shouldn't call ClearHistory in its destructor
**Proposal**
CalculatorHistory's descructor is currently implemented as
```
CalculatorHistory::~CalculatorHistory(void)
{
    ClearHistory();
}

void CalculatorHistory::ClearHistory()
{
    m_historyItems.clear();
}
```
Here, `m_historyItems` is a `std::vector` which will handle cleaning up its own memory in its destructor.  This implementation was likely done at a time when CalculatorHistory wasn't using a `std::vector` and had to manage its own memory, but now that it is using a `std::vector`, having a descructor that calls `ClearHistory` is unnecessary code.

**Requested Assignment**

gsfreema:
Its memory will be cleaned up by std::vector's destructor.

- Removed CalculatorDestructor code so the compiler would default it.

- Manually tested.
- Ran existing unit tests

### DIFF
--- a/src/CalcManager/CalculatorHistory.cpp
+++ b/src/CalcManager/CalculatorHistory.cpp
@@ -69,11 +69,6 @@ shared_ptr<HISTORYITEM> const& CalculatorHistory::GetHistoryItem(_In_ unsigned i
     return m_historyItems.at(uIdx);
 }
 
-CalculatorHistory::~CalculatorHistory(void)
-{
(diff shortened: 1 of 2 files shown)


mcooley:
Thanks for cleaning this up!

PR #738, opened 2019-10-26
Make ControlSizeTrigger consistent with AdaptiveTrigger + fix conditional

### LINKED ISSUE #737: Make ControlSizeTrigger consistent with AdaptiveTrigger + bug fix
**Describe the bug**
Unlike AdaptiveTrigger, the visual state will never be triggered if you write the following code: `<trigger:ControlSizeTrigger MinWidth="0" />` which can be very useful for future PR.

Similarly, the current code doesn't work if the control associated is already fully loaded.

(the original control in https://github.com/microsoft/Windows-universal-samples/blob/master/Samples/XamlStateTriggers/cs/CustomTriggers/ControlSizeTrigger.cs has the same issues, which should also be fixed)

In addition to this improvement, we also need to fix a bug in the following code:
`if (Source != nullptr && MinWidth >= 0 || MinHeight >= 0)`
which is equivalent to 
`if ((Source != nullptr && MinWidth >= 0) || MinHeight >= 0)`
and not 
`if (Source != nullptr && (MinWidth >= 0 || MinHeight >= 0))`
which was, I suppose, the intention.

 - OS Build: 10.0.18363.0
 - Architecture: X64
 - Region: en-US
 - Dev Version Installed: True

**Requested Assignment**

rudyhuyn:
- fix conditional
- make `MinWidth` and `MinHeight` equal to -1 by default (same than AdaptiveTrigger)
- support Visual State trigger with `MinWidth="0"` or `MinHeight="0"`
- decrease the number of conditionals in `ControlSizeTrigger::UpdateIsActive`
- make ControlSizeTrigger works even if the control is already loaded (SizeChanged not called)

- manually with the Calculator + a test app

### DIFF
--- a/src/Calculator/Views/StateTriggers/ControlSizeTrigger.cpp
+++ b/src/Calculator/Views/StateTriggers/ControlSizeTrigger.cpp
@@ -23,58 +23,50 @@ ControlSizeTrigger::~ControlSizeTrigger()
     UnregisterSizeChanged(Source);
 }
 
-void ControlSizeTrigger::OnSourcePropertyChanged(FrameworkElement^ oldValue, FrameworkElement^ newValue)
+void ControlSizeTrigger::OnSourcePropertyChanged(FrameworkElement ^ oldValue, FrameworkElement ^ newValue)
 {
     UnregisterSizeChanged(oldValue);
     RegisterSizeChanged(newValue);
 }
 
-void ControlSizeTrigger::RegisterSizeChanged(FrameworkElement^ element)
+void ControlSizeTrigger::RegisterSizeChanged(FrameworkElement ^ element)
 {
-    if (element == nullptr) { return; }
+    if (element == nullptr)
+    {
+        return;
+    }
 
     if (element != Source)
     {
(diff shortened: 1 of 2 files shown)


src/Calculator/Views/StateTriggers/ControlSizeTrigger.cpp
```diff
@@ -51,30 +53,19 @@ void ControlSizeTrigger::UnregisterSizeChanged(FrameworkElement^ element)
...
-        else
-        {
-            SetActive(sourceSize.Width >= MinWidth);
-        }
+        SetActive(sourceSize.Height >= MinHeight && (MinWidth < 0 || sourceSize.Width >= MinWidth));
     }
     else
     {
-        SetActive(false);
+        SetActive(MinWidth >= 0 && sourceSize.Width >= MinWidth);
```
rudyhuyn: We can remove `MinWidth>=0` because sourceSize _should_ always be >= 0, but I prefer to keep it to be sure this assumption is and will always be true

PR #748, opened 2019-10-29
Taking into account Windows settings to disable/enable animations

### LINKED ISSUE #747: The calculator should react to changes in user settings to the enable/disable animations
**Describe the bug**
The Calculator caches the value of the setting "Show animations in Windows" and never checks it again during the lifetime of the instance.
**Steps To Reproduce**
- step 1: Launch the Calculator
- step 2: Switch from Standard mode to Scientific mode
- step 3: Go to settings>Display, change the value and wait ~5 seconds
- step 4: Switch from Standard mode to Scientific mode
--> Animation should be played in it was not the case in step 2 (or vice-versa)

**Expected behavior**
The calculator should react to user settings and change the behavior of animations

**Screenshots**

 - OS Build:
 - Architecture:
 - Application Version:
 - Region:
 - Dev Version Installed:

**Requested Assignment**
If possible, I would like to fix it

rudyhuyn:
## Fixes #747 

The Calculator caches the value of the setting "Show animations in Windows" and never checks it again during the lifetime of the instance. This PR modifies this behavior.

- remove all cached values of UISettings::AnimationsEnabled and check the value before playing a Storyboard.

- manually


src/Calculator/App.xaml.cpp
```diff
@@ -240,7 +230,6 @@ void App::OnAppLaunch(IActivatedEventArgs ^ args, String ^ argument)
     //#endif
 
     auto userSettings = ref new Windows::UI::ViewManagement::UISettings();
```
mcooley: It looks like` userSettings` is unused now.

PR #750, opened 2019-10-29
force a call to m_expressionContainer->Measure in UpdateScrollButtons

### LINKED ISSUE #749: these actions can crash the program
xxx

rudyhuyn:
## Fixes #749 

The UpdateScrollButtons method can modify the values of `m_expressionContainer->Padding' and `m_expressionContent->Margin', these modifications can launch the event `ScrollViewer::ViewChanged' which will call UpdateScrollButtons again. 

In the current version of the code, to prevent loops, first we unregister ViewChanged events, then we modify Padding and Margin and we register again afterwards.

Based on the issue reported by the user, this is obviously not enough, ViewChanged will still be called after the end of `UpdateScrollButtons' in some cases.

To solve this problem, we will call `ScrollViewer.Measure` to be sure ScrollViewer::ViewChanged will be called before we register ViewChanged event again.

- force a call to m_expressionContainer->Measure in UpdateScrollButtons so Scrollviewer::ViewChanged event will be called before we register again

- manually

### DIFF
--- a/src/Calculator/Controls/OverflowTextBlock.cpp
+++ b/src/Calculator/Controls/OverflowTextBlock.cpp
@@ -197,6 +197,7 @@ void OverflowTextBlock::UpdateScrollButtons()
             m_expressionContainer->Padding = Thickness(left, 0, right, 0);
             m_expressionContent->Margin = Thickness(-left, 0, -right, 0);
             m_expressionContainer->UpdateLayout();
+            m_expressionContainer->Measure(m_expressionContainer->RenderSize);
 
             m_containerViewChangedToken = m_expressionContainer->ViewChanged +=
                 ref new EventHandler<ScrollViewerViewChangedEventArgs ^>(this, &OverflowTextBlock::OnViewChanged);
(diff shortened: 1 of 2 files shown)


src/Calculator/Controls/OverflowTextBlock.cpp
```diff
@@ -197,6 +197,7 @@ void OverflowTextBlock::UpdateScrollButtons()
             m_expressionContainer->Padding = Thickness(left, 0, right, 0);
```
joseartrivera: Why exactly do we need to update the Padding/Margin here? Can't we just apply those to xaml directly?

rudyhuyn: it needs to be code-behind because we need to unregister to ViewChanged event first (to prevent UI cycle) and we need to call UpdateLayout and Measure after these values are set

joseartrivera: Makes sense, but why do we need to change the margin/padding in code-behind in the first place?

joseartrivera: Discussed in person

PR #752, opened 2019-10-29
Fully separate Calculator and CalcManager

### LINKED ISSUE #751: Calculator shouldn't use objects/methods from CalcEngine
To improve our architecture, we need to fully separate Calculator and CalcEngine and only use objects/methods from `CalcViewModel` in Calculator.

**Expected behavior**
If Calculator needs to access to a method/class from CalcManager, we should use CalcViewModel as a bridge between the 2 projects.

**Requested Assignment**

rudyhuyn:
## Fixes #751

The only type from CalcEngine used in Calculator is CalculatorVector. We shouldn't access/use this type of object in Calculator.

- Add a method `StandardCalculatorViewModel::SelectHistoryItem` to manage the selection of a history item in the CalcViewModel instead in Calculator.
- Make SetPrimaryDisplay, SetHistoryExpressionDisplay, SetExpressionDisplay and SetTokens private so CalculatorVector won't be used as a parameter of a public method.

- manually

### DIFF
--- a/src/CalcViewModel/Common/Utils.h
+++ b/src/CalcViewModel/Common/Utils.h
@@ -188,7 +188,7 @@ public:
 private:                                                                                                                                                       \
     static Windows::UI::Xaml::DependencyProperty ^ s_##n##Property;                                                                                            \
                                                                                                                                                                \
-public:
+private:
 
 // Utilities for DependencyProperties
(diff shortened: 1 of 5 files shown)


rudyhuyn:
`git rebase` to solve conflicts

PR #756, opened 2019-10-29
Replacing CalculatorVector usage with std::vector

### LINKED ISSUE #718: CalculatorVector<TType> should be replaced with std::vector
**Edit**
The original proposal below was to make CalculatorVector functions noexcept.  It was then suggested to remove it all together.  This issue is being repurposed for that.  The comments below are left for historical purposes.

**Proposal**
All of `CalculatorVector<TType>`'s functions return a `ResultCode` which indicates that they should not throw an exception.  Knowing this, the exception overhead for those functions can be reduced by making those functions noexcept.

It should be noted that based on the current implementation, `CalculatorVector<TType>` could still allow exceptions to escape.  However, I don't think this is intentional since all functions return a `ResultCode`.  An instance where one of these functions could throw would be in code like this:

```
    ResultCode SetAt(_In_ unsigned int index, _In_opt_ TType item)
    {
        try
        {
            m_vector[index] = item;
        }
        catch (const std::out_of_range& /*ex*/)
        {
            return E_BOUNDS;
        }
        return S_OK;
    }
```

If `TType` were to throw an exception during copy construction that isn't `std::out_of_range`, it would escape this function.  Since it is unlikely that `TType` will ever be anything that would throw in this case, it should be safe to make the function noexcept and have the application terminate if this unexpected behavior ever occurred.  It's highly unlikely that there is code catching whatever the copy constructor would be throwing anyway.

**Requested Assignment**

gsfreema:
- Removes CalculatorVector and replaces usage with std::vector
- Assumption made here are that memory allocations
are not recoverable.
- If it can be proved that an index
will be in range, then the indexing operation is used.
If not (without manual checks) the std::vector::at function
is used to throw an exception in case of a programmer bug.

- Manual testing was done
- Currently unable to run unit tests locally with the following errors.  Any help there would be greatly appreciated just in case someone has run into this before..

PRI175: 0x80070002 - Processing Resources failed with error: The system cannot find the file specified.	CalculatorUnitTests	C:\Users\gsfreema\Documents\GitHub\calculator\src\CalculatorUnitTests\GENERATEPROJECTPRIFILE	1

PRI252: 0xdef00071 - File C:\Users\gsfreema\Documents\GitHub\calculator\src\CalculatorUnitTests\x64\Debug\Microsoft.VisualStudio.TestTools.CppUnitTestFramework.CppUnitTestExtension\en-US\Microsoft.VisualStudio.TestTools.CppUnitTestFramework.CppUnitTestExtension.Resources.Resources.resw not found.	CalculatorUnitTests	C:\Users\gsfreema\Documents\GitHub\calculator\src\CalculatorUnitTests\GENERATEPROJECTPRIFILE	1


src/CalcManager/CalculatorHistory.cpp
```diff
@@ -7,25 +7,46 @@
...
 using namespace CalculationManager;
 
+namespace
+{
+    static wstring GetGeneratedExpression(const vector<pair<wstring, int>>& tokens)
+    {
+        wstring expression;
+        const size_t nTokens = tokens.size();
+
+        for (size_t i = 0; i < nTokens; ++i)
```
rudyhuyn: NIT: The original code was this way, but we can probably simplify it using `for (auto const& s : tokens)` instead

src/CalcViewModel/HistoryItemViewModel.cpp
```diff
@@ -27,37 +27,17 @@ HistoryItemViewModel::HistoryItemViewModel(
...
         _In_ String ^ fallbackExpression)
 {
     // updating accessibility names for expression and result
     wstringstream accExpression{};
     accExpression << L"";
 
-    unsigned int nTokens;
-    HRESULT hr = spTokens->GetSize(&nTokens);
-    if (SUCCEEDED(hr))
+    for (const auto& tokenItem : *spTokens)
```
rudyhuyn: a lot easier to read! 👍

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -319,64 +319,60 @@ void StandardCalculatorViewModel::SetHistoryExpressionDisplay(
...
     const auto& localizer = LocalizationSettings::GetInstance();
 
     const wstring separator = L" ";
     for (unsigned int i = 0; i < nTokens; ++i)
     {
-        if (SUCCEEDED(tokens->GetAt(i, &currentToken)))
+        auto currentToken = (*tokens)[i];
+        
+        Common::TokenType type;
+        bool isEditable = (currentToken.second == -1) ? false : true;
```
rudyhuyn: NIT: 
```
bool isEditable = currentToken.second != -1;
```

src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -1497,12 +1469,9 @@ void StandardCalculatorViewModel::Recalculate(bool fromHistory)
...
 CommandType StandardCalculatorViewModel::GetSelectedTokenType(_In_ unsigned int tokenPosition)
 {
-    pair<wstring, int> token;
-    shared_ptr<IExpressionCommand> tokenCommand;
-    IFTPlatformException(m_tokens->GetAt(tokenPosition, &token));
-
+    const pair<wstring, int>& token = m_tokens->at(tokenPosition);
     unsigned int tokenCommandIndex = token.second;
-    IFTPlatformException(m_commands->GetAt(tokenCommandIndex, &tokenCommand));
+    const shared_ptr<IExpressionCommand>& tokenCommand = m_commands->at(tokenCommandIndex);
```
rudyhuyn: We can probably remove `tokenCommandIndex` and directly use token.second.

src/CalcManager/CEngine/History.cpp
```diff
@@ -367,52 +371,41 @@ int CHistoryCollector::IchAddSzToEquationSz(wstring_view str, int icommandIndex)
...
     int minIdx = -1;
-    unsigned int nTokens = 0;
-    std::pair<std::wstring, int> currentPair;
-    m_spTokens->GetSize(&nTokens);
+    unsigned int nTokens = static_cast<unsigned int>(m_spTokens->size());
 
     for (unsigned int i = ich; i < nTokens; i++)
     {
-        IFT(m_spTokens->GetAt(i, &currentPair));
+        const auto& currentPair = (*m_spTokens)[i];
```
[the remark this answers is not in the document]
gsfreema: I was originally going to do that but noticed that `i` starts at `ich` so I left the loop alone.

src/CalcManager/CEngine/History.cpp
```diff
@@ -449,30 +436,25 @@ void CHistoryCollector::UpdateHistoryExpression(uint32_t radix, int32_t precisio
...
+            const std::shared_ptr<IExpressionCommand>& expCommand = m_spCommands->at(commandPosition);
+
             if (expCommand != nullptr && CalculationManager::CommandType::OperandCommand == expCommand->GetCommandType())
             {
-                std::shared_ptr<COpndCommand> opndCommand = std::static_pointer_cast<COpndCommand>(expCommand);
+                const std::shared_ptr<COpndCommand>& opndCommand = std::static_pointer_cast<COpndCommand>(expCommand);
                 if (opndCommand != nullptr)
                 {
                     token.first = opndCommand->GetString(radix, precision);
-                    IFT(m_spTokens->SetAt(i, token));
```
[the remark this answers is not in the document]
gsfreema: Nice catch.  Instead of setting a new copy of the token, would it make sense to have `token` declared as `auto&`.  Then any changes should be updated in place.

src/CalcManager/CalculatorHistory.cpp
gsfreema: I didn't update this to a range based for due to the check index check happening inside the loop so that it inserts a space after each token except for the last token.

src/CalcViewModel/StandardCalculatorViewModel.cpp
gsfreema: I'm cool with that change but I wanted to point out that tokenCommandIndex is also used on line 1348.  I wasn't sure if you all would like to keep better named variable around or to use `token.second` in both places.

src/CalcManager/CEngine/History.cpp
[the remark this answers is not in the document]
rudyhuyn: Yes, it would work!

src/CalcManager/CalculatorHistory.cpp
rudyhuyn: what do you think about this code?
```
bool isFirst = true;
for (auto const& token : tokens)
{
  if(isFirst)
  {
    isFirst = false;
    expression += L' ';
  }  
  expression.append(token.first);
}
```

src/CalcViewModel/StandardCalculatorViewModel.cpp
rudyhuyn: you can use token.second in both cases. Both are correct (and the compiler will generate the same code), so do what you like best!

src/CalcManager/CalculatorHistory.cpp
gsfreema: That works.  This is where I wish the C++ standard library had some sort of Join function.

src/CalcViewModel/StandardCalculatorViewModel.cpp
gsfreema: Cool, I'll do that.  I was just thinking for future code, it might not be a bad idea to used C++17's structured binding with these pairs.  I believe you can do that within the range for loops and have better names than `token.first` and `token.second`.

src/CalcManager/CalculatorHistory.cpp
rudyhuyn: Yes, I wrote the code from my phone, you are right, I was thinking about:
```
bool isFirst = true;
for (auto const& token : tokens)
{
  expression.append(token.first);
  if(isFirst)
  {
    isFirst = false;
    expression += L' ';
  }  
}
```

but your version works too! 

Based on old screenshots, it was also the case of the current operation part (above the result). We can keep it this way for the moment but I will open a new issue so we can chat about it.

image

PR #762, opened 2019-11-02
Make sure DateDifference::operator== compares DateTime::day

### LINKED ISSUE #761: DateDifference::operator== doesn't compare datetime::day values
**Describe the bug**
DateDifference::operator== fails to compare datetime::day values

```
bool operator==(const DateDifference^ dd) const
{
    return year == dd->year && month == dd->month && week == dd->week && day == day;
}
```

**Requested Assignment**

rudyhuyn:
- Fix DateDifference::operator== and compare `day` to `dd.day`

- manually


joseartrivera:
Good catch! Does this actually get used anywhere in code still?

PR #768, opened 2019-11-05
Replace wstring used in public methods by Platform::String in CalcViewModel

### LINKED ISSUE #757: Prefer Platform::String over wstring in CalcViewModel
**Describe the bug**
CalcViewModel currently uses wstring and Platform::String to represent strings. We should prefer Platform::String to make it easier to interact with XAML and WinRT api and reduce the number of conversions between the 2 formats in the Calculator project.

**Requested Assignment**

rudyhuyn:
- replace all wstring from public methods to `Platform::String`

also
- replace std::task by `IAsyncOperation`
- convert `LocalizationStringUtil` and `AppResourceProvider` to "ref classes" 
- use wstringstream when we concatenate many strings.

- manually


src/CalcViewModel/HistoryViewModel.cpp
```diff
@@ -164,7 +164,11 @@ void HistoryViewModel::OnClearCommand(_In_ Platform::Object ^ e)
             UpdateItemSize();
         }
 
-        MakeHistoryClearedNarratorAnnouncement(HistoryResourceKeys::HistoryCleared, m_localizedHistoryCleared);
+        if (m_localizedHistoryCleared == nullptr)
```
mcooley: Thanks, this pattern is much more clear than the previous code :)

PR #776, opened 2019-11-06
Updating StandardCalculatorViewModel to use algorithms

### LINKED ISSUE #775: StandardCalculatorViewModel code can be simplified with the use of standard algorithms
**Problem Statement**
There many instances where a find or a count is being done manually in StandardCalculatorViewModel instead of using existing algorithms in the STL.

**Evidence or User Insights**
In more object oriented focused language, people tend to use find and count algorithms because they are easily discovered through the collection object being used.  C++ has a different philosophy on algorithms in order to make them more generic and have better performance, but since they aren't part of collection classes, people tend to forget to use them.

**Proposal**
Start using std::find, std::count, etc. where appropriate to simplify code.

**Goals**
Simplify code base to make it easier to read and maintain.

**Requested Assignment**

gsfreema:
- Using default wstring constructor instead of taking empty string literal
- Updated to for range-for where appropriate
- Used std::find for IsOp* code that was doing it by hand
- Used std::count to calculate LengthWithoutPadding
- Used existing wstring constructor to pad a string

- Manual Testing

### DIFF
--- a/src/CalcViewModel/StandardCalculatorViewModel.cpp
+++ b/src/CalcViewModel/StandardCalculatorViewModel.cpp
@@ -1542,7 +1542,7 @@ bool StandardCalculatorViewModel::IsRecoverableCommand(int nOpCode)
 
 size_t StandardCalculatorViewModel::LengthWithoutPadding(wstring str)
 {
-    return std::count(str.begin(), str.end(), L' ');
+    return str.length() - count(str.begin(), str.end(), L' ');
 }
 
 wstring StandardCalculatorViewModel::AddPadding(wstring binaryString)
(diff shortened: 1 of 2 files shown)


src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -1558,29 +1534,15 @@ bool StandardCalculatorViewModel::IsRecoverableCommand(int nOpCode)
...
-    size_t count = 0;
-    for (size_t i = 0; i < str.length(); i++)
-    {
-        if (str[i] != L' ')
-        {
-            count++;
-        }
-    }
-    return count;
+    return std::count(str.begin(), str.end(), L' ');
```
rudyhuyn: This code counts the number of space characters whereas the original code does the contrary. 

Suggestion:
`return str.length() - count(...)`

rudyhuyn: You can remove `std::` too.

gsfreema: Oh man, I can't believe I missed that `!=` in the original code.  I sure am glad you caught that.

PR #782, opened 2019-11-06
Update share output

joseartrivera:
- Convert MathML to linear for sharing until we find a better way to share MathML
- Tweak the share output to be closer to the comps
- Remove the call to SetBitmap since it is just adding a duplicate image to the payload

- Manual tests

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -50,10 +50,10 @@
                 <graphControl:Grapher Name="GraphingControl"
                                       EquationsSource="{x:Bind ViewModel.Equations, Mode=OneWay}"
                                       ForceProportionalAxes="True"
-                                      LosingFocus="OnLosingFocus"
-                                      LostFocus="OnGraphLostFocus"
+                                      LosingFocus="GraphingControl_LosingFocus"
+                                      LostFocus="GraphingControl_LostFocus"
                                       UseSystemFocusVisuals="True"
-                                      VariablesUpdated="GraphVariablesUpdated">
+                                      VariablesUpdated="GraphingControl_VariablesUpdated">
                     <graphControl:Grapher.Background>
                         <SolidColorBrush Color="White"/>
                     </graphControl:Grapher.Background>
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -9,8 +9,7 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:vm="using:CalculatorApp.ViewModel"
              DataContextChanged="GraphingCalculator_DataContextChanged"
-             mc:Ignorable="d"
-             >
+             mc:Ignorable="d">
 
     <UserControl.Resources>
         <Style x:Key="ZoomRepeatButtonStyle" TargetType="RepeatButton">
@@ -33,7 +32,7 @@
         <converters:BooleanToVisibilityNegationConverter x:Name="BooleanToVisibilityNegationConverter"/>
     </UserControl.Resources>
 
-    <Grid x:Name="RootGrid" >
+    <Grid x:Name="RootGrid">
         <Grid.RowDefinitions>
             <RowDefinition x:Name="RowHamburger" Height="{StaticResource HamburgerHeightGridLength}"/>
             <RowDefinition/>
@@ -48,13 +47,13 @@
               Grid.Row="1"
               Grid.Column="0">
             <Grid Grid.Row="0" Margin="4,7,4,4">
-                <graphControl:Grapher Name="GraphingControl" 
-                                  EquationsSource="{x:Bind ViewModel.Equations, Mode=OneWay}"
-                                  ForceProportionalAxes="True"
-                                  UseSystemFocusVisuals="True"
-                                  VariablesUpdated="GraphVariablesUpdated"
-                                  LostFocus="OnGraphLostFocus"
-                                  LosingFocus="OnLoosingFocus">
+                <graphControl:Grapher Name="GraphingControl"
+                                      EquationsSource="{x:Bind ViewModel.Equations, Mode=OneWay}"
+                                      ForceProportionalAxes="True"
+                                      LosingFocus="OnLosingFocus"
+                                      LostFocus="OnGraphLostFocus"
+                                      UseSystemFocusVisuals="True"
+                                      VariablesUpdated="GraphVariablesUpdated">
                     <graphControl:Grapher.Background>
                         <SolidColorBrush Color="White"/>
                     </graphControl:Grapher.Background>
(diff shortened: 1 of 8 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -105,79 +109,100 @@ void CalculatorApp::GraphingCalculator::OnShareClick(Platform::Object ^ sender,
...
+    auto resourceLoader = ResourceLoader::GetForCurrentView();
+
     try
     {
-        // Get our title from the localized resources
-        auto EmptyEquationString = resourceLoader->GetString(L"EmptyEquationString");
-
-        std::wstring rawHtml = L"<p><img src='graph.png'></p>";
+        std::wstring rawHtml;
+        std::wstring graphImageHtml = L"<p><img src='graph.png' width='600'></p>";
```
rudyhuyn: We should add a `alt` attribute to make it accessible

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -105,79 +109,100 @@ void CalculatorApp::GraphingCalculator::OnShareClick(Platform::Object ^ sender,
...
-        for (unsigned i = 0; i < equations->Size; i++)
-        {
-            auto expression = equations->GetAt(i)->Expression->Data();
-            auto color = equations->GetAt(i)->LineColor->Color;
+        bool hasEquations = false;
 
-            if (equations->GetAt(i)->Expression->Length() == 0)
+        if (equations->Size > 0)
+        {
+            rawHtml += L"<span style=\"color: rgb(68, 114, 196); font-family: calibri, sans-serif; font-style: bold; font-size : 13pt;\">";
```
rudyhuyn: should we force the font instead of using the default font chosen by the user in Outlook/Mail?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -105,79 +109,100 @@ void CalculatorApp::GraphingCalculator::OnShareClick(Platform::Object ^ sender,
...
+            rawHtml += resourceLoader->GetString(L"EquationsShareHeader")->Data();
+            rawHtml += L"</span>";
+            rawHtml += L"<table cellpadding=\"0\">";
+            for (unsigned i = 0; i < equations->Size; i++)
             {
-                expression = EmptyEquationString->Data();
+                auto expression = equations->GetAt(i)->Expression;
+                auto color = equations->GetAt(i)->LineColor->Color;
+
+                if (expression->IsEmpty())
```
rudyhuyn: very very NIT: move `auto color = equations->GetAt(i)->LineColor->Color;` after the conditional

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -105,79 +109,100 @@ void CalculatorApp::GraphingCalculator::OnShareClick(Platform::Object ^ sender,
...
+                {
+                    continue;
+                }
+
+                hasEquations = true;
+
+                expression = GraphingControl->ConvertToLinear(expression);
+
+                std::wstring equatinColorHtml = L"color:rgb(";
+                equatinColorHtml += color.R.ToString()->Data();
```
rudyhuyn: typo: equati**O**nColorHtml

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -105,79 +109,100 @@ void CalculatorApp::GraphingCalculator::OnShareClick(Platform::Object ^ sender,
...
+                equatinColorHtml += L",";
+                equatinColorHtml += color.G.ToString()->Data();
+                equatinColorHtml += L",";
+                equatinColorHtml += color.B.ToString()->Data();
+                equatinColorHtml += L");";
+
+                rawHtml +=
+                    L"<tr><td><span style=\"line-height: 0; font-size: 24pt;" + equatinColorHtml + L"\">&#x25A0;</span></td><td><div style=\"margin: 4px 0px 0px 0px; \">";
+                rawHtml += expression->Data();
```
rudyhuyn: we should probably HTML encode this expression in case it uses some reserved characters.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -48,13 +47,13 @@
...
-                                  EquationsSource="{x:Bind ViewModel.Equations, Mode=OneWay}"
-                                  ForceProportionalAxes="True"
-                                  UseSystemFocusVisuals="True"
-                                  VariablesUpdated="GraphVariablesUpdated"
-                                  LostFocus="OnGraphLostFocus"
-                                  LosingFocus="OnLoosingFocus">
+                <graphControl:Grapher Name="GraphingControl"
+                                      EquationsSource="{x:Bind ViewModel.Equations, Mode=OneWay}"
+                                      ForceProportionalAxes="True"
+                                      LosingFocus="OnLosingFocus"
```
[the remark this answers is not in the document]
DavidShoe: I wouldn't do that, your making a hungarianish prefixing for the function names which is fragile.  They are in the controls space and so should be pretty clear as to the context.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -187,21 +212,16 @@ void GraphingCalculator::OnDataRequested(DataTransferManager ^ sender, DataReque
 
         // Set the thumbnail image (in case the share target can't handle HTML)
         requestData->Properties->Thumbnail = bitmapStream;
-
-        // And the bitmap (in case the share target can't handle HTML)
-        requestData->SetBitmap(bitmapStream);
```
DavidShoe: I wouldn't remove this, if they are sharing to outlook or other possible non html target the thumbnail / bitmap can end up being rendered differently.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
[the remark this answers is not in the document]
joseartrivera: This actually isn't in the graph control space, OnLosingFocus could be referring to any of the controls on this page so it could be useful to give it a more descriptive name.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: This is a good call out, do you know of an easy way to do this in c++/cx?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: I noticed that sharing to outlook doesn't seem to bring over any data, including this bitmap, seems like an issue on the receiving end? While sharing via mail it includes the html **and** the image as an attachment which seems noisy.

sanderl:
It looks like you have an extra space for an equation that shouldn't be there. It shows up in mail, but not in OneNote.

image

joseartrivera:
> It looks like you have an extra space for an equation that shouldn't be there. It shows up in mail, but not in OneNote.
> 
> image

Fixed, I think it's because I had an unnecessary open table tag

rudyhuyn:
Perfect! Thanks a lot!

PR #786, opened 2019-11-08
Update build version to 1910

volkcor:
updating minor versioning to 1910 release

### DIFF
--- a/build/pipelines/azure-pipelines.release.yaml
+++ b/build/pipelines/azure-pipelines.release.yaml
@@ -9,7 +9,7 @@ pr: none
 
 variables:
   versionMajor: 10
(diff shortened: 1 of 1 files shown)


build/pipelines/azure-pipelines.release.yaml
```diff
@@ -9,7 +9,7 @@ pr: none
 
 variables:
   versionMajor: 10
-  versionMinor: 1909
+  versionMinor: 1910
   versionBuild: $[counter('10.1909.*', 0)]
```
sanderl: You need to update the build version too.

volkcor: Updated build versioning too

PR #787, opened 2019-11-08
Fix focus for EquationTextBox

joseartrivera:
- After submitting an equation with the enter key we should move focus away from the EquationTextBox . Previously this would happen anytime the equation was submitted, now it only occurs if it was submitted and the EquationTextBox has focus, allowing the user to tab away to submit and keep focus flowing to the next element as normal.

- Manual tests

### DIFF
--- a/src/Calculator/Controls/EquationTextBox.cpp
+++ b/src/Calculator/Controls/EquationTextBox.cpp
@@ -144,7 +144,7 @@ void EquationTextBox::OnRichEditBoxTextChanged(Object ^ sender, RoutedEventArgs 
 
 void EquationTextBox::OnRichEditBoxGotFocus(Object ^ sender, RoutedEventArgs ^ e)
 {
-    m_isFocused = true;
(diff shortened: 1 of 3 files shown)


src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -144,7 +144,7 @@ void EquationTextBox::OnRichEditBoxTextChanged(Object ^ sender, RoutedEventArgs
 
 void EquationTextBox::OnRichEditBoxGotFocus(Object ^ sender, RoutedEventArgs ^ e)
 {
-    m_isFocused = true;
+    m_HasFocus = true;
```
sanderl: Is this only for RichEditFocus of for the whole control?

joseartrivera: Technically the whole control, but currently we only care when the RichEditBox gets focus

PR #788, opened 2019-11-09
Update Standard Mode Calculator UI Tests

### LINKED ISSUE #610: Make Automated UI Tests for the Standard Calculator, faster, reliable, and AoT tests and Error Tests
**Problem Statement**
Currently, the automated UI tests for Standard calculator mode are slow, keyboard test are flakey, and they do not include AoT tests.  

**Evidence or User Insights**
It would be helpful to developers and testers to have some streamlined, reliable UI tests for the Standard calculator mode that also include tests for the new AoT feature.  Making these test reliable, and fast will encourage the use of these tests.

**Proposal**
- These tests would allow developers to know that their feature/code change, when introduced to the application, does not break other aspects of the application.  
- These tests allow testers to focus on more creative, testing scenarios during manual testing rather than spending a lot of extra time executing tests that an automaton could complete.

**Goals**
- Make Standard calculator mode automated UI test, streamlined, faster, and more reliable
- Include a couple of basic error tests
- Include AoT tests
- Use these tests as a gate when checking in new features

**Non-Goals**
- The UI automated tests are not a replacement for Unit tests
- The UI automated tests will not cover end-to-end testing (It will cover smoke/BVT type tests)

**Low-Fidelity Concept**
In a forked repo, I made small changes to the UI test framework, and to the UI tests that impact the Standard Mode Functional:
https://github.com/v-credwa/calculator/tree/autocal/src
- Has error tests
- Has an AoT tests
- Resize app window before tests so that all required buttons and the history panel is visible
- Does not have changes to make keyboard tests more reliable
- May need more/fewer tests to streamline

**Requested Assignment**
If possible, I would like to implement this

v-credwa:
- I'm working on updating standard calculator test with the end goal to have most repeated manual tests automated
- This work addresses resolves #610 "Make Automated UI Tests for the Standard Calculator, faster, reliable, and AoT tests and Error Tests"

### Description test coverage:
- Standard Mode Functional Tests
    - Basic UI Functionality via Mouse Input Tests
        - These automated tests verify mouse input in the Calculator UI, while simultaneously verify standard calculator functionality
        - These test cases are P0
    - Basic Calculator Functionality via Keyboard Input Tests
        - These automated tests verify the functionality of all hotkeys, that are common for all calculator modes, as well as the all keyboard input used in standard mode on the Standard Calculator
            - Note: Active issue "Bug 23811901: Clicking on the History Label causes the [Shift] + [Ctrl] + [D] hotkeys to break" causes KeyboardInput_HistoryHotkeys() to fail. It is currently commented out until the active bug is fixed.
        - These test cases are are P1
    - Arithmetic Operators Functionality via Mixed Input
        - These automated tests not only verify each operator individually, but they also simultaneously verify mixed user input.
        - These test cases are mostly P2 with one P0 test case
    - AoT (Always on top) Tests
        - These automated tests verify basic AoT calculator UI functionality
        - These test cases are a mix of P0, P1, and P2
    - Error Tests
        - These automated tests verify errors display gracefully
        - These test cases are are P1
- History Functional Tests
    - Standard History Tests
        - These automated tests verify using history related button/keyboard actions in the History Panel, and in the History flyout while in standard mode
        - These test cases are P2
- Memory Functional Tests
    - Standard History Tests
        - These automated tests verify using memory related button/keyboard actions in the Memory Panel, and in the Memory flyout while in standard mode
        - These test cases are P2
- Calculator App Stress Tests
    - Currently contains a very mild, menu navigation stress tests, in the future it will contain multiple session stress test and a more robust menu navigation stress test.  It is an on demand test suit.
        - These test cases are P3

- I locally ran all the tests as whole
- I locally ran the P0 tests as group 10 times repeatedly using a for loop
- I locally ran each P0 test 100 times repeatedly using a for loop


EriWong:
We should probably git ignore out directories as those are things that probably shouldn't be in the repo.

src/Calculator/Calculator.vcxproj.filters
```diff
@@ -465,6 +465,7 @@
     <Page Include="Views\TitleBar.xaml">
       <Filter>Views</Filter>
     </Page>
+    <Page Include="$(MSBuildThisFileDirectory)DensityStyles\Compact.xaml" />
```
EriWong: ><Page Include="$(MSBuildThisFileDirectory)DensityStyles\Compact.xaml" /> [](start = 4, length = 72)

did you add this manually?  If not, I'm curious as to how it was done.  I think this is probably not quite what you want to add to the filters file, but rather a relative path to the page and a Filter sub element to cover having the folder structure be preserved in visual studio.

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace CalculatorUITestFramework
+{
+    public class CalcHistoryFlyout
+    {
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public HistoryPanel HistoryPanel = new HistoryPanel();
+        public WindowsElement Header => this.session.TryFindElementByAccessibilityId("Header");
+        public WindowsElement AppName => this.session.TryFindElementByAccessibilityId("AppName");
+        public WindowsElement HistoryFlyout => this.session.TryFindElementByAccessibilityId("HistoryFlyout");
+        public AppiumWebElement ListViewItem => WinAppDriver.Instance.CalculatorSession.FindElementByClassName("Popup").FindElementByClassName("ListViewItem");
```
EriWong: >FindElementByClassName [](start = 88, length = 22)

Finding by classname is something that we should probably try to avoid as if someone adds another popup or listviewitem to the product you might find the wrong one.

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -6,26 +6,41 @@
 using OpenQA.Selenium.Appium;
 using OpenQA.Selenium.Appium.Windows;
 using System.Collections.ObjectModel;
+using System.Drawing;
+using OpenQA.Selenium.Interactions;
+using System;
```
EriWong: Please sort your using statements.  If you're using visual studio to code this up, you can right click this and select "remove and sort usings".

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -47,15 +58,18 @@ public void TestInit()
             {
                 page.ClearAll();
             }
+            page.AppName.Click();
```
EriWong: >page.AppName.Click(); [](start = 12, length = 21)

it's probably worth adding a comment here as to why you are clicking the app name as part of test init.  Is this to ensure focus is in a known starting state?

v-credwa: This is to ensure focus.  I will add a comment

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Windows.Input;
+
+namespace CalculatorUITestFramework
+{
+    public class CalcHistoryFlyout
+    {
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public HistoryPanel HistoryPanel = new HistoryPanel();
+        public WindowsElement Header => this.session.TryFindElementByAccessibilityId("Header");
+        public WindowsElement AppName => this.session.TryFindElementByAccessibilityId("AppName");
```
EriWong: I think it's a little odd to have the Header and Appname properties here.  From test case code, I would be a bit confused if I saw "myHistoryPanel.AppName" giving me back an element.  Perhaps we should remove these and just use the session directly?  More generally speaking, it's a good idea to keep things that you don't expect people to consume from test code private.

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
```diff
@@ -0,0 +1,52 @@
...
+
+
+        /// <summary>
+        /// Opens the History Flyout.
+        /// </summary>
+        public void OpenHistoryFlyout()
+        {
+            this.HistoryPanel.ResizeWindowToNotDiplayHistoryLabel();
+            this.AppName.Click();
+            this.Header.SendKeys(Keys.Control + "h" + Keys.Control);
```
EriWong: >h [](start = 49, length = 1)

I think alt i is the command to open the history pane.  "h" opens the calc mode picker flyout.

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
v-credwa: These are used to send keys in the test.  Header is used throughout most of standard tests.  However, since calculator doesn't have a header in when it is in KOT(keep on top) mode, than AppName is used.

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
```diff
@@ -0,0 +1,52 @@
...
+            Actions moveToHistoryFlyout = new Actions(WinAppDriver.Instance.CalculatorSession);
+            moveToHistoryFlyout.MoveToElement(HistoryFlyout);
+            moveToHistoryFlyout.Perform();
```
EriWong: I don't think this is necessary...  From what I've read up on, this just moves the mouse to the historyflyout, but doesn't click it or anything.

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
v-credwa: - Pressing [Ctrl] + [H] open the history flyout panel when the calculator is to small to display the history panel
- Pressing [Alt] + [H] opens the navigation menu

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -34,22 +49,95 @@ public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
         /// </summary>
         public void ClearHistory()
         {
-            this.HistoryLabel.Click();
 
+            this.HistoryLabel.Click();
             try
             {
-                this.ClearHistoryButton.Click();
+                string sourceB = this.session.PageSource;
```
EriWong: >sourceB [](start = 23, length = 7)

what naming convention are you using here?  What does the "B" stand for?

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
v-credwa: I do that so the history flyout stays open longer, it did not work without it previously, but I can try again in case something else was causing it to fail.

src/CalculatorUITestFramework/HistoryPanel.cs
v-credwa: B does not stand for anything.  It was being used to make the variable unique.

src/CalculatorUITestFramework/MemoryPanel.cs
```diff
@@ -38,5 +54,110 @@ public ReadOnlyCollection<AppiumWebElement> GetAllMemoryListViewItems()
...
+                    this.ClearMemory.Click();
+                }
+                else
+                {
+                    return;
+                }
+            }
+            catch (WebDriverException ex)
+            {
+                if (ex.Message.Contains("element could not be located"))
```
EriWong: >ex.Message.Contains("element could not be located") [](start = 20, length = 51)

It looks like we are doing this in a number of other places, but it's generally better practice to key off of the type of exception rather than the message. 
ie:
try
{
...
}
catch(ElementNotFoundException ex)
{
//run code for this exception type
}
catch(OtherTypeOfException ex)
{
//run code for OtherTypeOfException
}

anything that isn't one of the above 2 exception types will just throw.  If an exception is both of those types (ie: one inherits from other) it will only run the first exception handler.

src/CalculatorUITestFramework/MemoryPanel.cs
```diff
@@ -38,5 +54,110 @@ public ReadOnlyCollection<AppiumWebElement> GetAllMemoryListViewItems()
...
+
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// If the Memory label is not displayed, resize the window
+        /// Two attempts are made, the the lable is not found a "not found" exception is thrown
+        /// </summary>
+        public void ResizeWindowToDiplayMemoryLabel()
```
EriWong: >ResizeWindowToDiplayMemoryLabel [](start = 20, length = 31)

minor typo in Display

src/CalculatorUITestFramework/HistoryPanel.cs
v-credwa: I will

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
v-credwa: These elements do not have automation IDs.  Would it be possible to give them automation IDs?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -47,15 +58,18 @@ public void TestInit()
...
         }
 
         [TestCleanup]
         public void TestCleanup()
         {
+            page.StandardOperators.EqualButton.Click();
```
EriWong: >page.StandardOperators.EqualButton.Click(); [](start = 12, length = 43)

Is this necessary?  I think ClearAll handles most of this...

src/Calculator/Calculator.vcxproj.filters
v-credwa: I did not add this manually.  I'm not sure how this was added.

src/CalculatorUITestFramework/HistoryPanel.cs
v-credwa: Done

src/CalculatorUITestFramework/HistoryPanel.cs
EriWong: I'd suggest not doing that and leaving them all as source.  You only need to keep variables unique to a scope.

src/CalculatorUITestFramework/CalcHistoryFlyout.cs
EriWong: I get why you have those, but I don't think this is the right place for them.  From a test author's perspective, the AppName element exists on the whole calculator and not the HistoryPanel.  I don't think we have a central place to put appwide elements, but perhaps we should create one?  Alternatively, making it private instead of public can help too as it becomes more of an implementation detail.

v-credwa: I think it's a good idea to have a global calculator cs that contains global calculator elements like these.  I will add that.

src/CalculatorUITestFramework/MemoryPanel.cs
```diff
@@ -38,5 +51,112 @@ public ReadOnlyCollection<AppiumWebElement> GetAllMemoryListViewItems()
...
+
+        /// <summary>
+        /// If the Memory button is not displayed, resize the window
+        /// </summary>
+        public void ResizeWindowToDiplayMemoryButton()
+        {
+            Point newWindowPostion = new Point(0, 0);
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Position = newWindowPostion;
+            string source5 = this.session.PageSource;
+            if (source5.Contains("Alt, M"))
```
sanderl: Why are we looking for the hotkeys?

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -14,29 +19,39 @@ public class StandardCalculatorPage
...
             this.NavigationMenu.ChangeCalculatorMode(CalculatorMode.StandardCalculator);
             Assert.IsNotNull(CalculatorResult);
         }
-
         /// <summary>
         /// Clear the Calculator display, Memory Panel and optionally the History Panel
         /// </summary>
         public void ClearAll()
         {
+            this.StandardNonKoTMode();
```
sanderl: Please use AoT to be consistent with the product code.

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -45,5 +60,177 @@ public string GetCalculatorResultText()
         {
             return this.CalculatorResult.Text.Replace("Display is", string.Empty).Trim();
         }
+        /// <summary>
+        /// Gets the text from the display control in KOT mode and removes the narrator text that is not displayed in the UI.
+        /// </summary>
+        /// <returns>The string shown in the UI.</returns>
+        public string GetKOTCalculatorResultText()
```
sanderl: I think CalculatorResult could be it's own class and you could have this as another method in it.

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -45,5 +60,177 @@ public string GetCalculatorResultText()
...
+                else
+                {
+                    throw new NotFoundException("Could not find the 'Keep on top' button");
+                }
+            }
+        }
+        ///// <summary>
+        ///// Ensures the calculator is in KOT(Keep on top) mode, and verifies that the KOT mode through the absence of the header
+        ///// </summary>
+        public void StandardKoTMode()
```
sanderl: It might make sense to have a StandardAoTCalculatorPage class

sanderl: And associated test class

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -45,5 +60,177 @@ public string GetCalculatorResultText()
...
+            string sourceG = this.session.PageSource;
+            if (sourceG.Contains("NormalAlwaysOnTopButton"))
+            {
+                this.EnterAlwaysOnTopButton.Click();
+                string sourceH = WinAppDriver.Instance.CalculatorSession.PageSource;
+                if (sourceH.Contains("Header"))
+                {
+                    throw new NotFoundException("Failed to enter 'Keep on top' mode; In KoT mode, Calculator does not have header");
+                }
+                else
```
sanderl: You can remove all of the else blocks and just let the method end.

src/CalculatorUITestFramework/WinAppDriver.cs
```diff
@@ -69,6 +72,9 @@ public void SetupCalculatorSession(TestContext context)
                 this.CalculatorSession = new WindowsDriver<WindowsElement>(this.windowsDriverService.ServiceUrl, options);
                 this.CalculatorSession.Manage().Timeouts().ImplicitWait = TimeSpan.FromSeconds(10);
                 Assert.IsNotNull(this.CalculatorSession);
+                this.MemoryPanel.ResizeWindowToDiplayMemoryLabel();
+                //Size windowSize = new Size(1200, 1050);
```
sanderl: remove if this isn't needed

src/CalculatorUITests/StandardModeFunctionalTests.cs
sanderl: It commits to the history panel so that the history delete button shows up. However, I think removing this and using the page source before clicking the button makes more sense.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -209,14 +215,18 @@ public void Operator_ClearEntry()
...
+            page.Header.SendKeys("890987");
+            page.Header.SendKeys(Keys.Equal);
             Assert.AreEqual("-820,897", page.GetCalculatorResultText());
         }
 
         [TestMethod]
         public void KeyboardInput_Multiply()
         {
             page.Header.SendKeys("-15*3=");
+            page.CalculatorResult.Click();
```
sanderl: why do you need to click on the CalculatorResult?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+        }
+        #endregion
+
+        #region KOT(Keep on top) Tests
+        [TestMethod]
+        [Priority(0)]
+        public void KOT_EnterExitKeepOnTop()
+        {
+            /// <summary>
+            /// Test Case 23456961: BVT: KOT: Verify entering and exiting KOT mode
```
sanderl: If this is just for aligning to VSTS manual tests, I would remove these tags. It is likely they will get out of sync and these will no longer be valid.

Same goes for the below tests.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            /// and it verifies exiting KOT mode
+            /// </summary>
+            page.StandardKoTMode();
+            page.KOTWindowSizeWithinRange();
+            page.StandardNonKoTMode();
+
+        }
+
+        [TestMethod]
+        [Priority(0)]
```
sanderl: Make sure priorities are added all of the tests in this class

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            /// <summary>
+            /// Test Case 23458498: BVT: KOT: Verify Memory function is not accessible while in KOT mode
+            /// This automated test verifies memory function cannot be triggered in KOT calculator
+            /// </summary>
+            ///
+            page.StandardOperators.NumberPad.Num9Button.Click();
+            page.StandardOperators.MinusButton.Click();
+            page.StandardOperators.NumberPad.Num3Button.Click();
+            page.StandardKoTMode();
+            page.Window.SendKeys(Keys.Enter);
```
sanderl: Please add comments for why you are sending these keys

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+        [Priority(1)]
+        public void KOT_HistoryFunction()
+        {
+            /// <summary>
+            /// Test Case 23458436: BVT: KOT: Verify History is not accessible while in KOT mode, but it is still tracked
+            /// This automated test verifies the history flyout cannot be opened in KOT mode and
+            /// that history is tracked while in KOT mode and avalible upon ext of KOT mode
+            /// </summary>
+            page.HistoryPanel.ClearHistory();
+            page.StandardOperators.NumberPad.Num3Button.Click();
```
sanderl: I don't think you need to do these to verify you can't open the history panel. Also the testcase name should be updated to somthing like AOT_HistoryPanelCannotBeOpenedInAOTMode

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            /// This automated test verifies the window scale retention is maintained when switching between KOT and Non-KOT
+            /// Note: This test case is in progress: Need to verify " All operator buttons (except for "CE", "C", "Delete" and "="), and memory buttons are disabled"
+            /// </summary>
+
+            Size largeKOTWindowSize = new Size(502, 502);
+            Point largeKOTWindowPosition = new Point(0, 0);
+            page.StandardKoTMode();
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Position = largeKOTWindowPosition;
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = largeKOTWindowSize;
+            page.Window.SendKeys("/");
```
sanderl: Send keys are not self-explanatory so they need comments. Please add comments for any send keys that are not obvious

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            page.Header.SendKeys(Keys.Subtract);
+            var memoryItems2 = page.MemoryPanel.GetAllMemoryListViewItems();
+            Assert.IsTrue(memoryItems2[0].Text.Equals("2", StringComparison.InvariantCultureIgnoreCase));
+            page.Header.SendKeys(Keys.Control + "L" + Keys.Control);
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("MemoryPaneEmpty"));
+        }
+
+        [TestMethod]
+        [Priority(0)]
+        public void BVT_MouseInput()
```
sanderl: What is the purpose of this test?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            Assert.IsTrue(memoryItems2[0].Text.Equals("2", StringComparison.InvariantCultureIgnoreCase));
+            page.MemoryPanel.MemoryClear.Click();
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("MemoryPaneEmpty"));
+        }
+
+        [TestMethod]
+        [Priority(2)]
+        public void BVT_VerifyMemoryFunction()
+        {
+            /// <summary>
```
sanderl: This test seems to be testing too much at once

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            page.CalcMemoryFlyout.OpenMemoryFlyout();
+            page.CalcMemoryFlyout.ClearMemory.Click();
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("MemoryPaneEmpty"));
+            page.MemoryPanel.ResizeWindowToDiplayMemoryLabel();
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("MemoryPaneEmpty"));
+        }
+
+        [TestMethod]
+        [Priority(2)]
+        public void BVT_HistoryPanel()
```
sanderl: Why is a BVT P2? Also this appears like it is testing too much

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            Assert.AreEqual("-1", page.GetCalculatorResultText());
+
+            page.HistoryPanel.ClearHistory();
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("HistoryEmpty"));
+        }
+
+        [TestMethod]
+        [Priority(2)]
+        public void BVT_HistoryFlyout()
+        {
```
sanderl: How does the history flyout behave different than the history panel? Is it the same control that is accessed in different ways?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            page.CalcHistoryFlyout.OpenHistoryFlyout();
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("HistoryEmpty"));
+
+            page.HistoryPanel.OpenHistoryPanel();
+            Assert.IsNotNull(WinAppDriver.Instance.CalculatorSession.FindElementByAccessibilityId("HistoryEmpty"));
+        }
+
+        [TestMethod]
+        [Priority(2)]
+        public void BVT_ArithmeticOperators()
```
sanderl: Is it a BVT or P2?

sanderl: Aren't these already tested in the smoke tests? How is this test different from the other tests that already use the standard operators?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            Assert.AreEqual("3 + 9 Minus (", page.GetCalculatorExpressionText());
+
+            //Verify Clear
+            page.StandardOperators.ClearButton.Click();
+            string sourceD = WinAppDriver.Instance.CalculatorSession.PageSource;
+            if (sourceD.Contains("CalculatorExpression"))
+            {
+                throw new NotFoundException("This test fails; the Calculator Expression should be cleared");
+            }
+            else
```
sanderl: remove this

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -304,5 +314,893 @@ public void Memory_ClearTest()
...
+            {
+                return;
+            }
+        }
+
+        [TestMethod]
+        [Priority(3)]
+        public void BVT_SwitchBetweenStandardAndOtherTypes()
+        {
```
sanderl: This also feels redundant to me.

v-credwa: This is a priority 3.  This test is in place as a stability test.  Switching between modes like this not only verifies that the user can switch from standard to any other mode, but that the user can switch to from any mode to standard.  Doing this also lightly stresses the app, checking to make sure it is stable

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: This actually correctly verifies that Clear worked because it checks the calculator expression.  If there isn't a cleared calculator expression then the test fails

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -6,26 +6,39 @@
...
 namespace CalculatorUITestFramework
 {
     public class HistoryPanel
     {
         private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public WindowsElement HistoryButton => this.session.TryFindElementByAccessibilityId("HistoryButton");
         public WindowsElement HistoryLabel => this.session.TryFindElementByAccessibilityId("HistoryLabel");
+        public AppiumWebElement HistoryEmpty => this.session.TryFindElementByAccessibilityId("DockPivot").FindElementByAccessibilityId("HistoryEmpty");
         public WindowsElement HistoryListView => this.session.TryFindElementByAccessibilityId("HistoryListView");
+        public WindowsElement HistoryList => this.session.TryFindElementByAccessibilityId("HistoryList");
```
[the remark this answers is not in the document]
v-credwa: I will remove HistoryList

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -6,26 +6,39 @@
...
+        public WindowsElement ListViewItem => this.session.FindElementByClassName("ListViewItem");
+        public WindowsElement ExprTextBlock => this.session.TryFindElementByAccessibilityId("ExprTextBlock");
```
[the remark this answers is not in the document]
v-credwa: I wound up not using this item.  I'll remove it.

src/CalculatorUITestFramework/HistoryPanel.cs
v-credwa: when I could, I'll leave it at source, in the cases were I can't, I'll use a numbering system to make it more clear.
E.g. 
source0
source1

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -34,22 +47,97 @@ public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
...
+                {
+                    this.ClearHistoryButton.Click();
+                }
+                else
+                {
+                    return;
+                }
             }
-            catch(WebDriverException ex)
+            catch (WebDriverException ex)
```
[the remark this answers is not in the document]
v-credwa: Sometimes

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -34,22 +47,97 @@ public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
...
                     return;
                 }
 
                 throw;
             }
         }
+
+        /// <summary>
+        /// If the History label is not displayed, resize the window
+        /// Two attempts are made, the the lable is not found a "not found" exception is thrown
```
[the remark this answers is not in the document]
v-credwa: Corrections will appear in my next commit

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -34,22 +47,97 @@ public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
...
+        /// </summary>
+        public void ResizeWindowToDiplayHistoryLabel()
+        {
+            Point newWindowPostion = new Point(0, 0);
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Position = newWindowPostion;
+            string source0 = this.session.PageSource;
```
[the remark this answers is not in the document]
v-credwa: In this case, I do need to keep source0, because source is checked at different steps in the method

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -34,22 +47,97 @@ public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
...
+                Size newWindowSize = new Size(1200, 1050);
+                WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize;
+                string source2 = this.session.PageSource;
+                if (source2.Contains("HistoryLabel"))
+                {
+                    return;
+                }
+                else
+                {
+                    Size newWindowSize2 = new Size(2097, 1282);
```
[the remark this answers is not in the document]
v-credwa: One size works on desktop, one size works on laptop

src/CalculatorUITestFramework/MemoryPanel.cs
```diff
@@ -38,5 +51,112 @@ public ReadOnlyCollection<AppiumWebElement> GetAllMemoryListViewItems()
...
+
+        /// <summary>
+        /// If the Memory label is not displayed, resize the window
+        /// Two attempts are made, the the lable is not found a "not found" exception is thrown
+        /// </summary>
+        public void ResizeWindowToDiplayMemoryLabel()
+        {
+            Point newWindowPostion = new Point(0, 0);
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Position = newWindowPostion;
+            string source3 = this.session.PageSource;
```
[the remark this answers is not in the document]
v-credwa: Some of them do need to be unique.  However, there were some that did not.  The changes will be in my next commit

src/CalculatorUITestFramework/MemoryPanel.cs
```diff
@@ -38,5 +51,112 @@ public ReadOnlyCollection<AppiumWebElement> GetAllMemoryListViewItems()
...
+                Size newWindowSize = new Size(1200, 1050);
+                WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize;
+                string source4 = this.session.PageSource;
+                if (source4.Contains("MemoryLabel"))
+                {
+                    return;
+                }
+                else
+                {
+                    Size newWindowSize2 = new Size(2097, 1282);
```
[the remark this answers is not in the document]
v-credwa: One works with laptops, one works with desktop.  Sizing to the largest size does not work in all cases.

src/CalculatorUITestFramework/MemoryPanel.cs
v-credwa: Because of the way this check works.  Using memory button name gives a false positive.  Looking for this does not

src/CalculatorUITestFramework/StandardCalculatorPage.cs
v-credwa: Change will appear in my next commit

src/CalculatorUITestFramework/WinAppDriver.cs
v-credwa: It is needed, it makes the test cases stable and run faster.  I will remove the commented out test. The change will appear in my next commit

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: Change will appear in next commit

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: The UI is interacted with in different ways

src/Calculator/Calculator.vcxproj.filters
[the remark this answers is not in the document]
v-credwa: I have removed this line.  Change will be in next commit

src/CalculatorUITestFramework/HistoryPanel.cs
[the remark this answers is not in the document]
v-credwa: I've removed the try/catch from this method

src/CalculatorUITestFramework/StandardCalculatorPage.cs
v-credwa: Not all of them, but the ones that only have "return" in them.

v-credwa: The changes will be in my next commit

src/CalculatorUITestFramework/StandardCalculatorPage.cs
v-credwa: In https://github.com/microsoft/calculator/pull/788/commits/ded929f19a56f1eef397c5864d4ec6fde59e78ab

src/CalculatorUITestFramework/StandardCalculatorPage.cs
v-credwa: In https://github.com/microsoft/calculator/pull/788/commits/ded929f19a56f1eef397c5864d4ec6fde59e78ab

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: In https://github.com/microsoft/calculator/pull/788/commits/ded929f19a56f1eef397c5864d4ec6fde59e78ab

src/CalculatorUITestFramework/CalculatorResults.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using OpenQA.Selenium.Appium.Windows;
+
+namespace CalculatorUITestFramework
+{
+    public class CalculatorResults
+    {
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public WindowsElement CalculatorAlwaysOnTopResults => this.session.TryFindElementByAccessibilityId("CalculatorAlwaysOnTopResults");
```
EriWong: >public [](start = 8, length = 6)

make this private

src/CalculatorUITestFramework/GlobalCalculatorUI.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            return this.Header.Text;
+        }
+        /// <summary>
+        /// Gets the text from the Calculator Expression control and removes narrator text that is not displayed in the UI.
+        /// </summary>
+        /// <returns>The string shown in the UI.</returns>
+        public string GetCalculatorExpressionText()
+        {
+            return this.CalculatorExpression.Text.Replace("Expression is", string.Empty).Trim();
```
EriWong: >Replace("Expression is", string.Empty).Trim() [](start = 50, length = 45)

remove this.

src/CalculatorUITestFramework/CalculatorResults.cs
v-credwa: As part of commit a8c27208b04f206e706a029edd3930692db69383
In test framework, where it worked, I made elements private

src/CalculatorUITestFramework/GlobalCalculatorUI.cs
v-credwa: currently, this cannot be removed

src/CalculatorUITestFramework/GlobalCalculatorUI.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License.
+using OpenQA.Selenium.Appium.Windows;
+using System.Drawing;
+
+namespace CalculatorUITestFramework
+{
+    /// <summary>
+    /// This class contains the UI automation objects and helper methods available when the Calculator is in any mode.
+    /// </summary>
+    public class GlobalCalculatorUI
```
sanderl: You can call this something like "CalculatorApplication"

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
+        private WindowsElement HistoryListView => this.session.TryFindElementByAccessibilityId("HistoryListView");
+        public WindowsElement ListViewItem => this.session.FindElementByClassName("ListViewItem");
+        private WindowsElement ResultTextBlock => this.session.TryFindElementByAccessibilityId("ResultTextBlock");
         public WindowsElement ClearHistoryButton => this.session.TryFindElementByAccessibilityId("ClearHistory");
-        public WindowsElement HistoryEmptyLabel => this.session.TryFindElementByAccessibilityId("HistoryEmpty");
-
+        private WindowsElement HistoryFlyout => this.session.TryFindElementByAccessibilityId("HistoryFlyout");
 
+        /// <summary>
+        /// Opens the Memory Pane by clicking the Memory pivot label.
```
sanderl:
```suggestion
        /// Opens the History Pane by clicking the History pivot label.
```

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
+            if (source0.Contains("HistoryLabel"))
+            {
+                return;
+            }
+            else
             {
-                if (ex.Message.Contains("element could not be located"))
+                Size newWindowSize = new Size(1200, 1050);
+                WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize;
+                string source1 = this.session.PageSource;
```
sanderl:
```suggestion
                string source = this.session.PageSource;
```

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
                     return;
                 }
-
-                throw;
+                else
+                {
+                    Size newWindowSize2 = new Size(2097, 1282);
+                    WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize2;
+                }
+                string source2 = this.session.PageSource;
```
sanderl:
```suggestion
                source = this.session.PageSource;
```

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
+            {
+                return;
+            }
+            else
             {
-                if (ex.Message.Contains("element could not be located"))
+                Size newWindowSize = new Size(1200, 1050);
+                WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize;
+                string source1 = this.session.PageSource;
+                if (source1.Contains("HistoryLabel"))
```
sanderl:
```suggestion
                if (source.Contains("HistoryLabel"))
```

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
                 }
-
-                throw;
+                else
+                {
+                    Size newWindowSize2 = new Size(2097, 1282);
+                    WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize2;
+                }
+                string source2 = this.session.PageSource;
+                if (source2.Contains("HistoryLabel"))
```
sanderl:
```suggestion
                if (source.Contains("HistoryLabel"))
```

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
+                else
+                {
+                    throw new NotFoundException("Could not the History Label");
+                }
             }
         }
+        ///// <summary>
+        ///// If the History button is not displayed, resize the window
+        ///// </summary>
+        public void ResizeWindowToNotDisplayHistoryLabel()
```
sanderl:
```suggestion
        public void ReduceWindowSizeToShowHistoryButton()
```

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+    public class StandardAoTCalculatorPage
+    {
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public GlobalCalculatorUI GlobalCalculatorUI = new GlobalCalculatorUI();
+        public StandardOperatorsPanel StandardOperators = new StandardOperatorsPanel();
+        public NavigationMenu NavigationMenu = new NavigationMenu();
+        public WindowsElement EnterAlwaysOnTopButton => this.session.TryFindElementByAccessibilityId("NormalAlwaysOnTopButton");
+        public WindowsElement ExitAlwaysOnTopButton => this.session.TryFindElementByAccessibilityId("ExitAlwaysOnTopButton");
+        public AppiumWebElement ToolTip => WinAppDriver.Instance.CalculatorSession.FindElementByClassName("ToolTip").FindElementByClassName("TextBlock");
+
```
sanderl: Move this section to the bottom of the class
```suggestion
#region AoT Helper Methods
```

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+            {
+                if (sourceH.Contains("Keep on top"))
+                {
+                    Actions moveToAoTButton = new Actions(WinAppDriver.Instance.CalculatorSession);
+                    moveToAoTButton.MoveToElement(EnterAlwaysOnTopButton);
+                    moveToAoTButton.Perform();
+                }
+                else
+                {
+                    Actions moveToBTFVButton = new Actions(WinAppDriver.Instance.CalculatorSession);
```
sanderl: BTFV isn't a common acronym, add a comment or adjust the name.

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+                InAoTMode = true;
+            }
+            return InAoTMode.ToString();
+        }
+        /// <summary>
+        /// Checks to see if the size of the Calculator window is within the AoT arange of size
+        /// To-do: Find a way to get resolution and then do something like monitor quadrant location, or
+        /// something like (X >= (0.8*screenWidth))... etc.
+        /// </summary>
+        public void AoTWindowPositionWithinRange()
```
sanderl: This is testing the compact overlay api

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+                if (sourceH.Contains("Header"))
+                {
+                    throw new NotFoundException("Failed to enter 'Keep on top' mode; In AoT mode, Calculator does not have header");
+                }
+            }
+        }
+        /// <summary>
+        /// Checks to see if the size of the Calculator window is within the AoT arange of size
+        /// </summary>
+        public void AoTWindowSizeWithinRange()
```
sanderl: Is testing our implementation or just the compact overlay api?

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+    {
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public GlobalCalculatorUI GlobalCalculatorUI = new GlobalCalculatorUI();
+        public StandardOperatorsPanel StandardOperators = new StandardOperatorsPanel();
+        public NavigationMenu NavigationMenu = new NavigationMenu();
+        public WindowsElement EnterAlwaysOnTopButton => this.session.TryFindElementByAccessibilityId("NormalAlwaysOnTopButton");
+        public WindowsElement ExitAlwaysOnTopButton => this.session.TryFindElementByAccessibilityId("ExitAlwaysOnTopButton");
+        public AppiumWebElement ToolTip => WinAppDriver.Instance.CalculatorSession.FindElementByClassName("ToolTip").FindElementByClassName("TextBlock");
+
+        public void StandardNonAoTMode()
```
sanderl: Rename this to be more descriptive.

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+                else
+                {
+                    throw new NotFoundException("Could not find the 'Keep on top' button");
+                }
+            }
+        }
+        ///// <summary>
+        ///// Ensures the calculator is in AoT(Keep on top) mode, and verifies that the AoT mode through the absence of the header
+        ///// </summary>
+        public void StandardAoTMode()
```
sanderl:
```suggestion
        public void NavigateToStandardAoTMode()
```

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
sanderl: Or make the generic navigate to mode method to make sure it is not in AOT.

src/CalculatorUITestFramework/StandardCalculatorPage.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using OpenQA.Selenium.Appium.Windows;
 
+
```
sanderl: nit: remove extra line

src/CalculatorUITestFramework/StandardOperatorsPanel.cs
```diff
@@ -27,5 +26,8 @@ public class StandardOperatorsPanel
         public WindowsElement ClearEntryButton => this.session.TryFindElementByAccessibilityId("clearEntryButton");
         public WindowsElement ClearButton => this.session.TryFindElementByAccessibilityId("clearButton");
         public WindowsElement BackSpaceButton => this.session.TryFindElementByAccessibilityId("backSpaceButton");
+        public WindowsElement NegateButton => this.session.TryFindElementByAccessibilityId("negateButton");
+        public WindowsElement HistoryButton => this.session.TryFindElementByAccessibilityId("HistoryButton");
```
sanderl: History button isn't an operator, so it should be in a different class.

src/CalculatorUITestFramework/WinAppDriver.cs
```diff
@@ -12,6 +12,7 @@ namespace CalculatorUITestFramework
     public sealed class WinAppDriver
     {
         private WindowsDriverLocalService windowsDriverService = null;
+        public MemoryPanel MemoryPanel = new MemoryPanel();
```
sanderl: This doesn't look like it is used in this class. Only WinAppDriver code should be in this class.

src/CalculatorUITestFramework/GlobalCalculatorUI.cs
v-credwa: Renamed to "CalculatorApp."  Change will be in my next commit.

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
v-credwa: Change name of action to "moveToBackToFullViewVButton."  This mdification will be in my next commit

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
v-credwa: Modification will be in my next commit

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/StandardOperatorsPanel.cs
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/WinAppDriver.cs
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/StandardCalculatorPage.cs
v-credwa: Change will appear in next commit

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
v-credwa: This is test framework method that helps to verify that the calculator is in AoT mode.

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
v-credwa: This is test framework method that helps to verify that the calculator is in AoT mode.

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
+            string source = this.session.PageSource;
+            if (source.Contains("HistoryLabel"))
+            {
+                return;
+            }
+            else
             {
-                if (ex.Message.Contains("element could not be located"))
+                Size newWindowSize = new Size(1200, 1050);
+                WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = newWindowSize;
```
sanderl: `WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = new Size(1200, 1050)`

You don't need the extra variable, you can just assign the size directly.

Same comment for other code like this below.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -25,6 +27,9 @@ public static void ClassInitialize(TestContext context)
 
             // Ensure that calculator is in standard mode
             page.NavigateToStandardCalculator();
+
+            // Ensure that calculator window is large enough to display the memory/history panel; a good size for most tests
+            page.MemoryPanel.ResizeWindowToDiplayMemoryLabel();
```
sanderl: Since the history panel is shown by default should you look for that or just for the pivot?

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -43,277 +48,1082 @@ public static void ClassCleanup()
         [TestInitialize]
         public void TestInit()
         {
-            if ("0" != page.GetCalculatorResultText())
-            {
-                page.ClearAll();
-            }
+            page.EnsureCalculatorIsInStandardMode();
+            page.EnsureCalculatorResultTextIsZero();
+            page.CalculatorApp.EnsureCalculatorHasFocus();
```
sanderl: EnsureCalculatorHasFocus should be first, since the other two could fail if the calculator app doesn't have focus.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -43,277 +48,1082 @@ public static void ClassCleanup()
...
+        }
 
-            page.MemoryPanel.MemButton.Click();
-            page.StandardOperators.NumberPad.Input(3.3);
-            page.MemoryPanel.MemoryClear.Click();
-            page.MemoryPanel.OpenMemoryPanel();
-            Assert.IsNotNull(page.MemoryPanel.MemoryPaneEmptyLabel);
+        #endregion
+
+        #region Stress Tests
```
sanderl: Stress tests should be in their own test class. We can then ensure that we are doing a full stress test and run it on demand.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -43,277 +48,1082 @@ public static void ClassCleanup()
...
-            page.StandardOperators.NumberPad.Input(3.3);
-            page.MemoryPanel.MemRecall.Click();
-            var memoryItems = page.MemoryPanel.GetAllMemoryListViewItems();
-            Assert.IsTrue(memoryItems[0].Text.Equals("4", StringComparison.InvariantCultureIgnoreCase));
         }
 
         [TestMethod]
-        public void Memory_ClearTest()
+        [Priority(2)]
+        public void AoT_ErrorMessage()
```
sanderl: This should be broken into 3 different tests

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -43,277 +48,1082 @@ public static void ClassCleanup()
...
+            Assert.AreEqual("(502, 502)", page.CalculatorApp.GetCalculatorWindowSize());
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Size = mediumAoTWindowSize;
+            Assert.AreEqual("(396, 322)", page.CalculatorApp.GetCalculatorWindowSize());
+            page.StandardAoTCalculatorPage.NavigateToStandardMode();
         }
 
         [TestMethod]
-        public void Memory_RecallTest()
+        [Priority(2)]
+        public void AoT_ScaleRetention()
```
sanderl: Do we need this test? Are we doing special logic in AoT that would affect the scale retention of the app? If not, then we are likely just testing Compact Overlay and UWP.

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -43,277 +48,1082 @@ public static void ClassCleanup()
...
-            page.Header.SendKeys("70090-890987=");
-            Assert.AreEqual("-820,897", page.GetCalculatorResultText());
+            page.StandardOperators.InvertButton.Click();
+            Assert.IsTrue(page.CalculatorResults.GetCalculatorResultText() == "Cannot divide by zero");
         }
 
         [TestMethod]
-        public void KeyboardInput_Multiply()
+        [Priority(1)]
+        public void Error_CubedNegativeNumber()
```
sanderl: Please update this test to match the name, should be ErrorSquareRootNegativeNumber

src/CalculatorUITests/StandardModeFunctionalTests.cs
```diff
@@ -43,277 +48,1082 @@ public static void ClassCleanup()
...
             page.StandardOperators.NumberPad.Input(0);
-            Assert.AreEqual("0", page.GetCalculatorResultText());
+            page.StandardOperators.NumberPad.Num0Button.Click();
+            page.StandardOperators.NumberPad.Input(0);
+            Assert.AreEqual("0", page.CalculatorResults.GetCalculatorResultText());
         }
 
+        #endregion
+
+        #region Standard Memory Tests
```
sanderl: Do History and Memory panels behave the same regardless of mode? If so, it might make more sense to move these to history and memory test suites instead of having them in the standard, scientific, etc test suites.

.gitignore
```diff
@@ -293,4 +293,8 @@ src/GraphControl/GraphingImplOverrides.props
 src/CalcViewModel/DataLoaders/DataLoaderConstants.h
 !/build/config/TRexDefs/**
 !src/Calculator/TemporaryKey.pfx
-!src/CalculatorUnitTests/CalculatorUnitTests_TemporaryKey.pfx
\ No newline at end of file
+!src/CalculatorUnitTests/CalculatorUnitTests_TemporaryKey.pfx
```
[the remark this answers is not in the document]
v-credwa: This was not the change was requested by Eric, it causes these output folders to be ignored.  Otherwise,  the branch, repo, PR, etc would be flooded with changes not needed when pushing a commit.

src/CalculatorUITestFramework/CalculatorApp.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>
+        /// <returns>The string shown in the UI.</returns>
+        public string GetCalculatorHeaderText()
+        {
+            return this.Header.Text;
+        }
+        ///// <summary>
+        ///// Gets the calculators window's X and Y position and returns it as a string. e.g. "(0, 0)"
+        ///// </summary>
+        public string GetCalculatorWindowPosition()
```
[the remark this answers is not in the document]
v-credwa: It was, currently it is not.

src/CalculatorUITestFramework/CalculatorApp.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            Point getCalculatorWindowPosition = WinAppDriver.Instance.CalculatorSession.Manage().Window.Position;
+            int X = getCalculatorWindowPosition.X;
```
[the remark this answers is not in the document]
v-credwa: I can change these to lowercase

src/CalculatorUITestFramework/CalculatorApp.cs
```diff
@@ -0,0 +1,59 @@
...
+            int Y = getCalculatorWindowPosition.Y;
+            var CalculatorWindowPosition = (X, Y);
+            return CalculatorWindowPosition.ToString();
+        }
+        ///// <summary>
+        ///// Gets the calculators window's width and height and returns it as a string. e.g. "(322, 396)"
+        ///// </summary>
+        public string GetCalculatorWindowSize()
```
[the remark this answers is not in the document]
v-credwa: That was the first thing I tried; it did not work.

src/CalculatorUITestFramework/CalculatorResults.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <returns>The string shown in the UI.</returns>
+        public string GetCalculatorExpressionText()
+        {
+            return this.CalculatorExpression.Text.Replace("Expression is", string.Empty).Trim();
+        }
+        /// <summary>
+        /// Verifies that CalculatorResult is not null
+        /// </summary>
+        /// <returns>The string shown in the UI.</returns>
+        public void CheckIfResultsDisplayIsPresent()
```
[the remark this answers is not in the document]
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/CalculatorResults.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            Assert.IsNotNull(this.CalculatorResult);
+        }
+        /// <summary>
+        /// Verifies that Calculator Expression is clear
+        /// </summary>
+        /// <returns>The string shown in the UI.</returns>
+        public void CheckIfResultsExpressionIsClear()
```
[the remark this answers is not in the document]
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/CalculatorResults.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            string source = WinAppDriver.Instance.CalculatorSession.PageSource;
+            if (source.Contains("CalculatorExpression"))
+            {
+                throw new Exception("The Calculator Expression is not clear");
```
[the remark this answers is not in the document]
v-credwa: This is helps to give a clear error message when the calculator expression fails to be clear.

src/CalculatorUITestFramework/StandardAoTCalculatorPage.cs
```diff
@@ -0,0 +1,201 @@
...
+                if (source.Contains("Header"))
+                {
+                    throw new NotFoundException("Failed to enter 'Keep on top' mode; In AoT mode, Calculator does not have header");
+                }
+            }
+        }
+        /// <summary>
+        /// Checks to see if the size of the Calculator window is within the AoT arange of size
+        /// </summary>
+        public void AoTWindowSizeWithinRange()
```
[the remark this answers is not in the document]
v-credwa: Yes, resolution makes a difference. (I have not merged a version of this that takes into account laptop resolutions).  To verify that the calculator is in AoT mode, it should be in a certain position, in a certain range of size, and it should only have certain UI elements.  To test that the calculator is going into AoT mode, we should verify these things.

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: Change will be in my next commit

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: Good idea.  To address this idea, I created CalculatorAppFunctionalTests.cs  This cs will contain history and memory tests, and any other future tests that are not specific to one mode.  The change will be in my next commit.

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: Change will be in next commit.  Also, error testing was moved to CalculatorAppFunctionalTests.cs

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: This automated test was created to reduce manual testing.  Whether this Compact Overlay and UWP or specific to calculator, it does need to be tested to ensure that it does not break.  We can remove it from automation, but it should not be removed from testing.

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: I broke this up into two tests:
- AoT_ErrorMessage_ResultUndefined
- AoT_ErrorMessage_CannotDivideByZero
- AoT_ErrorMessage_ MessageRetentionUponExitingAoT

Change will be in next commit

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: Good idea.  To address this idea, I created CalculatorAppStressTests.cs.  This cs will contain this stress test and future stress tests.

src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -1,55 +1,139 @@
...
+        }
+        /// <summary>
+        /// If the History label is not displayed, resize the window
+        /// Two attempts are made; the label is not found a "not found" exception is thrown
+        /// </summary>
+        public void ResizeWindowToDisplayHistoryLabel()
+        {
+            Point newWindowPostion = new Point(8, 8);
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Position = newWindowPostion;
+            string source = this.session.PageSource;
```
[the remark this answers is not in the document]
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/HistoryPanel.cs
v-credwa: Change will be in my next commit

src/CalculatorUITestFramework/CalculatorApp.cs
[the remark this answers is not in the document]
v-credwa: Not anymore.  I'll remove it.

src/CalculatorUITestFramework/HistoryPanel.cs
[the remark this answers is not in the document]
v-credwa: Change in next commit

src/CalculatorUITestFramework/HistoryPanel.cs
v-credwa: Change in next commit

src/CalculatorUITests/StandardModeFunctionalTests.cs
v-credwa: Change in the next commit

src/CalculatorUITestFramework/CalculatorApp.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// </summary>
+    public class CalculatorApp
+    {
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+        public WindowsElement Header => this.session.TryFindElementByAccessibilityId("Header");
+        public WindowsElement Minimize => this.session.TryFindElementByAccessibilityId("Minimize");
+        public WindowsElement Maximize => this.session.TryFindElementByAccessibilityId("Maximize");
+        private WindowsElement AppName => this.session.TryFindElementByAccessibilityId("AppName");
+        public WindowsElement DockPanel => this.session.TryFindElementByAccessibilityId("DockPanel");
+        public WindowsElement Window => this.session.FindElementByClassName("Windows.UI.Core.CoreWindow");
```
EriWong: Put publics before privates and remove unused properties.

src/CalculatorUITestFramework/CalculatorApp.cs
```diff
@@ -0,0 +1,71 @@
...
+        ///// </summary>
+        public void EnsureCalculatorHasFocus()
+        {
+            this.AppName.Click();
+        }
+        /// <summary>
+        /// If the the Dock Panel for the History and Memory lists is not displayed, resize the window
+        /// Two attempts are made, the the lable is not found a "not found" exception is thrown
+        /// </summary>
+        public void ResizeWindowToDiplayMemoryHistoryDockPanel()
```
EriWong: >ResizeWindowToDiplayMemoryHistoryDockPanel [](start = 20, length = 42)

typo

src/CalculatorUITestFramework/CalculatorApp.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            WinAppDriver.Instance.CalculatorSession.Manage().Window.Position = new Point(8, 8);
```
EriWong: >new Point(8, 8); [](start = 79, length = 16)

add comments on what these numbers are

src/CalculatorUITestFramework/CalculatorResults.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            return this.CalculatorAlwaysOnTopResults.Text.Replace("Display is", string.Empty).Trim();
+        }
+        /// <summary>
+        /// Gets the text from the display control and removes the narrator text that is not displayed in the UI.
+        /// </summary>
+        /// <returns>The string shown in the UI.</returns>
+        public string GetCalculatorResultText()
+        {
+            return this.CalculatorResult.Text.Replace("Display is", string.Empty).Trim();
```
EriWong: >Display is" [](start = 55, length = 11)

handle localization

src/CalculatorUITestFramework/MemoryPanel.cs
```diff
@@ -1,30 +1,44 @@
...
+using OpenQA.Selenium.Interactions;
 using System.Collections.ObjectModel;
+using System.Drawing;
 
 namespace CalculatorUITestFramework
 {
     public class MemoryPanel
     {
-        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
         public WindowsElement MemoryClear => this.session.TryFindElementByAccessibilityId("ClearMemoryButton");
```
EriWong: >MemoryClear [](start = 30, length = 11)

probably worth renaming these to be a bit clearer.

PR #791, opened 2019-11-12
Refactor Equation to be treated as a model and update Key Graph Features

joseartrivera:
The current way we transfer data from the EquationViewModel -> Equation is through duplicated dependency properties that are connected via two-way bindings in xaml. This is because the GraphControl was initially designed to receive a DataTemplate where the Equation FrameworkElement was supposed to be used. This isn't an ideal design since a DataTemplate should be able to accept any type of UI but currently would only work when used with the Equation FrameworkElement.

This PR refactors the code so that the GraphControl works more like the [MapControl ](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.maps.mapcontrol)and the [MapRouteView ](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.maps.maprouteview)class. The GraphControl has a collection of Equations. To add an Equation to the Graph, we would append the equation to that collection and then call the PlotGraph() function.

This PR also updated Key Graph Features to use the new refactored way of transferring data. It was also updated to analyze an equation when the user clicks the button instead of analyzing them all at once.

- Manual tests


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -48,39 +47,50 @@
...
-                            <graphControl:Equation Expression="{x:Bind Expression, Mode=OneWay}" LineColor="{x:Bind LineColor, Mode=OneWay}"/>
-                        </DataTemplate>
-                    </graphControl:Grapher.EquationTemplate>
-
                 </graphControl:Grapher>
 
-                <StackPanel Grid.Row="0" Orientation="Horizontal" HorizontalAlignment="Right" VerticalAlignment="Top" Margin="0,0,10,10">
-                    <Button x:Name="ActiveTracing" Click="OnActiveTracingClick" MinWidth="40" MinHeight="40" Margin="0,0,10,0">
+                <StackPanel Grid.Row="0"
+                            Margin="0,0,10,10"
```
sanderl: nit: spacing

joseartrivera: Not clear what spacing you are referring to here?

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -110,7 +110,7 @@ void EquationInputArea::EquationTextBox_KeyGraphFeaturesButtonClicked(Object ^ s
     auto tb = static_cast<EquationTextBox ^>(sender);
     auto eq = static_cast<EquationViewModel ^>(tb->DataContext);
     EquationVM = eq;
```
sanderl: Do we still need to assign eq to EquationVM or can we just remove that line and have the event send eq?

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -31,83 +32,59 @@ namespace CalculatorApp::ViewModel
...
         , m_resourceLoader{ Windows::ApplicationModel::Resources::ResourceLoader::GetForCurrentView() }
     {
+        GraphEquation = equation;
     }
 
-    void EquationViewModel::OnPropertyChanged(String ^ propertyName)
+    void EquationViewModel::PopulateKeyGraphFeatures()
     {
-        if (propertyName == L"IsAnalysisUpdated")
+        if (GraphEquation->AnalysisError != 0)
```
sanderl: Should we have a null check before we start using GraphEquation?

sanderl:
>         InspectingDataSource ^ m_dataSource;

You can remove this

---
Refers to: src/GraphControl/Control/Grapher.h:206 in f242cd4. [](commit_id = f242cd44b8663ab3c96dbd1c64f09d1903da0125, deletion_comment = False)

sanderl:
> #include "InspectingDataSource.h"

This too

---
Refers to: src/GraphControl/Control/Grapher.h:6 in f242cd4. [](commit_id = f242cd44b8663ab3c96dbd1c64f09d1903da0125, deletion_comment = False)

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
sanderl: It looks like it's just on GitHub. I missed the lack of + on the left.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -48,39 +47,50 @@
...
+                <Popup x:Name="TraceValuePopup"
+                       Grid.Row="0"
+                       RenderTransformOrigin="0,1">
                     <Popup.RenderTransform>
                         <TranslateTransform x:Name="TraceValuePopupTransform" X="50" Y="150"/>
                     </Popup.RenderTransform>
-                    <Grid Background="{StaticResource ToolTipBackground}" BorderBrush="{StaticResource ToolTipBorderBrush}" BorderThickness="3,3,3,3">
-                        <TextBlock x:Name="TraceValue" Text="x=0,y=0" Margin="5,5,5,5" Foreground="{StaticResource ToolTipForeground}"/>
+                    <Grid Background="{StaticResource ToolTipBackground}"
+                          BorderBrush="{StaticResource ToolTipBorderBrush}"
```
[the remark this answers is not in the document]
sanderl: Same here.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
joseartrivera: Yes we are still binding to EquationVM in xaml

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
joseartrivera: Added a check in the constructor

src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -31,83 +32,62 @@ namespace CalculatorApp::ViewModel
...
     {
-    }
-
-    void EquationViewModel::OnPropertyChanged(String ^ propertyName)
-    {
-        if (propertyName == L"IsAnalysisUpdated")
+        if (equation == nullptr)
         {
-            OnIsAnalysisUpdatedChanged();
+            throw ref new InvalidArgumentException(L"Equation cannot be null");
```
sanderl: Can we be sure an end user will never hit this?

joseartrivera: Currently, this object is only constructed in one place and we always pass an equation. If the equation becomes null for some reason then we should fix that bug instead of silently failing.

sanderl: sounds good

PR #810, opened 2019-11-18
Fixing typo in StandardCalculatorViewModel::UpdatecommandsInRecording…

### LINKED ISSUE #809: StandardCalculatorViewModel::UpdatecommandsInRecordingMode has typo
The function StandardCalculatorViewModel::UpdatecommandsInRecordingMode is currently spelled with a lowercase `c` for `Commands`.

Some other things that could be refactored in this function is that it is creating a copy of the saved commands vector.  It also uses a raw for loop for iterating over those saved commands.

gsfreema:
- Capitalizing `Commands`
- Removing unnecessary copy of vector in StandardCalculatorViewModel::UpdateCommandsInRecordingMode
- Using range-for in StandardCalculatorViewModel::UpdateCommandsInRecordingMode

### DIFF
--- a/src/CalcViewModel/StandardCalculatorViewModel.cpp
+++ b/src/CalcViewModel/StandardCalculatorViewModel.cpp
@@ -1709,9 +1709,8 @@ void StandardCalculatorViewModel::UpdateOperand(int pos, String ^ text)
     }
 }
 
-void StandardCalculatorViewModel::UpdatecommandsInRecordingMode()
+void StandardCalculatorViewModel::UpdateCommandsInRecordingMode()
 {
-    vector<unsigned char> savedCommands = m_standardCalculatorManager.GetSavedCommands();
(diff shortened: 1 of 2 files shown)


src/CalcViewModel/StandardCalculatorViewModel.cpp
```diff
@@ -1709,22 +1709,18 @@ void StandardCalculatorViewModel::UpdateOperand(int pos, String ^ text)
...
     bool eMinusMode = false;
 
-    int num = 0;
-    Command val;
-    for (unsigned int i = 0; i < savedCommands.size(); ++i)
+    for (const auto savedCommand : m_standardCalculatorManager.GetSavedCommands())
     {
-        val = safe_cast<Command>(savedCommands[i]);
-        num = static_cast<int>(val);
+        const Command val = safe_cast<Command>(savedCommand);
```
mcooley: nit: I think this should just be static_cast

PR #816, opened 2019-11-20
Graphing mode continue

DavidShoe:
Add the graph mode settings dialog and required plumbing.

-
-
-


src/GraphControl/Control/Grapher.h
```diff
@@ -190,6 +190,106 @@ public
...
+		
+        // We can't use the EValTrigUnitMode enum directly in as the property type because it comes from another module which doesn't expose
+        // it as a public enum class.  So the compiler doesn't recognize it as a valid type for the ABI boundary.
+        property int TrigUnitMode
+        {
+            void set(int value)
+            {
+                if (value != (int)m_solver->EvalOptions().GetTrigUnitMode())
+                {
+                    m_solver->EvalOptions().SetTrigUnitMode((Graphing::EvalTrigUnitMode)value);
```
joseartrivera: These APIs are not yet available correct? Could we get those changes in a PR in the impl project so that we can test this change out?

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+                    y_min->Text = m_yMin.ToString();
+                    y_max->Text = m_yMax.ToString();
+
+                    m_ParentGrapher->SetDisplayRanges(m_xMin, m_xMax, m_yMin, m_yMax);
+                }
+            }
+        }
+    }
+    catch (const std::exception&)
+    {
```
joseartrivera: I think the try/catch should only surround the call to std::stod and add some code to revert the value back or show an indicator to the user that the value is not valid input.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+        m_ParentGrapher->TrigUnitMode = (int)Graphing::EvalTrigUnitMode::Grads;
+    }
+}
+
+void GraphingSettings::UpdateExtents(Platform::Object ^ sender)
+{
+    FrameworkElement ^ fe = (FrameworkElement ^) sender;
+    FrameworkElement ^ currentFocus = (FrameworkElement ^) FocusManager::GetFocusedElement();
+
+    bool thisControlHasFocus = (fe == currentFocus);
```
joseartrivera: Would fe->FocusState != ::FocusState::Unfocused work? I am not sure of the implementation of FocusManager::GetFocusedElement() but it may be doing additional work to find the focused control when we could just check the FocusState on the control we already have.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+                    double bounds = m_width / 2;
+                    m_xMin = -bounds;
+                    m_xMax = bounds;
+                    m_ParentGrapher->XAxisMin = m_xMin;
+                    m_ParentGrapher->XAxisMax = m_xMax;
+                }
+
+                if (widthChanged || heightChanged)
+                {
+                    x_min->Text = m_xMin.ToString();
```
joseartrivera: I think this may add trailing zeroes, if it does, could we run this through NumberFormattingUtils::TrimTrailingZeros?

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+        if (tb->Name == "y_max" && newVal != m_yMax)
+        {
+            heightChanged = true;
+            m_yMax = newVal;
+            m_ParentGrapher->YAxisMax = m_yMax;
+        }
+
+        // Only the focused element get's to trigger the recompute
+
+        if (thisControlHasFocus)
```
joseartrivera: Is this check necessary? If we only call this from OnLosingFocus and OnKeyDown then I think it should only fire once for the specific control that fired the event, unless OnKeyDown fires for all controls even if they aren't the one having text inputted?

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+                }
+            }
+        }
+    }
+    catch (const std::exception&)
+    {
+    }
+}
+
+void CalculatorApp::Controls::GraphingSettings::OnLoaded(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
joseartrivera: Is this called every time this UI is opened? If not then there is a risk that this only fires once in the apps lifetime and somehow goes out of sync with the underlying data source.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+void GraphingSettings::OnKeyDown(Platform::Object ^ sender, Windows::UI::Xaml::Input::KeyRoutedEventArgs ^ e)
+{
+    if (e->Key == Windows::System::VirtualKey::Enter)
+    {
+        UpdateExtents(sender);
+    }
+}
+
+void CalculatorApp::Controls::GraphingSettings::OnPreserveAspectRatioClicked(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
joseartrivera: Nit comment for here and throughout the PR. Style of the project is to not use the full namespaces in .cpp files
```suggestion
void GraphingSettings::OnPreserveAspectRatioClicked(Object ^ sender, RoutedEventArgs ^ e)
```

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+{
+    if (m_preserveAspectRatio)
+    {
+        // On now, turning it off
+        m_preserveAspectRatio = false;
+        PreserveAspectRatioSymbol->Opacity = 0.25;
```
joseartrivera: Nit: might be good to do this in a visual state instead in case we decide we want to make more UI changes in the future based off of this we can easily add that in xaml.

src/Calculator/Calculator.vcxproj.filters
```diff
@@ -473,6 +473,9 @@
       <Filter>Views\GraphingCalculator</Filter>
     </Page>
     <Page Include="$(MSBuildThisFileDirectory)DensityStyles\Compact.xaml" />
+    <Page Include="GraphingSettings.xaml">
+      <Filter>Controls</Filter>
```
joseartrivera: This should go into the GraphingCalculator folder instead

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
+<Page
+    x:Class="CalculatorApp.Controls.GraphingSettings"
```
joseartrivera: I don't think this belongs in the Controls namespace, also can we move the file into the GraphingCalculator folder in addition to the filter?

src/Calculator/GraphingSettings.xaml.h
```diff
@@ -0,0 +1,58 @@
+//
+// MyUserControl.xaml.h
+// Declaration of the MyUserControl class
+//
+
+#pragma once
+
+#include "..\..\src\GraphingInterfaces\IMathSolver.h"
```
joseartrivera: This feels off to me.. doesn't seem like we should be going out to the src folder to grab header files from other projects, especially since other parts of the code that currently reference these header files will actually be pointing to a different header file because of the override props file.

src/GraphControl/Control/Grapher.h
```diff
@@ -190,6 +190,106 @@ public
...
+            {
+                return (int)m_solver->EvalOptions().GetTrigUnitMode();
+            }
+        }
+
+        property double XAxisMin
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().first;
```
joseartrivera: Here and the rest of these properties, it is possible that m_graph is null at various points, in the cpp file we often check for this, I think we should add some checks here also and return 0.0 if it is null or some other default value.

src/GraphControl/Control/Grapher.h
```diff
@@ -190,6 +190,106 @@ public
...
+
+        property double YAxisMax
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().second;
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(YAxisMax, value);
```
joseartrivera: Shouldn't this be passing YAxisMin in this case?

src/GraphControl/Control/Grapher.h
```diff
@@ -190,6 +190,106 @@ public
...
+
+        property double YAxisMin
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().first;
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(value, YAxisMax);
```
joseartrivera: Feels odd that we are doing this directly in the setter instead of how the other properties have a corresponding call back in the cpp file such as OnForceProportionalAxesChanged.

src/CalcManager/CalcManager.vcxproj
```diff
@@ -45,8 +45,8 @@
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
     <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
-    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == ''">10.0.19019.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformMinVersion>10.0.17134.0</WindowsTargetPlatformMinVersion>
+    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == ''">10.0.19492.0</WindowsTargetPlatformVersion>
```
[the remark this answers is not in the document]
DavidShoe: I thought we couldn't make a commitment until the SDK released with some feature we needed?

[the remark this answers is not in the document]
DavidShoe: Talked to Steph, will move to 19019.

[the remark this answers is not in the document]
DavidShoe: With min version 17134.

src/CalcViewModel/CalcViewModel.vcxproj
```diff
@@ -437,12 +437,12 @@
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
-    <Import Project="..\..\packages\Microsoft.UI.Xaml.2.0.181018003.1\build\native\Microsoft.UI.Xaml.targets" Condition="Exists('..\..\packages\Microsoft.UI.Xaml.2.0.181018003.1\build\native\Microsoft.UI.Xaml.targets')" />
+    <Import Project="..\..\packages\Microsoft.UI.Xaml.2.2.190917002\build\native\Microsoft.UI.Xaml.targets" Condition="Exists('..\..\packages\Microsoft.UI.Xaml.2.2.190917002\build\native\Microsoft.UI.Xaml.targets')" />
```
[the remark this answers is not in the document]
DavidShoe: Merge problem. Will back it out.

src/CalcViewModel/GraphingCalculator/GraphingCalculatorViewModel.h
```diff
@@ -8,6 +8,38 @@
...
+            , m_YMin(yMin)
+            , m_XMax(xMax)
+            , m_YMax(yMax)
+        {
+            OutputDebugString(L" ViewStateViewModel(...)\r\n");
+        }
+
+        OBSERVABLE_OBJECT_CALLBACK(OnPropertyChanged);
+
+        OBSERVABLE_PROPERTY_RW_ALWAYS_NOTIFY(int, XMin);
```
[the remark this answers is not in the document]
DavidShoe: Just the pattern I found for the macro, no specific reason.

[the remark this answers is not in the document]
joseartrivera: ALWAYS_NOTIFY is a macro I added as a  temporary workaround until we had access to the TrimTrailingZeroes function I mentioned in my other comment, let’s change this to just be OBSERVABLE_PROPERTY_RW

src/CalcViewModel/GraphingCalculator/GraphingCalculatorViewModel.h
```diff
@@ -8,6 +8,38 @@
...
+            OutputDebugString(L" ViewStateViewModel()\r\n");
+        }
+
+        ViewStateViewModel(int xMin, int yMin, int xMax, int yMax)
+            : m_XMin(xMin)
+            , m_YMin(yMin)
+            , m_XMax(xMax)
+            , m_YMax(yMax)
+        {
+            OutputDebugString(L" ViewStateViewModel(...)\r\n");
```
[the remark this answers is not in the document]
DavidShoe: Turned out this entire class was vestigial, removed.

src/CalcViewModel/GraphingCalculator/GraphingCalculatorViewModel.h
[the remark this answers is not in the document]
DavidShoe: Removed

DavidShoe:
Closing this PR and opening a new with better history.

PR #821, opened 2019-11-22
Modify how we manage the visibility of the history button

### LINKED ISSUE #820: The History button is visible in Programmer mode when the window is small
**Describe the bug**
The History button is visible in Programmer mode when the windows is small.

**Steps To Reproduce**
- Launch the Programmer mode
- Decrease the size of the window (if necessary)

**Expected behavior**
The button should not be visible

 - OS Build: 10.0.19013.0
 - Architecture: X64
 - Region: en-US
 - Dev Version Installed: True

**Requested Assignment**

rudyhuyn:
## Fixes #820 

FIx issue with Programmer mode

- Use a binding function instead of visual states to manage the visibility of the History button

- Manually

### DIFF
--- a/src/Calculator/Views/Calculator.xaml.cpp
+++ b/src/Calculator/Views/Calculator.xaml.cpp
@@ -714,3 +714,7 @@ void Calculator::Calculator_SizeChanged(Object ^ /*sender*/, SizeChangedEventArg
     }
 }
 
+::Visibility Calculator::HasHistory(bool isAlwaysOnTop, bool isProgrammer, ::Visibility dockPanelVisibility)
+{
+    return !isAlwaysOnTop && !isProgrammer && dockPanelVisibility == ::Visibility::Collapsed ? ::Visibility::Visible : ::Visibility::Collapsed;
+}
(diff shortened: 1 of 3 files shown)


src/Calculator/Views/Calculator.xaml.cpp
```diff
@@ -714,3 +714,7 @@ void Calculator::Calculator_SizeChanged(Object ^ /*sender*/, SizeChangedEventArg
     }
 }
 
+::Visibility Calculator::ShouldDisplayHistoryButton(bool isAlwaysOnTop, bool isProgrammer, ::Visibility dockPanelVisibility)
+{
+    return !isAlwaysOnTop && !isProgrammer && dockPanelVisibility == ::Visibility::Collapsed ? ::Visibility::Visible : ::Visibility::Collapsed;
```
marcomontoya: Is this the only mode where we don't want this button to be shown? I wonder if we should consider the scenario where this button appears in other modes it shouldn't either.

rudyhuyn: Yes. The history button should only appear when:
- in Normal mode and scientific mode, when the panel is not displayed and when not in always in top

PR #823, opened 2019-11-22
Hide the Edit Variables button when no variables are available

rudyhuyn:
- Hide the `Edit Variables` button when no variables are set.

- manually


joseartrivera:
Thank you for the fix! :)

PR #825, opened 2019-11-23
Fix prefix 0s after deleting decimal point

### LINKED ISSUE #817: Standard and Scientific modes allow multiple zeros to be prefixed
**Describe the bug**

When using the standard or scientific mode, the calculator allows multiple zeros to be prefixed inside the input field.
While it doesn't change the calculation results, this is weird behavior.

**Steps To Reproduce**
1. Enable 'Standard' or 'Scientific' mode.
2. Click the decimal point '.'
3. Press the backspace key to remove the decimal point.
4. Press the zero key multiple times. Other digits can be pressed too.
5. See error.

**Expected behavior**

After removing the decimal point (step 3 above), the initial zero should be replaced with the first non-zero digit typed.

**Screenshots**

image

image

 - OS Build: 10.0.18362.0
 - Architecture: x64
 - Region: en-US

**Additional context**

N/A

**Requested Assignment**

dovisutu:
- Added a verification if the remaining data is 0 only, and then pop this 0 out.
- 2 Tests corresponding

- Added an `UITest` and an `UnitTest`

Additional note: My respond may delay because I can't use git(hub) on weekdays.

### DIFF
--- a/src/CalcManager/CEngine/CalcInput.cpp
+++ b/src/CalcManager/CEngine/CalcInput.cpp
@@ -231,6 +231,10 @@ void CalcInput::Backspace()
         if (!m_base.IsEmpty())
         {
             m_base.value.pop_back();
+            if (m_base.value == L"0")
+            {
+                m_base.value.pop_back();
+            }
         }
 
         if (m_base.value.size() <= m_decPtIndex)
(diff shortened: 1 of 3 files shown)


dovisutu:
Anything going on? @rudyhuyn

PR #826, opened 2019-11-25
[Graph] Replace the Add button by a EquationTextBlock

rudyhuyn:
- Replace the Add button by a "ghost" EquationTextBlock
- Automatically create a new equation when 
    - the user presses Enter when the last EquationTextBox is focused
    - the last EquationTextBox loses the focus and the value has been modified
- Give the focus to the next equation when users press enter

Also
- Make sure the background of the `RichEditBox` is aligned with the Function Button
- Prevent the user to remove the `fake` `EquationTextBlock` 

- Manually with all themes and high contrast


src/Calculator/App.xaml
```diff
@@ -1419,103 +1422,74 @@
...
+                                    </VisualStateGroup>
                                 </VisualStateManager.VisualStateGroups>
 
+                                <Border x:Name="EquationTextBoxAddPanel"
+                                        x:Uid="EquationTextBoxAddPanel"
+                                        MinWidth="44"
+                                        MinHeight="44"
+                                        Background="{TemplateBinding EquationColor}"
+                                        Visibility="Collapsed">
+                                    <FontIcon Foreground="{StaticResource SystemChromeWhiteColor}"
```
joseartrivera: This should be SystemColorButtonTextColor in high contrast modes

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -90,9 +90,25 @@ void EquationInputArea::InputTextBox_Submitted(Object ^ sender, RoutedEventArgs
...
+    {
+        auto hasFocus = tb->HasFocus;
+        unsigned int index = 0;
+        if (Equations->IndexOf(eq, &index) && index == Equations->Size - 1)
+        {
+            AddNewEquation();
+        }
+        if (hasFocus)
+        {
+            FocusManager::TryMoveFocus(::FocusNavigationDirection::Left);
```
joseartrivera: I think this would feel better if we gave focus to the ghost equation box!

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -56,22 +56,22 @@ void EquationInputArea::OnEquationsPropertyChanged()
...
+        Equations->GetAt(Equations->Size - 1)->IsLastItem = false;
+    }
 
     m_lastLineColorIndex = (m_lastLineColorIndex + 1) % AvailableColors->Size;
 
     eq->LineColor = AvailableColors->GetAt(m_lastLineColorIndex);
     eq->IsLineEnabled = true;
     eq->FunctionLabelIndex = ++m_lastFunctionLabelIndex;
     Equations->Append(eq);
-    EquationInputList->ScrollIntoView(eq);
```
joseartrivera: I could see some benefit in leaving this so that the user can see that there is a ghost equation at the bottom of the scroll view when scrolling starts to occur.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -90,9 +90,25 @@ void EquationInputArea::InputTextBox_Submitted(Object ^ sender, RoutedEventArgs
...
+        }
+        if (hasFocus)
+        {
+            FocusManager::TryMoveFocus(::FocusNavigationDirection::Left);
+        }
+    }
+    else
     {
-        FocusManager::TryMoveFocus(::FocusNavigationDirection::Left);
+        if (tb->HasFocus)
```
joseartrivera: Move out the hasFocus variable from above on line 95 and use it here also

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
rudyhuyn: The scrolling cause the issue we spoke about yesterday (itemContainer recycled with the wrong value).


image

I will bring it back once EquationTextBox will be binded to the VM

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
rudyhuyn: Modified. Not only it gives the focus to the ghost equation box but I've added an extra feature:
- If you press enter on an Equation box, it will automatically give the focus to the next one (or create a new one if it's the last)

src/Calculator/App.xaml
rudyhuyn: Marco's PR will add new brushes, including one that will be used for this icon. Let's keep it like that for the moment, we will modify that after the other PR merged.

src/Calculator/Calculator.vcxproj.filters
```diff
@@ -1495,5 +1504,9 @@
   <ItemGroup>
     <CopyFileToFolders Include="$(GraphingImplDll)" />
     <CopyFileToFolders Include="$(GraphingEngineDll)" />
+    <CopyFileToFolders Include="$(GraphingImplDll)" />
```
sanderl: nit: you can remove these extra ones that were added.

src/Calculator/Utils/VisualTree.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+        if (retValue != nullptr)
+        {
+            break;
+        }
+    }
+
+    return retValue;
+}
+/*
```
sanderl: remove if this is no longer needed.

src/Calculator/Utils/VisualTree.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    if (parent->GetType() == typeName)
+    {
+        return parent;
+    }
+
+    return FindAscendant(parent, typeName);
+}
+
+/*
```
sanderl: remove if not needed

src/Calculator/Utils/VisualTree.cpp
```diff
@@ -0,0 +1,134 @@
+#include "pch.h"
```
sanderl: Please include the copyright header.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -108,6 +162,10 @@ void EquationInputArea::EquationTextBox_RemoveButtonClicked(Object ^ sender, Rou
             m_lastFunctionLabelIndex--;
         }
 
+        if (index == Equations->Size - 1 && Equations->Size > 1)
+        {
+            Equations->GetAt(Equations->Size - 2)->IsLastItem = true;
```
sanderl: Is IsLastItem the last item in the list of equations or the last item to have focus?

sanderl: Maybe adjust the name so that it is more clear.

src/Calculator/App.xaml
```diff
@@ -1419,103 +1413,74 @@
...
-                                            <Storyboard>
-                                                <ObjectAnimationUsingKeyFrames Storyboard.TargetName="EquationBoxBorder" Storyboard.TargetProperty="Background">
-                                                    <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource TextControlBackground}"/>
-                                                </ObjectAnimationUsingKeyFrames>
-                                                <ObjectAnimationUsingKeyFrames Storyboard.TargetName="EquationBoxBorder" Storyboard.TargetProperty="BorderBrush">
-                                                    <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource EquationBoxBorderBrush}"/>
-                                                </ObjectAnimationUsingKeyFrames>
-                                            </Storyboard>
+                                            <VisualState.Setters>
+                                                <Setter Target="EquationBoxBackground.Fill" Value="{ThemeResource TextControlBackground}"/>
```
sanderl: Should the ThemeResources match here? Since they are both TextControlBackground, why not use EquationBoxEquationPanelBrush for both the background fill and the borderbrush?

sanderl:
I tested out your changes. I'm not sure the '+' in the equation button works. It looks clickable, but clicking it does nothing. Even if there is an equation in the edit box, clicking the button does not submit the equation because the control still has focus. Since this button acts as a button in it's other modes, I think it should act like a button here too. It looks like the opacity has changed for the '+' but I don't think it's enough to suggest the button is disabled with the background colors.

image

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
rudyhuyn: changed! `IsLastItemInList`

rudyhuyn:
PR rebased!

joseartrivera:
For some reason on this branch when an equation has an error it no longer turns red until after I hover over it. Try typing "y=" to make an equation have an error.

rudyhuyn:
> For some reason on this branch when an equation has an error it no longer turns red until after I hover over it. Try typing "y=" to make an equation have an error.

I will change that

PR #827, opened 2019-11-25
Add error handling to graph and equations

joseartrivera:
- Updates graph control to flag equations if they fail to graph
- Updates graph control to flag equations if they successfully graph
- Graph control will not try to graph any equations that have previously failed, unless their expression has changed
- When graphing an equation that was previously successful, if the graphing fails, we will flag that equation and try graphing again without it
- Added UI to indicate that an equation has failed to graph
- Some other code tweaks

- Manual tests

### DIFF
--- a/src/Calculator/Controls/EquationTextBox.cpp
+++ b/src/Calculator/Controls/EquationTextBox.cpp
@@ -22,12 +22,12 @@ using namespace Windows::UI::Xaml::Controls::Primitives;
 DEPENDENCY_PROPERTY_INITIALIZATION(EquationTextBox, EquationColor);
 DEPENDENCY_PROPERTY_INITIALIZATION(EquationTextBox, ColorChooserFlyout);
 DEPENDENCY_PROPERTY_INITIALIZATION(EquationTextBox, EquationButtonContentIndex);
-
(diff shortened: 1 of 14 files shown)


src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -114,7 +114,7 @@ void EquationTextBox::OnKeyDown(KeyRoutedEventArgs ^ e)
 
 void EquationTextBox::OnLostFocus(RoutedEventArgs ^ e)
 {
-    if (!m_richEditBox->ContextFlyout->IsOpen)
+    if (m_richEditBox != nullptr && !m_richEditBox->ContextFlyout->IsOpen)
```
sanderl: this check needs to be added to OnKeyDown as well. I was able to crash the app because the richeditbox is null

joseartrivera: This sounds like it was caused by a local merge conflict from master, I merged and pushed again, could you give it another shot?

EriWong:
I think you need to also handle the case where the equation is disabled, and then edited.  After it becomes enabled again, it is marked as valid, even if it isn't rendered

PR #831, opened 2019-11-27
GraphControl: refactoring and optimizations

rudyhuyn:

Full refactoring of `Equation` to use only normal properties instead of dependency properties

Refactor GraphingControl
- The EquationViewModel used for KeyGraphFeatures is now hosted by `GraphingCalculatorViewModel` instead of `GraphControl`

Refactor VariableViewModel to not use PropertyChanged as an internal event
  - modify Min and Max even when Value is modified using the property
  - remove OBSERVABLE_OBJECT_CALLBACK(OnPropertyChanged)
  - replace the property `Value` and the method `SetValue`
  - replace OBSERVABLE_PROPERTY_RW_ALWAYS_NOTIFY by OBSERVABLE_PROPERTY_RW
  - extract `VariableViewModel` from `GraphingCalculatorViewModel.cpp`

Also
- use a `Color` instead of a `SolidColorBrush` for LineColor allowing us to remove the workaround using a blankBrush in `EquationInputArea`
        equationViewModel->LineColor = AvailableColors->GetAt(m_lastLineColorIndex);
- extract `KeyGraphFeatures` from `Equation`
- made many properties read-only
- rename `EquationInputAreaControl.EquationVM` as `EquationInputAreaControl.SelectedEquation`
- replace `GraphControl::PropertyChangedEventHandler` by the standard `PropertyChangedEventHandler`

- manually


src/CalcViewModel/GraphingCalculator/EquationViewModel.cpp
```diff
@@ -64,26 +66,23 @@ namespace CalculatorApp::ViewModel
 
         KeyGraphFeaturesItems->Clear();
 
-        AddKeyGraphFeature(m_resourceLoader->GetString(L"Domain"), GraphEquation->Domain, m_resourceLoader->GetString(L"KGFDomainNone"));
-        AddKeyGraphFeature(m_resourceLoader->GetString(L"Range"), GraphEquation->Range, m_resourceLoader->GetString(L"KGFRangeNone"));
-        AddKeyGraphFeature(m_resourceLoader->GetString(L"XIntercept"), GraphEquation->XIntercept, m_resourceLoader->GetString(L"KGFXInterceptNone"));
-        AddKeyGraphFeature(m_resourceLoader->GetString(L"YIntercept"), GraphEquation->YIntercept, m_resourceLoader->GetString(L"KGFYInterceptNone"));
-        AddKeyGraphFeature(m_resourceLoader->GetString(L"Minima"), GraphEquation->Minima, m_resourceLoader->GetString(L"KGFMinimaNone"));
-        AddKeyGraphFeature(m_resourceLoader->GetString(L"Maxima"), GraphEquation->Maxima, m_resourceLoader->GetString(L"KGFMaximaNone"));
+        AddKeyGraphFeature(m_resourceLoader->GetString(L"Domain"), graphEquation->Domain, m_resourceLoader->GetString(L"KGFDomainNone"));
```
rudyhuyn: Info: Nothing changed here except "GraphEquation" renamed "graphEquation"

src/CalcViewModel/Common/Utils.h
```diff
@@ -11,6 +11,8 @@
 // Utility macros to make Models easier to write
 // generates a member variable called m_<n>
 
+#define SINGLE_ARG(...) __VA_ARGS__
+
```
rudyhuyn: INFO: This is a hack to allow us to use arguments with a comma in macros

src/GraphControl/Utils.h
```diff
@@ -0,0 +1,610 @@
...
+        }                                                                                                                                                      \
+    }                                                                                                                                                          \
+                                                                                                                                                               \
+private:                                                                                                                                                       \
+    t m_##n;                                                                                                                                                   \
+                                                                                                                                                               \
+public:
+
+#define OBSERVABLE_PROPERTY_RW_ALWAYS_NOTIFY(t, n)                                                                                                             \
```
joseartrivera: Let's delete this from this file... I think we need to just remove this workaround now that we have access to TrimTrailingZeroes function after merging with master.

src/GraphControl/Models/KeyGraphFeaturesInfo.h
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+#include "Utils.h"
+#include <string>
```
joseartrivera: Here and in the rest of the files add standard headers to the pch instead

src/CalcViewModel/Common/Utils.h
```diff
@@ -11,6 +11,8 @@
 // Utility macros to make Models easier to write
```
joseartrivera: Can we remove OBSERVABLE_PROPERTY_RW_ALWAYS_NOTIFY from this file?

src/CalcViewModel/GraphingCalculator/VariableViewModel.h
```diff
@@ -0,0 +1,69 @@
...
+            , m_SliderSettingsVisible(false)
+            , m_Min(0.0)
+            , m_Step(0.1)
+            , m_Max(2.0)
+        {
+        }
+
+        OBSERVABLE_OBJECT();
+        OBSERVABLE_PROPERTY_R(Platform::String ^, Name);
+        OBSERVABLE_PROPERTY_RW(double, Min);
```
joseartrivera: Since the work around is removed, we will need to pass the values through TrimTrailingZeroes in GraphingCalculator.xaml.cpp in the GraphingCalculator::SubmitTextbox function.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -173,17 +168,11 @@ void EquationInputArea::ReloadAvailableColors(bool isHighContrast)
         return;
     }
 
-    // Use a blank brush to clear out the color before setting it. This is needed because going
```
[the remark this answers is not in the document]
rudyhuyn: The issue was due to the brush not changing when you use a different High Contrast color set (only the property SolidColorBrush.Color was updated). Using a Color instead of a Brush fixes this issue

src/GraphControl/Utils.h
rudyhuyn: removed!

src/GraphControl/Models/KeyGraphFeaturesInfo.h
rudyhuyn: cleaned!

src/CalcViewModel/Common/Utils.h
rudyhuyn: done

src/CalcViewModel/GraphingCalculator/VariableViewModel.h
rudyhuyn: Good catch!

I did a little different to manage all cases:
- 10.0000 -> 10
- 0000001 -> 1
- 1212. ->1212 
- '     12' -> 12
etc...

Instead of removing the trailing zeros, I convert the double back to string and replace the value.

rudyhuyn:
rebased!

PR #834, opened 2019-11-30
Remove adjustment from local time

### LINKED ISSUE #792: Date Calculator showing wrong result in some instances when only one field is changed
The calculator app shows difference between yesterday and today as same

1. Open calculator and change to date calculation.
2. By default, From and To are populated to Today's date.
3. Change From to yesterday.

You will notice that the date difference is still "Same dates" which is incorrect. See screenshot. It looks like the bug repros when you change only one of the fields - either From or To. When both are changed, it appears to be working correctly.

image

j-windsor:
It looks like the CalendarDatePicker control now provides a UTC Date as the value, making it so the conditions in #557 do not apply anymore. The conversion from local time actually pushes the date forward a day. Changes due to #557 are reverted here.

Tested Manually

### DIFF
--- a/src/CalcViewModel/DateCalculatorViewModel.cpp
+++ b/src/CalcViewModel/DateCalculatorViewModel.cpp
@@ -60,7 +60,7 @@ DateCalculatorViewModel::DateCalculatorViewModel()
     auto today = calendar->GetDateTime();
 
     // FromDate and ToDate should be clipped (adjusted to a consistent hour in UTC)
-    m_fromDate = m_toDate = ClipTime(today, true);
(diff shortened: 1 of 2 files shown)


rudyhuyn:
Good find and thank you for taking care of that!

I will check if all versions of Windows 10 supported by the calculator return a UTC time.

In the meantime, instead of reverting the previous PR, can you change all ClipTime calls to set adjustUsingLocalTime to False (in case we have a regression in the future). Thank you!

j-windsor:
Thanks @rudyhuyn! I have added back the behavior and set the default `adjustUsingLocalTime = false` for readability and signature simplicity.

PR #841, opened 2019-12-03
Build GraphingCalculator against SDK 18362

mcooley:
Allow the feature/GraphingCalculator branch to build without having a pre-release SDK installed on the build machines.
- Retarget projects back to 18362
- For the few APIs we need to use which are newer than 18362, use COM interop to call the methods directly

This change can be reverted once the Windows 10 version 2004 SDK is released and installed on the Azure Pipelines hosted build agents.

### DIFF
--- a/src/Calculator/Controls/MathRichEditBox.cpp
+++ b/src/Calculator/Controls/MathRichEditBox.cpp
@@ -16,6 +16,19 @@ using namespace Windows::Foundation::Collections;
 
 DEPENDENCY_PROPERTY_INITIALIZATION(MathRichEditBox, MathText);
 
+// TODO remove when Windows 10 version 2004 SDK is adopted
+namespace Windows_2004_Prerelease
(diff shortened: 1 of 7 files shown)


src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -16,6 +16,19 @@ using namespace Windows::Foundation::Collections;
 
 DEPENDENCY_PROPERTY_INITIALIZATION(MathRichEditBox, MathText);
 
+// TODO remove when Windows 10 version 2004 SDK is adopted
```
sanderl: Is this the right version number? Should it be 20004?

mcooley: "2004" is the product name that a user would see in the Settings app (April, 20**20**). Final build number not yet known.

PR #847, opened 2019-12-04
Add Graph Settings

DavidShoe:
Adds a settings button which controls a flyout panel allowing the user to control the graph extents and the trig unit mode.

New control

-
-

Manual, build debug/release.


src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
+<Page
```
joseartrivera: Should this be a UserControl?

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
...
+            <Setter Property="FocusVisualMargin" Value="0" />
+            <Setter Property="Template">
+                <Setter.Value>
+                    <ControlTemplate TargetType="RadioButton">
+                        <Grid
+                        Name="LayoutRoot"
+                        Margin="0,0"
+                        Background="{ThemeResource RadioButtonBackground}">
+                            <Border
+                            Name="AccessibilityBorder"
```
joseartrivera: Spacing is off, do you have XamlStyler installed? It should handle most of this stuff automatically, check out instructions on [README.md](https://github.com/microsoft/calculator/blob/master/README.md)

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
+<Page
+    x:Class="CalculatorApp.Controls.GraphingSettings"
```
joseartrivera: I don't think this belongs in the Controls namespace, also can we move the file into the GraphingCalculator folder in addition to the filter?

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
...
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:CalculatorApp"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    DataContextChanged="GraphingCalculator_DataContextChanged"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}" Loaded="OnLoaded"
+    >
+    <Page.Resources>
```
joseartrivera: I don't think many of these colors will work with high contrast. We should have a ResourceDictionary for each theme, check out App.xaml for an example on handling high contrast brushes.

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
...
+                        </Grid>
+                    </ControlTemplate>
+                </Setter.Value>
+            </Setter>
+        </Style>
+    </Page.Resources>
+
+    <Grid>
+        <StackPanel Margin="10,10,10,10">
+            <TextBlock x:Name="GridHeading" Text="Grid_d" FontSize="28"/>
```
joseartrivera: I think we just delete the Text="Grid_d" from here and the rest of the TextBlocks. Also I think assigning the localized strings would be better if we use x:Uid instead. See my other comment in the code that assigns the localize strings for more details.

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
...
+            <Grid>
+                <Grid.RowDefinitions>
+                    <RowDefinition Height="30"/>
```
joseartrivera: You should use MinHeight here so that when the user changes their font size from accessibility settings the rows automatically change their height

src/Calculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,136 @@
...
+                </Grid.RowDefinitions>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="150"/>
+                    <ColumnDefinition Width="150"/>
+                    <ColumnDefinition Width="70"/>
+                </Grid.ColumnDefinitions>
+
+                <TextBlock Grid.Row="0" Grid.Column="0" x:Name="x_MinHeading" Text="x-min_d" VerticalAlignment="Center"/>
+                <TextBlock Grid.Row="0" Grid.Column="1" x:Name="x_MaxHeading" Text="x-max_d" VerticalAlignment="Center"/>
+                <TextBox Grid.Row="1" Grid.Column="0" x:Name="x_min" LosingFocus="OnLosingFocus" KeyDown="OnKeyDown" VerticalAlignment="Center"/>
```
joseartrivera: I think we should remain consistent with the casing and call this y_Min, same comment below for y_Max, etc.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
+#include "pch.h"
+
+#include "GraphingSettings.xaml.h"
+#include <CalcViewModel\Common\AppResourceProvider.cpp>
```
joseartrivera: nit:
```suggestion
#include "CalcViewModel\Common\AppResourceProvider.cpp"
```

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+using namespace Windows::UI::Xaml::Media;
+using namespace Windows::UI::Xaml::Navigation;
+
+GraphingSettings::GraphingSettings()
+{
+    InitializeComponent();
+
+    auto resLoader = AppResourceProvider::GetInstance();
+
+    x_MinHeading->Text = resLoader.GetResourceString(L"x-min");
```
joseartrivera: If these never change, it would be cleaner and consistent with the rest of the project to assign them in xaml instead by changing the resource name from x-min -> x-min.Text and setting the TextBlock Uid to x:Uid="x-min". Same comment for the rest of this function.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    if (Gradians->IsChecked->Value == true)
+    {
+        m_ParentGrapher->TrigUnitMode = (int)Graphing::EvalTrigUnitMode::Grads;
+    }
+}
+
+void GraphingSettings::UpdateExtents(Platform::Object ^ sender)
+{
+    FrameworkElement ^ fe = (FrameworkElement ^) sender;
```
joseartrivera: nit, here and in the rest of the file we should use static_cast/dynamic_cast instead of c-style casts
```suggestion
    auto ^ fe = static_cast<FrameworkElement ^>(sender);
```

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+    FrameworkElement ^ currentFocus = (FrameworkElement ^) FocusManager::GetFocusedElement();
+
+    bool thisControlHasFocus = (fe == currentFocus);
```
joseartrivera: Would fe->FocusState != ::FocusState::Unfocused work? I am not sure of the implementation of FocusManager::GetFocusedElement() but it may be doing additional work to find the focused control when we could just check the FocusState on the control we already have.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+    TextBox ^ tb = reinterpret_cast<TextBox ^>(sender);
```
joseartrivera: Any reason this has to be reinterpret_cast? If not, I think we should change this to be static_cast instead.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+        if (tb->Name == "y_max" && newVal != m_yMax)
+        {
+            heightChanged = true;
+            m_yMax = newVal;
+            m_ParentGrapher->YAxisMax = m_yMax;
+        }
+
+        // Only the focused element get's to trigger the recompute
+
+        if (thisControlHasFocus)
```
joseartrivera: Is this check necessary? If we only call this from OnLosingFocus and OnKeyDown then I think it should only fire once for the specific control that fired the event, unless OnKeyDown fires for all controls even if they aren't the one having text inputted?

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+                    double bounds = m_width / 2;
+                    m_xMin = -bounds;
+                    m_xMax = bounds;
+                    m_ParentGrapher->XAxisMin = m_xMin;
+                    m_ParentGrapher->XAxisMax = m_xMax;
+                }
+
+                if (widthChanged || heightChanged)
+                {
+                    x_min->Text = m_xMin.ToString();
```
joseartrivera: I think this may add trailing zeroes, if it does, could we run this through NumberFormattingUtils::TrimTrailingZeros?

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+                    y_min->Text = m_yMin.ToString();
+                    y_max->Text = m_yMax.ToString();
+
+                    m_ParentGrapher->SetDisplayRanges(m_xMin, m_xMax, m_yMin, m_yMax);
+                }
+            }
+        }
+    }
+    catch (const std::exception&)
```
joseartrivera: I think the try/catch should only surround the call to std::stod and add some code to revert the value back or show an indicator to the user that the value is not valid input.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+    {
+    }
+}
+
+void CalculatorApp::Controls::GraphingSettings::OnLoaded(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
joseartrivera: Is this called every time this UI is opened? If not then there is a risk that this only fires once in the apps lifetime and somehow goes out of sync with the underlying data source.

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+void GraphingSettings::OnKeyDown(Platform::Object ^ sender, Windows::UI::Xaml::Input::KeyRoutedEventArgs ^ e)
+{
+    if (e->Key == Windows::System::VirtualKey::Enter)
+    {
+        UpdateExtents(sender);
+    }
+}
+
+void CalculatorApp::Controls::GraphingSettings::OnPreserveAspectRatioClicked(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
joseartrivera: Nit comment for here and throughout the PR. Style of the project is to not use the full namespaces in .cpp files
```suggestion
void GraphingSettings::OnPreserveAspectRatioClicked(Object ^ sender, RoutedEventArgs ^ e)
```

src/Calculator/GraphingSettings.xaml.cpp
```diff
@@ -0,0 +1,220 @@
...
+{
+    if (m_preserveAspectRatio)
+    {
+        // On now, turning it off
+        m_preserveAspectRatio = false;
+        PreserveAspectRatioSymbol->Opacity = 0.25;
```
joseartrivera: Nit: might be good to do this in a visual state instead in case we decide we want to make more UI changes in the future based off of this we can easily add that in xaml.

src/Calculator/GraphingSettings.xaml.h
```diff
@@ -0,0 +1,55 @@
+//
+// MyUserControl.xaml.h
+// Declaration of the MyUserControl class
+//
+
+#pragma once
+
+#include "..\..\src\GraphingInterfaces\IMathSolver.h"
```
joseartrivera: This feels off to me. doesn't seem like we should be going out to the src folder to grab header files from other projects, especially since other parts of the code that currently reference these header files will actually be pointing to a different header file because of the override props file.

src/GraphControl/Control/Grapher.h
```diff
@@ -148,6 +148,104 @@ public
...
         void AnalyzeEquation(GraphControl::Equation ^ equation);
+       // We can't use the EValTrigUnitMode enum directly in as the property type because it comes from another module which doesn't expose
+        // it as a public enum class.  So the compiler doesn't recognize it as a valid type for the ABI boundary.
+        property int TrigUnitMode
+        {
+            void set(int value)
+            {
+                if (value != (int)m_solver->EvalOptions().GetTrigUnitMode())
+                {
+                    m_solver->EvalOptions().SetTrigUnitMode((Graphing::EvalTrigUnitMode)value);
```
joseartrivera: These APIs are not yet available correct? Could we get those changes in a PR in the impl project so that we can test this change out?

src/GraphControl/Control/Grapher.h
```diff
@@ -148,6 +148,104 @@ public
...
+            {
+                return (int)m_solver->EvalOptions().GetTrigUnitMode();
+            }
+        }
+
+        property double XAxisMin
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().first;
```
joseartrivera: Here and the rest of these properties, it is possible that m_graph is null at various points, in the cpp file we often check for this, I think we should add some checks here also and return 0.0 if it is null or some other default value.

src/GraphControl/Control/Grapher.h
```diff
@@ -148,6 +148,104 @@ public
...
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(value, XAxisMax);
```
joseartrivera: Feels odd that we are doing this directly in the setter instead of how the other properties have a corresponding call back in the cpp file such as OnForceProportionalAxesChanged.

src/GraphControl/Control/Grapher.h
```diff
@@ -148,6 +148,104 @@ public
...
+
+        property double XAxisMax
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().second;
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(XAxisMax, value);
```
joseartrivera: Shouldn't this be passing YAxisMin in this case?

src/Calculator/GraphingSettings.xaml
DavidShoe: The text assignments are just meant to be design time so that when looking at the control in the editor it isn't just a bunch of blanks.

src/Calculator/GraphingSettings.xaml
DavidShoe: Could be but doesn't have to be, it's never embedded into another page just used for the flyout.

src/Calculator/GraphingSettings.xaml
DavidShoe: I didn't, was just using default VS handling.  Fixed.

src/Calculator/GraphingSettings.xaml
DavidShoe: Fixed

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Fixed

src/Calculator/GraphingSettings.xaml.h
DavidShoe: Unneeded, and removed.

src/GraphControl/Control/Grapher.h
DavidShoe: Steph has them coming in with the build system stuff that's going in. So they will be live shortly.

src/GraphControl/Control/Grapher.h
DavidShoe: Yep, good catch.  Both the X & Y were wrong.  Fixed.

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Never saw it add trailing zeros.

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Because we call update extents on the other axis controls potentially (when aspect ratio locked) so this code keeps us from going recursive.

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Changed

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: FIxed, VS autogenerated method I forgot to cleanup.

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Looks like it.  Changed.

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Reverting the changes would be odd for the user if they have good typing skills and they are already doing the muscle memory to fix the error as they type it.  @grochocki What do you think?

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Yep, I should have gone back and done these after I learned this lesson in the tracing PR.  Fixed.

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Changed.  Interesting, had not used the x:Uid automatic resource replacement before.

src/Calculator/GraphingSettings.xaml
DavidShoe: The localized string by way of x:Uid is done.  But still left the design time strings there.

src/Calculator/GraphingSettings.xaml
DavidShoe: Done.

src/Calculator/GraphingSettings.xaml
DavidShoe: Done

src/Calculator/GraphingSettings.xaml.cpp
DavidShoe: Done

src/Calculator/GraphingSettings.xaml.cpp
joseartrivera: That's a good point, I think we should just show an error icon and maybe some red background then.

PR #848, opened 2019-12-05
Hide or Disable graphing calculator menu item when necessary

rudyhuyn:
Hide the graphing calculator menu item when the OS doesn't support the feature or disable it when a group policy is set.

- Hide the Graphing Calculator menu when running Windows versions prior to Vibranium
- Disable the Graphing Calculator menu when a group policy is set 
- Switch to the default view (Standard mode) when Graphing Calculator was the last used feature but the feature was disable later by the group policy
- modify accelerator keys of Programmer mode and Date Calculation if the Graphing Calculator menu item is visible or not (`3 and 4` or `4 and 5`)

Also:
- simplify `NavCategory` class
- remove NavCategory::Position (never used)

- manually

### DIFF
--- a/src/CalcViewModel/Common/NavCategory.cpp
+++ b/src/CalcViewModel/Common/NavCategory.cpp
@@ -52,17 +52,18 @@ wchar_t* towchar_t(int number)
 
 extern "C"
 {
-    WINADVAPI LSTATUS APIENTRY RegOpenKeyW(_In_ HKEY hKey, _In_opt_ LPCWSTR lpSubKey, _Out_ PHKEY phkResult);
-#define RegOpenKey RegOpenKeyW
-    WINADVAPI LSTATUS APIENTRY RegQueryValueExW(
-        _In_ HKEY hKey,
-        _In_opt_ LPCWSTR lpValueName,
-        _Reserved_ LPDWORD lpReserved,
-        _Out_opt_ LPDWORD lpType,
-        _Out_writes_bytes_to_opt_(*lpcbData, *lpcbData) __out_data_source(REGISTRY) LPBYTE lpData,
-        _When_(lpData == NULL, _Out_opt_) _When_(lpData != NULL, _Inout_opt_) LPDWORD lpcbData);
-    WINADVAPI LSTATUS APIENTRY RegCloseKey(_In_ HKEY hKey);
-#define RegQueryValueEx RegQueryValueExW
+    WINADVAPI LSTATUS APIENTRY RegGetValueW(
+        _In_ HKEY hkey,
+        _In_opt_ LPCWSTR lpSubKey,
+        _In_opt_ LPCWSTR lpValue,
+        _In_ DWORD dwFlags,
+        _Out_opt_ LPDWORD pdwType,
+        _When_(
+            (dwFlags & 0x7F) == RRF_RT_REG_SZ || (dwFlags & 0x7F) == RRF_RT_REG_EXPAND_SZ || (dwFlags & 0x7F) == (RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ)
+                || *pdwType == REG_SZ || *pdwType == REG_EXPAND_SZ,
+            _Post_z_) _When_((dwFlags & 0x7F) == RRF_RT_REG_MULTI_SZ || *pdwType == REG_MULTI_SZ, _Post_ _NullNull_terminated_)
+            _Out_writes_bytes_to_opt_(*pcbData, *pcbData) PVOID pvData,
+        _Inout_opt_ LPDWORD pcbData);
 }
 
 bool IsGraphingModeAvailable()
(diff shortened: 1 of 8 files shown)


src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -49,151 +44,258 @@ static constexpr int CURRENCY_ID = 16;
...
+    DWORD bufferSize{ sizeof(allowGraphingCalculator) };
+    // Make sure to call RegOpenKey only on Windows 10 1903+
+    if (RegOpenKey(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\Calculator"), &key) != ERROR_SUCCESS)
+    {
+        _isGraphingModeEnabledCached = true;
+        return true;
+    }
+    if (RegQueryValueEx(key, TEXT("AllowGraphingCalculator"), NULL, NULL, reinterpret_cast<LPBYTE>(&allowGraphingCalculator), &bufferSize) != ERROR_SUCCESS)
+    {
+        RegCloseKey(key);
```
mcooley: If RegOpenKey succeeds but RegQueryValueEx does not succeed, you'll leak the handle.

We could consider taking a dependency on the wil library (github.com/microsoft/wil) which provides a unique_hkey RAII type.

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -49,151 +44,258 @@ static constexpr int CURRENCY_ID = 16;
...
+    HKEY key;
+    DWORD allowGraphingCalculator{ 0 };
+    DWORD bufferSize{ sizeof(allowGraphingCalculator) };
+    // Make sure to call RegOpenKey only on Windows 10 1903+
+    if (RegOpenKey(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\Calculator"), &key) != ERROR_SUCCESS)
+    {
+        _isGraphingModeEnabledCached = true;
+        return true;
+    }
+    if (RegQueryValueEx(key, TEXT("AllowGraphingCalculator"), NULL, NULL, reinterpret_cast<LPBYTE>(&allowGraphingCalculator), &bufferSize) != ERROR_SUCCESS)
```
mcooley: Consider using RegGetValueW instead:

RegGetValueW(HKEY_LOCAL_MACHINE, L"Software\\...\\\Calculator", L"AllowGraphingCalculator", RRF_RT_DWORD, ...)

That way you can do this in a single call and avoid dealing with handle lifetime management. Also you can enforce that the key is of the expected type.

src/CalcViewModel/Common/NavCategory.cpp
mcooley: But actually just use RegGetValueW.

src/CalcViewModel/Common/NavCategory.cpp
mcooley: Also, don't use the TEXT() macro in code that doesn't build for both Unicode and ASCII. Just use L""

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -49,151 +44,260 @@ static constexpr int CURRENCY_ID = 16;
...
+            RRF_RT_REG_DWORD,
+            nullptr,
+            reinterpret_cast<LPBYTE>(&allowGraphingCalculator),
+            &bufferSize) != ERROR_SUCCESS)
+    {
+        _isGraphingModeEnabledCached = true;
+        return true;
+    }
+
+    _isGraphingModeEnabledCached = allowGraphingCalculator != 0;
```
EriWong: >allowGraphingCalculator [](start = 35, length = 23)

I think there's a bug here...  I do not have the reg key set but allowGraphingCalculator is set to 0 which is graying out the option.

rudyhuyn: Fixed!

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -49,151 +44,263 @@ static constexpr int CURRENCY_ID = 16;
...
+    }
+
+    DWORD allowGraphingCalculator{ 0 };
+    DWORD bufferSize{ sizeof(allowGraphingCalculator) };
+    // Make sure to call RegGetValueW only on Windows 10 1903+
+    if (RegGetValueW(
+            HKEY_LOCAL_MACHINE,
+            L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\Calculator",
+            L"AllowGraphingCalculator",
+            RRF_RT_REG_DWORD | RRF_RT_REG_BINARY,
```
mcooley: RRF_RT_DWORD

PR #850, opened 2019-12-05
Refactoring LocalizationSettings

### LINKED ISSUE #849: LocalizationSettings.h could use code clean up and unit tests
- LocalizationSettings.h doesn't currently have unit tests.
- There are many instances where an algorithm such as find or copy can be used instead of manual implementations.
- m_resolvedName can be changed to be a Platform::String to prevent the need to create a new one each time GetLocaleName is called.
- GetDecimalSeparatorStr and GetNumberGroupingSeparatorStr could be made more efficient by using the constructor taking a wchar_t instead of default constructing a string and pushing back a character.

gsfreema:
- Provided public constructor taking a DecimalFormatter for unit testing specific locales.
- Changed m_resolvedName to be a Platform::String ^ so it doesn't need to create a new one each time.
- Used algorithms over raw loops.
- Return strings of a single character created by constructor instead of constructor and push_back call.

- Added unit tests for much of LocalizatonSettings.

### DIFF
--- a/src/CalcViewModel/Common/LocalizationSettings.h
+++ b/src/CalcViewModel/Common/LocalizationSettings.h
@@ -251,7 +251,7 @@ namespace CalculatorApp
                 return m_numberGrouping;
             }
 
-            Platform::String^ RemoveGroupSeparators(Platform::String^ source) const
+            Platform::String ^ RemoveGroupSeparators(Platform::String ^ source) const
             {
                 std::wstring destination;
                 std::copy_if(
--- a/src/CalcViewModel/Common/LocalizationSettings.h
+++ b/src/CalcViewModel/Common/LocalizationSettings.h
@@ -14,45 +14,48 @@ namespace CalculatorApp
         {
         private:
             LocalizationSettings()
-            {
-                int result = 0;
-
                 // Use DecimalFormatter as it respects the locale and the user setting
-                Windows::Globalization::NumberFormatting::DecimalFormatter ^ formatter;
-                formatter = LocalizationService::GetInstance()->GetRegionalSettingsAwareDecimalFormatter();
+                : LocalizationSettings(LocalizationService::GetInstance()->GetRegionalSettingsAwareDecimalFormatter())
+            {
+            }
+
+        public:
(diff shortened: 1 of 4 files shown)


src/CalcViewModel/Common/LocalizationSettings.h
```diff
@@ -176,56 +174,39 @@ namespace CalculatorApp
                     return localizedString;
                 }
 
-                size_t i = 0;
-                auto localizedStringData = localizedString->Data();
-                size_t length = localizedString->Length();
-                std::unique_ptr<wchar_t[]> englishString(new wchar_t[length + 1]); // +1 for the null termination
+                std::wstring englishString;
+                englishString.reserve(localizedString->Length());
```
rudyhuyn: very minor: can we use a stringstream instead?

src/CalcViewModel/Common/LocalizationSettings.h
```diff
@@ -265,34 +238,26 @@ namespace CalculatorApp
...
+                return std::wstring(1, m_numberGroupSeparator);
             }
 
             std::wstring GetNumberGroupingStr() const
             {
                 return m_numberGrouping;
             }
 
-            Platform::String ^ RemoveGroupSeparators(Platform::String ^ source) const
+            Platform::String^ RemoveGroupSeparators(Platform::String^ source) const
```
rudyhuyn: Can you revert this line to stay consistent with the rest of the code?

src/CalcViewModel/Common/LocalizationSettings.h
gsfreema: I don't mind making this change, but since C++ isn't the language I use for my day job, I'm curious why a std::wstringstream is the better choice in this situation.  I put together a simple function to remove spaces on compiler explorer and clang is able to generate way less assembly using std::wstring over std::wstringstream.
https://godbolt.org/z/PjXGFW

src/CalcViewModel/Common/LocalizationSettings.h
gsfreema: Done

PR #854, opened 2019-12-06
Add context menu to rich edit to allow for keyboard support

EriWong:
- Adds context menu for rich edit that allows for showing key graph features, changing colors of equations, removing equations.

manually tested keyboard only interactions
manually tested mouse interactions

### DIFF
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -1890,7 +1890,7 @@
 
                                 <VisualStateManager.VisualStateGroups>
                                     <VisualStateGroup x:Name="CommonStates">
-                                        
+
                                         <VisualState x:Name="Normal">
                                             <Storyboard>
                                                 <ObjectAnimationUsingKeyFrames Storyboard.TargetName="EquationBoxBorder" Storyboard.TargetProperty="BorderBrush">
@@ -2179,7 +2179,27 @@
                                                                   AcceptsReturn="false"
                                                                   InputScope="Text"
                                                                   MaxLength="2048"
-                                                                  TextWrapping="NoWrap"/>
+                                                                  TextWrapping="NoWrap">
+                                            <Controls:MathRichEditBox.ContextFlyout>
+                                                <MenuFlyout x:Name="MathRichEditContextMenu">
+                                                    <MenuFlyoutItem x:Name="FunctionAnalysisMenuItem">
+                                                        <MenuFlyoutItem.Icon>
+                                                            <FontIcon FontFamily="{StaticResource CalculatorFontFamily}" Glyph="&#xE3B5;"/>
+                                                        </MenuFlyoutItem.Icon>
+                                                    </MenuFlyoutItem>
+                                                    <MenuFlyoutItem x:Name="ChangeFunctionStyleMenuItem">
+                                                        <MenuFlyoutItem.Icon>
+                                                            <FontIcon FontFamily="{StaticResource CalculatorFontFamily}" Glyph="&#xE790;"/>
+                                                        </MenuFlyoutItem.Icon>
+                                                    </MenuFlyoutItem>
+                                                    <MenuFlyoutItem x:Name="RemoveFunctionMenuItem">
+                                                        <MenuFlyoutItem.Icon>
+                                                            <FontIcon FontFamily="{StaticResource CalculatorFontFamily}" Glyph="&#xECC9;"/>
+                                                        </MenuFlyoutItem.Icon>
+                                                    </MenuFlyoutItem>
+                                                </MenuFlyout>
+                                            </Controls:MathRichEditBox.ContextFlyout>
+                                        </Controls:MathRichEditBox>
                                         <!-- TODO: Use brush overrides here instead of a new style, use a new style for the EquationButton above once that has more functionality -->
                                         <Button x:Name="DeleteButton"
                                                 Grid.Column="3"
@@ -2197,14 +2217,14 @@
                                                 IsTabStop="False"
                                                 Visibility="Collapsed"/>
                                         <FontIcon x:Name="ErrorIcon"
-                                                Grid.Column="3"
-                                                MinWidth="34"
-                                                VerticalAlignment="Stretch"
-                                                FontFamily="{ThemeResource SymbolThemeFontFamily}"
-                                                AutomationProperties.AccessibilityView="Raw"
-                                                Glyph="&#xE783;"
-                                                FontSize="16"
-                                                Visibility="Collapsed"/>
+                                                  Grid.Column="3"
+                                                  MinWidth="34"
+                                                  VerticalAlignment="Stretch"
+                                                  FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                                                  FontSize="16"
+                                                  AutomationProperties.AccessibilityView="Raw"
+                                                  Glyph="&#xE783;"
+                                                  Visibility="Collapsed"/>
                                         <Button x:Name="RemoveButton"
                                                 x:Uid="removeButton"
                                                 Grid.Column="3"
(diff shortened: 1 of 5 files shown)


src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -114,9 +119,13 @@ void EquationInputArea::EquationTextBox_RemoveButtonClicked(Object ^ sender, Rou
 
 void EquationInputArea::EquationTextBox_KeyGraphFeaturesButtonClicked(Object ^ sender, RoutedEventArgs ^ e)
 {
+    // ensure the equation has been submitted before trying to get key graph features out of it
+    EquationInputArea::InputTextBox_Submitted(sender, e);
```
joseartrivera: I think this would degrade the KGF analysis performance since we have to wait for the graph again, is this necessary? I think it would still work without this and also graph if it hasn't already due to losing focus.

src/Calculator/App.xaml
```diff
@@ -2179,7 +2179,27 @@
                                                                   AcceptsReturn="false"
                                                                   InputScope="Text"
                                                                   MaxLength="2048"
-                                                                  TextWrapping="NoWrap"/>
+                                                                  TextWrapping="NoWrap">
+                                            <Controls:MathRichEditBox.ContextFlyout>
+                                                <MenuFlyout x:Name="MathRichEditContextMenu">
```
joseartrivera: Do we need to add access/shortcut keys?

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -323,3 +352,11 @@ bool EquationTextBox::ShouldDeleteButtonBeVisible()
     }
     return (!text->IsEmpty() && m_HasFocus);
 }
+
+void EquationTextBox::OnRichEditMenuOpening(Platform::Object ^ sender, Platform::Object ^ args)
+{
+    if (m_kgfEquationMenuItem != nullptr)
+    {
+        m_kgfEquationMenuItem->IsEnabled = EquationTextBox::RichEditHasContent();
```
joseartrivera: KGF button should also be disabled/collapsed when the equation is in an error state but would still have richedit content.

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -50,13 +56,18 @@ void EquationTextBox::OnApplyTemplate()
...
-         auto resProvider = AppResourceProvider::GetInstance();
          auto equationButtonMessage = m_equationButton->IsChecked->Value ? resProvider->GetResourceString(L"showEquationButtonToolTip") : resProvider->GetResourceString(L"hideEquationButtonToolTip");
          toolTip->Content = equationButtonMessage;
          ToolTipService::SetToolTip(m_equationButton, toolTip);
          AutomationProperties::SetName(m_equationButton, equationButtonMessage);
     }
 
+    if (m_richEditContextMenu != nullptr)
+    {
+        m_richEditContextMenu->Opening +=
```
joseartrivera: Visual state bug:
1. Type "y=mx + b" but don't submit
2. Right click text box and click analyze equation
3. Click back button on KGF page
4. Equation text box is in the focus visual state even though it does not have focus

src/Calculator/Controls/EquationTextBox.cpp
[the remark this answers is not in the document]
EriWong: This is kind of intentional because I think having disappearing UI in this case can be confusing as there isn't really any other UI hinting at what should be shown and what shouldn't.

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -323,3 +352,11 @@ bool EquationTextBox::ShouldDeleteButtonBeVisible()
     }
     return (!text->IsEmpty() && m_HasFocus);
 }
+
+void EquationTextBox::OnRichEditMenuOpening(Platform::Object ^ sender, Platform::Object ^ args)
```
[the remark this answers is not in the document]
EriWong: good catch.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -88,7 +88,12 @@ void EquationInputArea::InputTextBox_Submitted(Object ^ sender, RoutedEventArgs
 {
     auto tb = static_cast<EquationTextBox ^>(sender);
     auto eq = static_cast<EquationViewModel ^>(tb->DataContext);
-    eq->Expression = tb->GetEquationText();
+
+    // eq can be null if the equation has been removed
```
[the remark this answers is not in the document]
EriWong: Good thought, but according to https://docs.microsoft.com/en-us/cpp/cpp/static-cast-operator?view=vs-2019

The static_cast operator converts a null pointer value to the null pointer value of the destination type.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
EriWong: So I really wanted to have the experience be "Enter the equation, hit the context menu key, and then hit enter".  Without triggering submitted, it would not actually update the method, at which point the eq object wouldn't have the right equation.  Because the context menu is on the rich edit, focus isn't leaving the rich edit and thus the blur code is not being run...  Agreed about the perf implications though.  I'm not entirely sure on what the right thing to do is though.

src/Calculator/Controls/EquationTextBox.cpp
EriWong: Unfortunately, we don't know if it's in an error state or not because we'd have to calculate it on keystroke, which would be heavy.  This is something of a compromise where we will enable it whenever there is a chance that it is a valid equation.

src/Calculator/Controls/EquationTextBox.cpp
EriWong: I spoke with Rudy about this, and he has a separate fix for this that manages focus on transition to and from KGF.

src/Calculator/App.xaml
```diff
@@ -2179,7 +2179,27 @@
                                                                   AcceptsReturn="false"
                                                                   InputScope="Text"
                                                                   MaxLength="2048"
-                                                                  TextWrapping="NoWrap"/>
+                                                                  TextWrapping="NoWrap">
+                                            <Controls:MathRichEditBox.ContextFlyout>
```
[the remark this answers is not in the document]
EriWong: good thought.  Will add those.

src/Calculator/App.xaml
EriWong: We don't have these defined yet.  That said, we should add them, but likely as a follow up to this.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
EriWong: added a check to see if we have focus.  If we do, then we submit otherwise we don't.  I think that this avoids calling the graph in cases where we know the graph is up to date.

src/Calculator/App.xaml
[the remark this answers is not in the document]
EriWong: Turns out we can intercept the original context menu opening event and then add in our own to the returned result, but I'm seeing some crashes when I try this in C++ (though it appears to work in C# prototypes).  I think I want to do this as a followup though as it might take a bit of engagement with the xaml team to figure out why the C++ code doesn't work.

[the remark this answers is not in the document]
joseartrivera: Sounds good

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
joseartrivera: Good idea!

PR #855, opened 2019-12-06
Add graph mode keyboard UI

joseartrivera:
- Adds the graphing mode keyboard UI, does not yet function, that will be added in a separate PR
- Added Trigonometry, Inequality, and Function operator panels
- Added visual states for small, medium, and large screens
- Updated icons, tab order, and enabled xy navigation

- Manual testing

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
@@ -1,0 +1,1194 @@
+<UserControl x:Class="CalculatorApp.GraphingNumPad"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:Windows10version1803="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractNotPresent(Windows.Foundation.UniversalApiContract, 7)"
+             xmlns:Windows10version1809="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 7)"
+             xmlns:controls="using:CalculatorApp.Controls"
+             xmlns:converters="using:CalculatorApp.Converters"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:local="using:CalculatorApp"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:triggers="using:CalculatorApp.Views.StateTriggers"
+             d:DesignHeight="300"
+             d:DesignWidth="400"
+             XYFocusKeyboardNavigation="Enabled"
+             mc:Ignorable="d">
+    <UserControl.Resources>
+        <converters:BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter"/>
+        <converters:BooleanToVisibilityNegationConverter x:Key="BooleanToVisibilityNegationConverter"/>
+    </UserControl.Resources>
+
+    <Grid x:Name="GraphingOperators" UseLayoutRounding="False">
+        <Grid.RowDefinitions>
+            <RowDefinition x:Name="OperatorPanelRow" Height="Auto"/>
+            <RowDefinition Height="1*"/>
+            <RowDefinition Height="1*"/>
+            <RowDefinition Height="1*"/>
+            <RowDefinition Height="1*"/>
+            <RowDefinition Height="1*"/>
+            <RowDefinition Height="1*"/>
+            <RowDefinition Height="1*"/>
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="1*"/>
+            <ColumnDefinition Width="1*"/>
+            <ColumnDefinition Width="1*"/>
+            <ColumnDefinition Width="1*"/>
+            <ColumnDefinition Width="1*"/>
+        </Grid.ColumnDefinitions>
+
+        <VisualStateManager.VisualStateGroups>
+            <VisualStateGroup x:Name="SizeLayouts">
+                <VisualState x:Name="Large">
+                    <VisualState.StateTriggers>
+                        <triggers:ControlSizeTrigger MinWidth="878"
+                                                     MinHeight="851"
+                                                     Source="{x:Bind GraphingOperators}"/>
+                    </VisualState.StateTriggers>
+                    <VisualState.Setters>
+                        <Setter Target="NegateButton.FontSize" Value="{StaticResource CalcStandardOperatorCaptionSizeExtraLarge}"/>
+                        <Setter Target="PiButton.FontSize" Value="{ThemeResource CalcButtonCaptionSize}"/>
+                        <Setter Target="OpenParenthesisButton.FontSize" Value="{ThemeResource CalcButtonCaptionSize}"/>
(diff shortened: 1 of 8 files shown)


src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -0,0 +1,1194 @@
...
+                                This is because the content from within the Flyout do not inherit the IsEnabled property of the flyout parent,
+                                causing the shortcut keys to be used when the control would normally be disabled.
+                            -->
+                            <ToggleButton x:Name="TrigShiftButton"
+                                          x:Uid="trigShiftButton"
+                                          Style="{StaticResource CaptionToggleEmphasizedButtonStyle}"
+                                          FontFamily="{StaticResource CalculatorFontFamily}"
+                                          AutomationProperties.AutomationId="trigShiftButton"
+                                          Checked="TrigFlyoutShift_Toggle"
+                                          Content="&#xF897;"
```
EriWong: >F897 [](start = 54, length = 4)

Looking at this, I don't think this glyph is indicative of inverse trig functions.

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -0,0 +1,1194 @@
...
+                                                           Click="FlyoutButton_Clicked"
+                                                           Content="coth⁻¹"
+                                                           IsEnabled="{x:Bind TrigButton.IsEnabled, Mode=OneWay}"/>
+                            </Grid>
+                        </Grid>
+                    </Flyout>
+                </controls:OperatorPanelButton.FlyoutMenu>
+            </controls:OperatorPanelButton>
+
+            <controls:OperatorPanelButton x:Name="InequalityButton"
```
EriWong: ><controls:OperatorPanelButton x:Name="InequalityButton" [](start = 12, length = 55)

I don't think the graphing engine supports inequalities right now.  We should probably remove this until it does.

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -0,0 +1,1194 @@
...
+                    <Flyout x:Name="FuncFlyout"
+                            Windows10version1809:AreOpenCloseAnimationsEnabled="False"
+                            FlyoutPresenterStyle="{ThemeResource OperatorPanelFlyoutStyle}"
+                            Placement="Bottom">
+
+                        <Grid x:Name="FuncGrid"
+                              MinWidth="194"
+                              MinHeight="48"
+                              XYFocusKeyboardNavigation="Enabled">
+                            <Grid.RowDefinitions>
```
EriWong: What should we do when this one is clicked?  Do we add double bars and set the cursor between the two?

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
EriWong: Ah.  I see.  We're doing the same thing physical calculators do.

EriWong:
It's probably worth adding tooltips to all of the buttons

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
joseartrivera: It does support inequalities! Try typing `x>6` into an equation

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
joseartrivera: That would work, if we want to be consistent with other buttons like this then abs(-6) also works

joseartrivera:
> It's probably worth adding tooltips to all of the buttons

I would want to do this for all keyboards in the calculator app (for consistency and since adding them all at once would be easier than using different resw keys per mode). Can we open this up as a standalone issue scoped to the all the keyboards in the entire app?

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
EriWong: actually, this is probably not a conversation we need to have here as it isn't super relevant to the code at hand.

EriWong:
That makes sense.  Let's do that.

---
In reply to: [563395734](https://github.com/microsoft/calculator/pull/855#issuecomment-563395734) [](ancestors = 563395734)

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -0,0 +1,1194 @@
+<UserControl x:Class="CalculatorApp.GraphingNumPad"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:Windows10version1803="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractNotPresent(Windows.Foundation.UniversalApiContract, 7)"
+             xmlns:Windows10version1809="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 7)"
```
rudyhuyn: We should use the standard naming convention for these namespaces:
`contract7NotPresent`, `contract7Present`

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -0,0 +1,1194 @@
...
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:local="using:CalculatorApp"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:triggers="using:CalculatorApp.Views.StateTriggers"
+             d:DesignHeight="300"
+             d:DesignWidth="400"
+             XYFocusKeyboardNavigation="Enabled"
+             mc:Ignorable="d">
+    <UserControl.Resources>
+        <converters:BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter"/>
```
rudyhuyn: not used, can be removed!

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -0,0 +1,1194 @@
...
+
+                        <Setter Target="LessThanFlyoutButton.FontSize" Value="{ThemeResource CalcOperatorTextIconCaptionSize}"/>
+                        <Setter Target="LessThanOrEqualFlyoutButton.FontSize" Value="{ThemeResource CalcOperatorTextIconCaptionSize}"/>
+                        <Setter Target="EqualsFlyoutButton.FontSize" Value="{ThemeResource CalcOperatorTextIconCaptionSize}"/>
+                        <Setter Target="GreaterThanOrEqualFlyoutButton.FontSize" Value="{ThemeResource CalcOperatorTextIconCaptionSize}"/>
+                        <Setter Target="GreaterThanFlyoutButton.FontSize" Value="{ThemeResource CalcOperatorTextIconCaptionSize}"/>
+
+                        <Setter Target="OperatorPanelRow.MinHeight" Value="{ThemeResource OperatorPanelButtonRowSizeSmall}"/>
+
+                        <Setter Target="TrigButton.ChevronFontSize" Value="{ThemeResource OperatorPanelChevronFontSizeSmall}"/>
```
rudyhuyn: Should we use Style instead of setting individual properties? (for TrigButton, InequalityButton and FuncButton)

PR #856, opened 2019-12-09
Fix graph rendering of inequalities

joseartrivera:
- Use command plotIneq2D instead of plotEq2d when the equation contains an inequality

- Manual test

### DIFF
--- a/src/GraphControl/Control/Equation.cpp
+++ b/src/GraphControl/Control/Equation.cpp
@@ -404,6 +404,8 @@ namespace GraphControl
     wstring Equation::GetRequestHeader()
     {
         wstring expr{ Expression->Data() };
+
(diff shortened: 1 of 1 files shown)


src/GraphControl/Control/Equation.cpp
```diff
@@ -404,7 +404,12 @@ namespace GraphControl
     wstring Equation::GetRequestHeader()
     {
         wstring expr{ Expression->Data() };
-        if (expr.find(L">=<") != wstring::npos)
+        if (expr.find(L">&#x3E;<") != wstring::npos || expr.find(L">&#x3C;<") != wstring::npos || expr.find(L">&#x2265;<") != wstring::npos
+            || expr.find(L">&#x2264;<") != wstring::npos)
```
EriWong: can we add a comment on what each of these are?

PR #866, opened 2019-12-13
Save and restore the value of EquationInputArea when users scroll

rudyhuyn:
The current implementation never saved/restored the value of EquationInputAreas when users scrolled the list and DataTemplate where recycled by the list virtualization

- Refactor MathRichEditBox and add a (real) dependency property (the existing one were used a regular property) we can use for bindings.
- Move submission detection from EquationInputArea to MathRichEditBox

Also: 
- Because the submission detection is faster, other items don't get the focus when users press many times Enter.

### How to test?

- Enter 4-5 equations
- Type `Enter` many times to add more than 16 'empty' equations
**Before**: Because of the virtualization, new equations will be initialized with some of the first equations you typed (and vice-versa, the first equations will disappear if you scroll up)
**After**: The equations won't be mixed up

- manually

### DIFF
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -2101,6 +2101,7 @@
                                                                   FontWeight="{TemplateBinding FontWeight}"
                                                                   AcceptsReturn="false"
                                                                   InputScope="Text"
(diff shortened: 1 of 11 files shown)


rudyhuyn:
> Focus cannot be moved while getting or losing focus

I can't repro but I will try to catch this issue

PR #875, opened 2019-12-17
Cleaning up some UnitConverter code and making some of it more efficient

### LINKED ISSUE #874: UnitConverter has many instances where code can be simiplified ore made more efficient
- Some raw loops could be made to be range-for loops.
- wstringstream is often used for building a string which has lots of overhead compared to appending to a wstring.
- Some functions take a `wstring` or `const wchar_t*` where a `wstring_view` would be more appropriate.
  - `wstring_view` does not require constructing a string if a `const wchar_t*` is passed in.
  - `wstring_view` is constexpr constructable so it does not have to calculate its size at runtime when using string literals.

gsfreema:
- Using range-for loops where appropriate
- Stop using wstringstream where a wstring could be used instead
- Take advantage of wstring_view where appropriate

- Manual testing
- Ran existing unit tests

### DIFF
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -109,11 +109,10 @@ CategorySelectionInitializer UnitConverter::SetCurrentCategory(const Category& i
     {
         if (m_currentCategory.id != input.id)
         {
-            vector<Unit>& unitVector = m_categoryToUnits[m_currentCategory];
-            for (unsigned int i = 0; i < unitVector.size(); i++)
+            for (auto& unit : m_categoryToUnits[m_currentCategory])
             {
-                unitVector[i].isConversionSource = (unitVector[i].id == m_fromType.id);
-                unitVector[i].isConversionTarget = (unitVector[i].id == m_toType.id);
+                unit.isConversionSource = (unit.id == m_fromType.id);
+                unit.isConversionTarget = (unit.id == m_toType.id);
             }
             m_currentCategory = input;
             if (!m_currentCategory.supportsNegative && m_currentDisplay.front() == L'-')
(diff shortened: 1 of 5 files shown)


src/CalcManager/UnitConverter.cpp
```diff
@@ -192,41 +191,51 @@ void UnitConverter::SwitchActive(const wstring& newValue)
...
     size_t delimiterIndex = w.find(delimiter);
     size_t startIndex = 0;
-    vector<wstring> serializedTokens = vector<wstring>();
-    while (delimiterIndex != wstring::npos)
+    vector<wstring> serializedTokens;
+    while (delimiterIndex != wstring_view::npos)
     {
-        serializedTokens.push_back(w.substr(startIndex, delimiterIndex - startIndex));
-        startIndex = delimiterIndex + (int)wcslen(delimiter);
+        serializedTokens.emplace_back(w.substr(startIndex, delimiterIndex - startIndex));
```
gsfreema: `substr` on a `wstring_view` is very cheap and returns another `wstring_view`.  This means that a `wstring` will need to be constructed for `serializedTokens` so `emplace_back` is now appropriate so the `wstring` will be constructed directly in the container.

src/CalcManager/UnitConverter.cpp
```diff
@@ -192,41 +191,51 @@ void UnitConverter::SwitchActive(const wstring& newValue)
...
 }
 
-wstring UnitConverter::CategoryToString(const Category& c, const wchar_t* delimiter)
+wstring UnitConverter::CategoryToString(const Category& c, wstring_view delimiter)
 {
-    wstringstream out(wstringstream::out);
-    out << Quote(std::to_wstring(c.id)) << delimiter << Quote(std::to_wstring(c.supportsNegative)) << delimiter << Quote(c.name) << delimiter;
-    return out.str();
+    return Quote(std::to_wstring(c.id))
+        .append(delimiter)
```
gsfreema: I would have preferred to use `operator+` here, but there is not an overload between `wstring` and `wstring_view`.  However, `append` will keep growing the initial string returned by the first call to `Quote` as appropriate.  Since the size of a `wstring_view` is already known, the call to append will be more efficient because the `wstring` knows the resulting string size and knows how many characters to copy upfront.

src/CalcManager/UnitConverter.cpp
```diff
@@ -555,7 +562,7 @@ future<pair<bool, wstring>> UnitConverter::RefreshCurrencyRatios()
     return async([this, currencyDataLoader, sharedLoadResult]() {
         sharedLoadResult.wait();
         bool didLoad = sharedLoadResult.get();
-        wstring timestamp = L"";
+        wstring timestamp;
```
gsfreema: No need to use the constructor taking a `const wchar_t*` when a default `wstring` is already an empty string.

src/CalcManager/UnitConverter.cpp
```diff
@@ -649,7 +656,7 @@ vector<tuple<wstring, Unit>> UnitConverter::CalculateSuggested()
         if (stod(roundedString) != 0.0 || m_currentCategory.supportsNegative)
         {
             TrimTrailingZeros(roundedString);
-            returnVector.push_back(make_tuple(roundedString, entry.type));
+            returnVector.emplace_back(roundedString, entry.type);
```
gsfreema: Using `emplace_back` here instead of `push_back` so that the tuple can be directly constructed in the container vs. constructed and moved into the container.

src/CalcManager/UnitConverter.cpp
```diff
@@ -788,7 +795,7 @@ void UnitConverter::InitializeSelectedUnits()
         return;
     }
 
-    vector<Unit> curUnits = itr->second;
+    const vector<Unit>& curUnits = itr->second;
```
gsfreema: Avoiding making a copy of this vector.

src/CalcManager/UnitConverter.h
```diff
@@ -50,10 +50,6 @@ namespace UnitConversionManager
             accessibleName = nameValue1 + L" " + nameValue2;
         }
 
-        virtual ~Unit()
```
gsfreema: It doesn't appear that anything is supposed to inherit from `Unit` so there is no need to have a virtual destructor which adds overhead to this type by giving it a vtable.

src/CalcManager/UnitConverter.h
```diff
@@ -145,10 +141,6 @@ namespace UnitConversionManager
         {
         }
 
-        virtual ~ConversionData()
```
gsfreema: It doesn't appear that anything is supposed to inherit from `ConversionData` so there is no need to have a virtual destructor which adds overhead to this type by giving it a vtable.

src/CalcManager/UnitConverter.cpp
rudyhuyn: 👍

src/CalculatorUnitTests/UnitConverterViewModelUnitTests.cpp
```diff
@@ -227,7 +227,7 @@ void UnitConverterMock::SwitchActive(const std::wstring& newValue)
         return L"TEST";
     };
 
-void UnitConverterMock::RestoreUserPreferences(_In_ const std::wstring& /*userPreferences*/){};
+void UnitConverterMock::RestoreUserPreferences(_In_ std::wstring_view /*userPreferences*/){};
```
rudyhuyn: NIT: std and _In_ can be removed

PR #878, opened 2019-12-18
High Contrast support for Graphing Calculator

rudyhuyn:
- Add high contrast support to GraphControl
- Add high contrast support to EquationInputTextArea

Also:
- Rewrite button styles used in EquationInputTextArea to only modify their brushes instead of fully modify their templates (easier to maintain + less verbose)
- Make EquationInputTextArea buttons more fluent.
- localize the string "Enter an equation"
- slighly modify the error state of EquationInputTextArea (display all 4 borders instead of only 3)
- move all resources only used by Graphing Calculator from App.xaml to GraphingCalculator.xaml

Note: we want to keep the graph white background, black lines with dark theme for the moment.

- Manually

### DIFF
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -3,7 +3,6 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:Controls="using:CalculatorApp.Controls"
              xmlns:common="using:CalculatorApp.Common"
-             xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)"
              xmlns:local="using:CalculatorApp">
 
     <Application.Resources>
@@ -15,8 +14,6 @@
                 <ResourceDictionary x:Key="Default">
                     <Thickness x:Key="HighContrastThicknessTop">0,0,0,0</Thickness>
                     <x:Double x:Key="HighContrastStrokeThickness">0</x:Double>
-                    <x:Double x:Key="EquationButtonOverlayPointerOverOpacity">0.3</x:Double>
-                    <x:Double x:Key="EquationButtonOverlayPressedOpacity">0.5</x:Double>
                     <Color x:Key="AltHighColor">#FF000000</Color>
                     <Color x:Key="ChromeMediumLowColor">#FF2B2B2B</Color>
                     <Color x:Key="OperatorPanelScrollButtonBackgroundColor">#FF858585</Color>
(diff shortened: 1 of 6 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -353,39 +353,42 @@
...
                         </ResourceDictionary>
                         <ResourceDictionary x:Key="Dark">
                             <CornerRadius x:Key="TopButtonCornerRadius">4,4,0,0</CornerRadius>
                             <CornerRadius x:Key="BottomButtonCornerRadius">0,0,4,4</CornerRadius>
                             <CornerRadius x:Key="LeftButtonCornerRadius">4,0,0,4</CornerRadius>
                             <CornerRadius x:Key="RightButtonCornerRadius">0,4,4,0</CornerRadius>
+                            <Color x:Key="GrapherAxesColor">White</Color>
+                            <Color x:Key="GrapherBackground">Black</Color>
                         </ResourceDictionary>
                         <ResourceDictionary x:Key="HighContrast">
```
EriWong: >HighContrast [](start = 51, length = 12)

I'm not seeing things in the graph change when we go directly from high contrast black to high contrast white.  I'm not sure if this is our bug or a UWP bug though.

src/Calculator/App.xaml
```diff
@@ -224,21 +191,12 @@
...
-                    <SolidColorBrush x:Key="EquationButtonOverlayBackgroundBrush" Color="Transparent"/>
-                    <SolidColorBrush x:Key="EquationBoxErrorBackgroundBrush" Color="{ThemeResource SystemColorButtonFaceColor}"/>
-                    <SolidColorBrush x:Key="EquationBoxErrorBorderBrush" Color="{ThemeResource SystemColorButtonTextColor}"/>
-                    <SolidColorBrush x:Key="EquationButtonHideLineForegroundBrush" Color="{StaticResource SystemColorGrayTextColor}"/>
                     <!-- TODO: Figure out what colors we can use in high contrast -->
                     <SolidColorBrush x:Key="EquationBrush1" Color="{ThemeResource SystemColorGrayTextColor}"/>
                     <SolidColorBrush x:Key="EquationBrush2" Color="{ThemeResource SystemColorHighlightColor}"/>
                     <SolidColorBrush x:Key="EquationBrush3" Color="{ThemeResource SystemColorHotlightColor}"/>
-                    <SolidColorBrush x:Key="EquationBrush4" Color="{ThemeResource SystemColorWindowColor}"/>
+                    <SolidColorBrush x:Key="EquationBrush4" Color="{ThemeResource SystemColorWindowTextColor}"/>
```
joseartrivera: Not sure if related to this PR, but when sharing while in high contrast, the legend in the share output does not match up with the graph line colors.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -894,3 +858,27 @@ String ^ Grapher::ConvertToLinear(String ^ mmlString)
 
     return ref new String(linearExpression.c_str());
 }
+
+void Grapher::OnAxesColorPropertyChanged(Windows::UI::Color /*oldValue*/, Windows::UI::Color newValue)
+{
+    if (m_graph)
+    {
+        auto axesColor = Graphing::Color(newValue.R, newValue.G, newValue.B, newValue.A);
+        m_graph->GetOptions().SetAxisColor(axesColor);
```
joseartrivera: I am surprised these don't get reset when we reinitialize m_graph, normally we would want to add calls to these in UpdateGraphOptions() also.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
joseartrivera: Could be related to an issue I experienced before where the high contrast colors seem to be equal to each other when switching between the modes. I had a work around where I set them to transparent inbetween, not sure if it would help here.

src/Calculator/App.xaml
rudyhuyn: Not directly related to this PR, but something we should fix.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml
```diff
@@ -218,6 +219,502 @@
...
+                                            </controls:MathRichEditBox.ContextFlyout>
+                                        </controls:MathRichEditBox>
+
+                                        <Button x:Name="FunctionButton"
+                                                x:Uid="functionAnalysisButton"
+                                                Grid.Column="1"
+                                                MinWidth="34"
+                                                Margin="1,2"
+                                                VerticalAlignment="Stretch"
+                                                Style="{ThemeResource ButtonRevealStyle}"
```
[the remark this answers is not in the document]
rudyhuyn: The border is only visible when you "pointer over" the buttons and are important to indicate that an action is possible (this is the default border style used for actions). But I agree, it doesn't contrast well with some colors (yellow for example),I will decrease the opacity of the border

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
rudyhuyn: The issue was because we used `<StaticResource x:Key... />` so the color wasn't updated when we switch from a High Contrast theme to another High Contrast theme. Unfortunately `<ThemeResource x:Key` don't work but I found a workaround! Instead of setting the color in the ResourceDictionary, I create a new Style, the one being able to host a `{ThemeResource}`.

@joseartrivera This issue was different, it was because when you switch from a High Contrast mode to another one, the system only updates the Color, not the Brush, meaning GraphingControl which used the brush wasn't able to detect that the color change. This doesn't happen with other controls because they use the color of the brush we pass, not the case of GraphControl which create a copy of the color.

PR #879, opened 2019-12-18
Add Graph Settings

rudyhuyn:
Taking over @sanderl's PR.

- Add Graph Settings
- Add GraphSettingsViewModel to simplify the UI part of the control and manage all the logic
- Support error states (when minimum is higher than maximum)
- Disable keyboard when the flyout is opened
- Only load GraphSettings after the user clicks on the button to improve launch performance

- Manually


src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,168 @@
...
+                    <RowDefinition Height="10"/>
+                    <RowDefinition Height="auto"/>
+                </Grid.RowDefinitions>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="2*"/>
+                    <ColumnDefinition Width="10"/>
+                    <ColumnDefinition Width="2*"/>
+                </Grid.ColumnDefinitions>
+
+                <TextBox x:Name="SettingsXMin"
```
joseartrivera: Really long input causes the UI to stretch and a scrollbar to appear, I think we need to give these TextBoxes a reasonable max width.

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,168 @@
...
+                         x:Uid="GraphSettingsXMin"
+                         Style="{x:Bind SelectTextBoxStyle(ViewModel.XError), Mode=OneWay}"
+                         PreviewKeyDown="GridSettingsTextBox_PreviewKeyDown"
+                         Text="{x:Bind ViewModel.XMin, Mode=TwoWay}"/>
+                <TextBox x:Name="SettingsXMax"
```
joseartrivera: Pressing ctrl+c to try and copy a value within any of the text boxes causes a crash

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,168 @@
...
+            </Grid>
+
+            <TextBlock x:Name="UnitsHeading"
+                       x:Uid="UnitsHeading"
+                       Margin="0,16,0,6"
+                       Style="{StaticResource SubTitleTextBoxStyle}"
+                       AutomationProperties.HeadingLevel="Level2"/>
+
+            <StackPanel Orientation="Horizontal">
+                <RadioButton x:Name="Radians"
```
joseartrivera: Changing from Radians/Degrees/Gradians while the graph is empty appears to have no effect and resets the graph to default settings. This causes the values in the y and x min/max to become out of sync with the actual graph values as they also get reset but the textbox has the old values.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -493,3 +496,20 @@ void CalculatorApp::GraphingCalculator::ActiveTracing_KeyUp(Windows::UI::Core::C
         args->Handled = true;
     }
 }
+
+void GraphingCalculator::GraphSettingsButton_Click(Object ^ sender, RoutedEventArgs ^ e)
+{
+    if (m_flyoutGraphSettings == nullptr)
```
joseartrivera: Clicking away from the flyout settings to submit your values never shows possible errors to the user, so it may look like nothing occurred. Would it make sense to have some visual indicator on the icon itself? Like an error badge or something? Maybe not?

src/GraphControl/Control/Grapher.h
```diff
@@ -155,6 +156,105 @@ public
...
+        property double YAxisMax
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().second;
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(YAxisMin, value);
+                m_graph->GetOptions().SetDefaultYRange(newValue);
```
joseartrivera: Graph settings all get reset when a new equation is inputted, we need to add a call for all of these properties within the UpdateGraphOptions method to fix this.

src/GraphControl/Control/Grapher.h
```diff
@@ -155,6 +156,105 @@ public
...
+            }
+        }
+
+        property double YAxisMin
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().first;
+            }
+            void set(double value)
```
joseartrivera: Is there a reason all of these properties do the business logic in the setter rather than in a callback method like the other properties?

src/GraphControl/Control/Grapher.h
```diff
@@ -155,6 +156,105 @@ public
...
+            {
+                std::pair<double, double> newValue(value, YAxisMax);
+                m_graph->GetOptions().SetDefaultYRange(newValue);
```
joseartrivera: We should check if m_graph or m_renderMain is null before doing these operations so that the rest of the community can manipulate these without crashing the app. Here and in the rest of the properties.

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
joseartrivera: On hover in High Contrast White the foreground of the text is black and background is also black.

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,168 @@
...
+                <Setter Property="Template">
+                    <Setter.Value>
+                        <ControlTemplate TargetType="RadioButton">
+                            <Grid Name="LayoutRoot" Background="{ThemeResource ToggleButtonBackground}">
+                                <VisualStateManager.VisualStateGroups>
+                                    <VisualStateGroup x:Name="CommonStates">
+                                        <VisualState x:Name="Normal"/>
+                                        <VisualState x:Name="PointerOver">
+                                            <VisualState.Setters>
+                                                <Setter Target="LayoutRoot.Background" Value="{ThemeResource ToggleButtonBackgroundPointerOver}"/>
```
joseartrivera: In high contrast ToggleButtonBackgroundPointerOver is the same as the regular background, is that just the default style?

src/Calculator/Calculator.vcxproj
```diff
@@ -16,7 +16,7 @@
     <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
     <AppxDefaultResourceQualifierUAP_Contrast>black</AppxDefaultResourceQualifierUAP_Contrast>
     <AppxBundle>Always</AppxBundle>
-    <PackageCertificateKeyFile>TemporaryKey.pfx</PackageCertificateKeyFile>
+    <!--<PackageCertificateKeyFile>TemporaryKey.pfx</PackageCertificateKeyFile>-->
```
joseartrivera: I think we don't want to commit this

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -493,3 +496,20 @@ void CalculatorApp::GraphingCalculator::ActiveTracing_KeyUp(Windows::UI::Core::C
...
+        auto graphSettings = ref new GraphingSettings();
+        graphSettings->SetGrapher(this->GraphingControl);
+        m_flyoutGraphSettings = ref new Flyout();
+        m_flyoutGraphSettings->Content = graphSettings;
+        auto that(this);
+        m_flyoutGraphSettings->Closed += ref new Windows::Foundation::EventHandler<Platform::Object ^>(
+            [that](Platform::Object ^, Platform::Object ^) { that->GraphingNumPad->IsEnabled = true; });
+    }
+
+    GraphingNumPad->IsEnabled = false;
```
joseartrivera: Why is this needed? Seems weird that this is the only flyout that does this

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: Why doesn't this initialization happen somewhere more central? Would protect against other methods of editing the graph settings in the future such as from a hot key, etc.

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -0,0 +1,168 @@
...
+                    <ColumnDefinition Width="2*"/>
+                    <ColumnDefinition Width="10"/>
+                    <ColumnDefinition Width="2*"/>
+                </Grid.ColumnDefinitions>
+
+                <TextBox x:Name="SettingsXMin"
+                         x:Uid="GraphSettingsXMin"
+                         Style="{x:Bind SelectTextBoxStyle(ViewModel.XError), Mode=OneWay}"
+                         PreviewKeyDown="GridSettingsTextBox_PreviewKeyDown"
+                         Text="{x:Bind ViewModel.XMin, Mode=TwoWay}"/>
```
joseartrivera: I think this needs to be two way, or have some formatting logic since the texbox accepts input like '5-44' but I am not really sure what the double value is and it never becomes updated in the UI

src/GraphControl/Control/Grapher.h
```diff
@@ -155,6 +156,105 @@ public
...
+
+        property double XAxisMin
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().first;
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(value, XAxisMax);
```
joseartrivera: Not directly related to this PR but it makes reproing a crash easier when the axis is set to very large/small numbers

Try setting any of the y/x max and mins to very large or very low numbers, then hover your mouse over a graphed line near those large numbers.

in RenderMain::Render() the call to renderer->GetClosePointData crashes with 'An invalid parameter was passed to a function that considers invalid parameters fatal.'

I think it crashes because the range is a double but the API takes a float for the xValueOut and yValueOut params. We may want to file a bug, but also may want to consider handling the ranges as doubles and limiting their size if we can't get a bug fix.

joseartrivera: Actually I don't repro this outside of this branch, so maybe something in this branch caused the crash?

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
rudyhuyn: Workaround added.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: I've modified the behavior.

If one of the textbox in in an error state, I prevent the popup to be closed, asking the user to fix the values

src/Calculator/Calculator.vcxproj
rudyhuyn: 👍

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: We can also use `KeyboardShortcutManager::HonorShortcuts` too but I prefer this way:
- We should probably start using KeyboardAccelerator instead of this custom implementation
-  `KeyboardShortcutManager::HonorShortcuts` prevents us to use shortcuts in the popup

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
rudyhuyn: I don't understand, can you precise?

src/GraphControl/Control/Grapher.h
rudyhuyn: 👀

rudyhuyn: Yes, we should file a bug, 
`        HRESULT GetClosePointData(float inScreenPointX, float inScreenPointY, int& formulaIdOut, float& xScreenPointOut, float& yScreenPointOut, float& xValueOut, float& yValueOut) override`

It should use double, not float (all other fields use double (xmin, xmax, etc...)

rudyhuyn: In the meantime, we should ignore this exception

rudyhuyn: Unfortunately, we can't try... catch this exception, it happens on another thread. I'm able to repro the issue with the zoom out function, you need to have an equation and zoom out up to 1e17

src/GraphControl/Control/Grapher.h
rudyhuyn: because these properties are normal, not dependency properties. The main reason is because they don't "host" the data, they are are a bridge to access to the data

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: Why do we need to block shorcuts from the flyout menu? Does it cause it to close or something?

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
joseartrivera: Try typing '5-44' into one of the textboxes. It is accepted and converted to some double behind the scenes, but the resulting double is never displayed in the UI.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: to prevent the keys to blink when you press numbers

joseartrivera: @EriWong removed the blinking already and that should be in master, so this shouldn't be necessary anymore :)

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
rudyhuyn: Fixed! std::stod will stop at the first character it doesn't know how to manage. std::stod("5-44") will stop the parsing after the '5' character for example.

I've updated to code to force stod to fully parse the string

rudyhuyn:
Rebased!

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: These buttons don't blink in master, no need to add this code.

rudyhuyn: In this case you are right, this code is no more necessary! Removed!

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -0,0 +1,291 @@
...
+                {
+                    return;
+                }
+                m_XMax = value;
+                if (m_Graph != nullptr)
+                {
+                    try
+                    {
+                        size_t sz;
+                        auto number = std::stod(value->Data(), &sz);
```
joseartrivera: If I zoom out far enough, and click the graph settings, the numbers are red and I can never close the flyout

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -0,0 +1,291 @@
...
+        {
+            bool get()
+            {
+                return m_Graph != nullptr && m_Graph->TrigUnitMode == (int)Graphing::EvalTrigUnitMode::Radians;
+            }
+            void set(bool value)
+            {
+                if (value && m_Graph != nullptr && m_Graph->TrigUnitMode != (int)Graphing::EvalTrigUnitMode::Radians)
+                {
+                    m_Graph->TrigUnitMode = (int)Graphing::EvalTrigUnitMode::Radians;
```
joseartrivera: Changing the trig mode appears to change the ranges on the Graph but they are not immediately updated in the flyout

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -0,0 +1,291 @@
...
+                m_XMin = value;
+                if (m_Graph != nullptr)
+                {
+                    try
+                    {
+                        size_t sz;
+                        auto number = std::stod(value->Data(), &sz);
+                        if (value->Length() == sz)
+                        {
+                            m_Graph->XAxisMin = m_XMinValue = number;
```
joseartrivera: I think this is adding trailing zeroes, setting y-max to 75 turns it into 75.000000 in the textbox, maybe consider running through TrimTrailingZeroes?

rudyhuyn: Done

src/GraphControl/Control/Grapher.h
```diff
@@ -100,6 +102,130 @@ public
...
+                    m_renderMain->RunRenderPass();
+                }
+            }
+        }
+
+        void GetDisplayRanges(double* xMin, double* xMax, double* yMin, double* yMax)
+        {
+            try
+            {
+                m_graph->GetRenderer()->GetDisplayRanges(*xMin, *xMax, *yMin, *yMax);
```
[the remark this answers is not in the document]
rudyhuyn: the try/catch wasn't for `m_graph` but for GetDisplayRanges crashing when the value returned are out of range

src/GraphControl/Control/Grapher.h
```diff
@@ -100,6 +102,130 @@ public
...
+            }
+            catch (const std::exception&)
+            {
+                OutputDebugString(L"GetDisplayRanges failed\r\n");
+            }
+        }
+
+        void SetDisplayRanges(double xMin, double xMax, double yMin, double yMax)
```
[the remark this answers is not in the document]
rudyhuyn: It happens when you zoom out and set a value lower than the current value displayed, the value passed to `m_renderer->SetDisplayRanges` are the correct ones, We should report this bug to team working on the renderer, but it should not block this PR.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -505,3 +508,37 @@ void CalculatorApp::GraphingCalculator::ActiveTracing_KeyUp(Windows::UI::Core::C
...
+
+    m_flyoutGraphSettings->ShowAt(GraphSettingsButton);
+}
+
+void GraphingCalculator::OnSettingsFlyout_Closing(
+    FlyoutBase ^ sender,
+    FlyoutBaseClosingEventArgs ^ args)
+{
+    auto graphingSetting = static_cast<GraphingSettings ^>(static_cast<Flyout ^>(sender)->Content);
+    args->Cancel = graphingSetting->CanBeClose();
```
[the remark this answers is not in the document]
rudyhuyn: Good find! But this is not a bug specific to the Graph Calculator:

- Open Date Calculator
- Open the date selector
- Type Alt+1
==> The Date popup will remain

- Open one of the calculator
- Open the Memory/History popup
- Type Alt + 3
==> the memory popup will remain

- Open the Programmer Calculator
- Open the Bit Shift popup
- Type Alt + 1
==> the Bit Shift popup will remain

We need to modify how we handle the navigation between the features. I will cover these issues in another PR.

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
rudyhuyn: this should not happen, we call `m_renderer->SetDisplayRanges` with the same value after the user selects a different unit. Fixing SetDisplayRanges will also fix this issue.

src/GraphControl/Control/Grapher.h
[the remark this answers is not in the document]
rudyhuyn: This is due to the `Grapher.ForceProportionalAxes`, we cannot give the user the freedom to set values and enforce proportional axes. 

2 solutions:
- we don't enforce proportional axes
- we modify the value of the 3 other fields when we modify 1.

@grochocki, your opinion?

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
rudyhuyn: Can you send me screenshots? From what I see, it doesn't. The issue I see isn't the range but how the graph displays them:
Radian:
image

Degree:
image

Both display the same ranges but the way the graph shows them isn't deterministic

src/GraphControl/Control/Grapher.h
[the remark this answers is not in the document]
rudyhuyn: Oh! I was expecting the mock to mock this part too. The best solution would be to improve the mock and include a fake IGraph and IGraphingOption, so the community will be able to test graph settings, change values, etc..

src/GraphControl/Control/Grapher.h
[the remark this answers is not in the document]
rudyhuyn: Solution 2 does not seem possible with the current APIs, in some cases, Grapher won't keep the value you have just typed instead of modifying one of the 3 other values.

src/GraphControl/Control/Grapher.h
```diff
@@ -100,6 +102,130 @@ public
...
+        property double YAxisMax
+        {
+            double get()
+            {
+                return m_graph->GetOptions().GetDefaultXRange().second;
+            }
+            void set(double value)
+            {
+                std::pair<double, double> newValue(YAxisMin, value);
+                m_graph->GetOptions().SetDefaultYRange(newValue);
```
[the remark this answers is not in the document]
rudyhuyn: conditional added (but no more necessary, the mock implements GetOptions now

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -69,12 +71,12 @@ GraphingCalculator::GraphingCalculator()
 
     // OemMinus and OemAdd aren't declared in the VirtualKey enum, we can't add this accelerator XAML-side
     auto virtualKey = ref new KeyboardAccelerator();
-    virtualKey->Key = (VirtualKey)187; //OemMinus key
+    virtualKey->Key = (VirtualKey)187; // OemMinus key
```
joseartrivera: Not related to this PR, but ctrl + '-' and ctrl + '+' do the opposite of what clicking the + and - do.

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -0,0 +1,297 @@
...
+        {
+            bool get()
+            {
+                return m_Graph != nullptr && m_Graph->TrigUnitMode == (int)Graphing::EvalTrigUnitMode::Radians;
+            }
+            void set(bool value)
+            {
+                if (value && m_Graph != nullptr && m_Graph->TrigUnitMode != (int)Graphing::EvalTrigUnitMode::Radians)
+                {
+                    m_Graph->TrigUnitMode = (int)Graphing::EvalTrigUnitMode::Radians;
```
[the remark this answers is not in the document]
rudyhuyn: This is "expected". The Graph control will automatically reset X and Y ranges when you select a different trigo unit, so we call `RefreshPosition()` in the view model after we set a different unit.

Another solution is to not call RefreshPosition and update the UI with the new XMin, Xmax, YMin and YMax values, but it will be very frustrating for users to set these values again.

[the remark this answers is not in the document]
joseartrivera: Could we set the correct x/y min/max values after the graph is initialized but before we render the graph to avoid the flickering?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: solved

PR #881, opened 2019-12-18
Using wstring instead of wstringstream where appropriate

### LINKED ISSUE #880: Added overhead of using wstringstream when wstring would work
Throughout the codebase, wstringstream is often used for building a result string.  This is useful if you're trying to do special formatting/localization, but most of the use cases are simply building strings by appending characters or other strings so no special formatting/localization is happening.  Unlike Platform::String, wstring is mutable and will grow its internal array as needed.  This would be optimal in many cases over using wstringstream.

gsfreema:
- Replaced most wstringstream usage with wstring.
- In a couple of places, used if...else instead of ternary operator since the later requires the true and false parts to be of the same type.  This resulted in extra wstring instances to be created which isn't necessary.
- In one pace, used Platform:String's constructor taking a character and a count.  This seems more clear and (should be more efficient) than the prior code of `L"" + c`.  That code seems confusing to a C++ developer since it is normally something a compiler would warn about.  I'm assuming it's a special C++/CX feature.

- Ran existing unit tests.
- Manual testing.

### DIFF
--- a/src/CalcManager/CEngine/CalcInput.cpp
+++ b/src/CalcManager/CEngine/CalcInput.cpp
@@ -269,34 +269,47 @@ bool CalcInput::IsEmpty()
 wstring CalcInput::ToString(uint32_t radix)
 {
     // In theory both the base and exponent could be C_NUM_MAX_DIGITS long.
-    wstringstream resStream;
-
     if ((m_base.value.size() > MAX_STRLEN) || (m_hasExponent && m_exponent.value.size() > MAX_STRLEN))
     {
         return wstring();
     }
 
+    wstring result;
+
     if (m_base.IsNegative())
     {
-        resStream << L'-';
+        result += L'-';
     }
 
-    resStream << (m_base.IsEmpty() ? L"0" : m_base.value);
+    if (m_base.IsEmpty())
+    {
+        result += L'0';
+    }
+    else
+    {
+        result += m_base.value;
+    }
 
     if (m_hasExponent)
     {
         // Add a decimal point if it is not already there
         if (!m_hasDecimal)
         {
-            resStream << m_decSymbol;
+            result += m_decSymbol;
         }
 
-        resStream << ((radix == 10) ? L'e' : L'^');
-        resStream << (m_exponent.IsNegative() ? L'-' : L'+');
-        resStream << (m_exponent.IsEmpty() ? L"0" : m_exponent.value);
-    }
+        result += ((radix == 10) ? L'e' : L'^');
+        result += (m_exponent.IsNegative() ? L'-' : L'+');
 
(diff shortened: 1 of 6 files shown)


src/CalcManager/CEngine/CalcInput.cpp
```diff
@@ -269,34 +269,47 @@ bool CalcInput::IsEmpty()
...
     {
         return wstring();
     }
 
+    wstring result;
+
     if (m_base.IsNegative())
     {
-        resStream << L'-';
+        result += L'-';
```
rudyhuyn: `=` is ~15% faster than `+=`, let's use it here instead (as well as line 286 and 290)

src/CalcManager/CEngine/scidisp.cpp
```diff
@@ -364,11 +364,10 @@ wstring CCalcEngine::GroupDigits(wstring_view delimiter, vector<uint32_t> const&
     // now copy the negative sign if it is there
     if (isNumNegative)
     {
-        groupedStream << displayString[0];
+        result += displayString[0];
     }
 
-    auto groupedString = groupedStream.str();
-    wstring result(groupedString.rbegin(), groupedString.rend());
+    std::reverse(result.begin(), result.end());
```
rudyhuyn: NIT: can remove `std:`

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -1204,78 +1204,73 @@ wstring NumberToString(_Inout_ PNUMBER& pnum, int format, uint32_t radix, int32_
     }
 
     // Begin building the result string
-    wstringstream resultStream{};
+    wstring result{};
```
rudyhuyn: very NIT: `{}` not necessary

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -1204,78 +1204,73 @@ wstring NumberToString(_Inout_ PNUMBER& pnum, int format, uint32_t radix, int32_
...
 
     // Make sure negative zeros aren't allowed.
     if ((pnum->sign == -1) && (length > 0))
     {
-        resultStream << L'-';
+        result += L'-';
```
rudyhuyn: NIT: we can use `=` here

src/CalcManager/CEngine/CalcInput.cpp
gsfreema: I think we can use `=` for this line, but I don't believe we can use it for lines 286 and 290 because that would overwrite the value that is in result which might contain a `-` if IsNegative is true.

rudyhuyn: right!

PR #882, opened 2019-12-18
Graphing Calculator - Add Shortcuts to zoom in/out 

rudyhuyn:
- Support <kbd>Ctrl</kbd> + <kbd>-/+</kbd> (in addition to the existing <kbd>Ctrl</kbd> + <kbd>Numpad -/+</kbd>)

⚠ Important change:
- We use the new RS3+ UWP api to manage shortcuts instead of our own implementation. The UWP being more powerful and supporting multi-shortcuts. We should plan to migrate all other shortcuts in the future
- The current shortcut was localized, not this implementation, but I don't think it's something we need.

Also: 
- simplify KeyboardShortcutManager
- fix a bug in KeyboardShortcutManager (s_VirtualKeyControlInverseChordsForButtons were used instead of s_VirtualKeyInverseChordsForButtons)

- Manually, English and French keyboard layout.

### DIFF
--- a/src/CalcViewModel/Common/KeyboardShortcutManager.cpp
+++ b/src/CalcViewModel/Common/KeyboardShortcutManager.cpp
@@ -44,10 +44,15 @@ static multimap<int, multimap<MyVirtualKey, WeakReference>> s_VirtualKeyControlS
 static multimap<int, multimap<MyVirtualKey, WeakReference>> s_VirtualKeyInverseChordsForButtons;
 static multimap<int, multimap<MyVirtualKey, WeakReference>> s_VirtualKeyControlInverseChordsForButtons;
 
-static multimap<int, bool> s_ShiftKeyPressed;
-static multimap<int, bool> s_ControlKeyPressed;
-static multimap<int, bool> s_ShiftButtonChecked;
-static multimap<int, bool> s_IsDropDownOpen;
+static map<int, bool> s_ShiftKeyPressed;
+static map<int, bool> s_ControlKeyPressed;
+static map<int, bool> s_ShiftButtonChecked;
+static map<int, bool> s_IsDropDownOpen;
+
+static map<int, bool> s_ignoreNextEscape;
+static map<int, bool> s_keepIgnoringEscape;
+static map<int, bool> s_fHonorShortcuts;
+static map<int, Flyout ^> s_AboutFlyout;
 
 static reader_writer_lock s_keyboardShortcutMapLock;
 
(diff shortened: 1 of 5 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -66,6 +66,17 @@ GraphingCalculator::GraphingCalculator()
 
     // And when the actual trace value changes
     GraphingControl->TracingValueChangedEvent += ref new TracingValueChangedEventHandler(this, &GraphingCalculator::OnTracePointChanged);
+
+    // OemMinus and OemAdd aren't declared in the VirtualKey enum, we can't add this accelerator XAML-side
+    auto virtualKey = ref new KeyboardAccelerator();
+    virtualKey->Key = (VirtualKey)187; //OemMinus key
+    virtualKey->Modifiers = VirtualKeyModifiers::Control;
```
joseartrivera: Why the need for Ctrl instead of just +/- directly? I think it might be more intuitive and discoverable.

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
```diff
@@ -474,8 +469,8 @@ const std::multimap<MyVirtualKey, WeakReference>& GetCurrentKeyDictionary(MyVirt
         }
         else
         {
-            auto iterViewMap = s_VirtualKeyControlInverseChordsForButtons.find(viewId);
-            if (iterViewMap != s_VirtualKeyControlInverseChordsForButtons.end())
+            auto iterViewMap = s_VirtualKeyInverseChordsForButtons.find(viewId);
```
joseartrivera: Good find! I wonder, does this change the behavior for shortcuts that we have documented?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: Specs :D

I suppose to have the same behavior than Maps.

cc @grochocki

src/CalcViewModel/Common/KeyboardShortcutManager.cpp
rudyhuyn: I tried all the shortcut using Ctrl+shift+Key, luckily, all of them had an equivalent Ctrl+Key

PR #884, opened 2019-12-19
Wstring view overrun

### LINKED ISSUE #883: Potential buffer overrun in CCalcEngine::DigitGroupingStringToGroupingVector
**Describe the bug**
CCalcEngine::DigitGroupingStringToGroupingVector takes in a wstring_view and then iterates over its characters until it reaches a null terminating character.  The problem is that wstring_view isn't guaranteed to have a null terminating character within the range that the wstring_view is supposed to represent.

https://github.com/microsoft/calculator/blob/a21b4a2d1a2e57e8c2807e3d780ef05a541013f5/src/CalcManager/CEngine/scidisp.cpp#L221

**Steps To Reproduce**
Here is an example of how you can go past the end of a string_view with this approach.
https://godbolt.org/z/c2W3Lp

**Requested Assignment**

gsfreema:
- Updated loop inside CCalcEngine::DigitGroupingStringToGroupingVector to compare against an end point set to the beginning plus size of the wstring_view.

- Manual testing.
- Added a unit test to confirm that the grouping is based on the wstring_view passed in.

### DIFF
--- a/src/CalcManager/CEngine/scidisp.cpp
+++ b/src/CalcManager/CEngine/scidisp.cpp
@@ -218,8 +218,9 @@ vector<uint32_t> CCalcEngine::DigitGroupingStringToGroupingVector(wstring_view g
     vector<uint32_t> grouping;
     uint32_t currentGroup = 0;
     wchar_t* next = nullptr;
-    const wchar_t* end = groupingString.data() + groupingString.size();
-    for (auto itr = groupingString.data(); itr != end; ++itr)
+    const wchar_t* begin = groupingString.data();
+    const wchar_t* end = begin + groupingString.length();
+    for (auto itr = begin; itr != end; ++itr)
     {
         // Try to parse a grouping number from the string
         currentGroup = wcstoul(itr, &next, 10);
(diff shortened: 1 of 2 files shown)


src/CalcManager/CEngine/scidisp.cpp
```diff
@@ -215,10 +215,11 @@ int CCalcEngine::IsNumberInvalid(const wstring& numberString, int iMaxExp, int i
 \****************************************************************************/
 vector<uint32_t> CCalcEngine::DigitGroupingStringToGroupingVector(wstring_view groupingString)
 {
-    vector<uint32_t> grouping{};
+    vector<uint32_t> grouping;
     uint32_t currentGroup = 0;
     wchar_t* next = nullptr;
-    for (const wchar_t* itr = groupingString.data(); *itr != L'\0'; ++itr)
+    const wchar_t* end = groupingString.data() + groupingString.size();
+    for (auto itr = groupingString.data(); itr != end; ++itr)
```
rudyhuyn: Should we use     
` for (auto it = groupingString.cbegin(); it != groupingString.cend(); ++it)` 

instead?

gsfreema: That was my initial thought but the way itr was being used inside the loop as a const wchar_t* made that challenging. Maybe the iterator could have been casted but it felt weird.

gsfreema: That was my initial thought but the way itr was being used inside the loop as a const wchar_t* made that challenging. Maybe the iterator could have been casted but it felt weird.

rudyhuyn: Got it! We can keep it like that, but I would suggest to only use size() or length() but not both in the same function (line 236) and call `groupingString.data()` only one time if possible

PR #885, opened 2019-12-19
Update equation colors

joseartrivera:
- Updates equation colors and order of assignment
- Fixes KGF header background/margin
- Change error icon to warning
- Other style clean up

- manual tests

### DIFF
--- a/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
+++ b/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
@@ -115,8 +115,8 @@
 
                         <Setter Target="OperatorPanelRow.MinHeight" Value="{StaticResource OperatorPanelButtonRowSizeLarge}"/>
 
-                        <Setter Target="BitwiseButton.Style" Value="{ThemeResource OperatorPanelButtonLargeStyle}"/>
-                        <Setter Target="BitShiftButton.Style" Value="{ThemeResource OperatorPanelButtonLargeStyle}"/>
+                        <Setter Target="BitwiseButton.Style" Value="{StaticResource OperatorPanelButtonLargeStyle}"/>
+                        <Setter Target="BitShiftButton.Style" Value="{StaticResource OperatorPanelButtonLargeStyle}"/>
 
                         <Setter Target="BitwiseGrid.MinWidth" Value="387"/>
                         <Setter Target="BitwiseGrid.MinHeight" Value="192"/>
@@ -167,8 +167,8 @@
 
                         <Setter Target="OperatorPanelRow.MinHeight" Value="{StaticResource OperatorPanelButtonRowSizeMedium}"/>
 
-                        <Setter Target="BitwiseButton.Style" Value="{ThemeResource OperatorPanelButtonMediumStyle}"/>
-                        <Setter Target="BitShiftButton.Style" Value="{ThemeResource OperatorPanelButtonMediumStyle}"/>
+                        <Setter Target="BitwiseButton.Style" Value="{StaticResource OperatorPanelButtonMediumStyle}"/>
+                        <Setter Target="BitShiftButton.Style" Value="{StaticResource OperatorPanelButtonMediumStyle}"/>
 
                         <Setter Target="BitwiseGrid.MinWidth" Value="416"/>
                         <Setter Target="BitwiseGrid.MinHeight" Value="144"/>
@@ -217,8 +217,8 @@
 
                         <Setter Target="OperatorPanelRow.MinHeight" Value="{StaticResource OperatorPanelButtonRowSizeSmall}"/>
 
-                        <Setter Target="BitwiseButton.Style" Value="{ThemeResource OperatorPanelButtonSmallStyle}"/>
-                        <Setter Target="BitShiftButton.Style" Value="{ThemeResource OperatorPanelButtonSmallStyle}"/>
+                        <Setter Target="BitwiseButton.Style" Value="{StaticResource OperatorPanelButtonSmallStyle}"/>
+                        <Setter Target="BitShiftButton.Style" Value="{StaticResource OperatorPanelButtonSmallStyle}"/>
 
                         <Setter Target="BitwiseGrid.MinWidth" Value="194"/>
                         <Setter Target="BitwiseGrid.MinHeight" Value="96"/>
--- a/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
+++ b/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
@@ -115,13 +115,8 @@
 
                         <Setter Target="OperatorPanelRow.MinHeight" Value="{StaticResource OperatorPanelButtonRowSizeLarge}"/>
 
-                        <Setter Target="BitwiseButton.ChevronFontSize" Value="{ThemeResource OperatorPanelChevronFontSizeLarge}"/>
-                        <Setter Target="BitwiseButton.GlyphFontSize" Value="{ThemeResource OperatorPanelGlyphFontSizeLarge}"/>
-                        <Setter Target="BitwiseButton.FontSize" Value="{ThemeResource OperatorPanelFontSizeLarge}"/>
-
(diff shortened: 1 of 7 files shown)


rudyhuyn:
Reviewing...

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -67,8 +69,16 @@ void EquationInputArea::AddNewEquation()
         Equations->GetAt(Equations->Size - 1)->IsLastItemInList = false;
     }
 
-    m_lastLineColorIndex = (m_lastLineColorIndex + 1) % AvailableColors->Size;
-    auto eq = ref new EquationViewModel(ref new Equation(), ++m_lastFunctionLabelIndex, AvailableColors->GetAt(m_lastLineColorIndex)->Color);
+    // Cap equations at number of available colors
+    if (Equations->Size >= m_AvailableColors->Size)
```
rudyhuyn: Should we?

What are we trying to solve?

If we don't want 2 equations to not use the same color, this won't prevent it (if you remove an equation and add a new one, it will probably reuse the same color than an existing one). IMO, we should not limit the number of equations because of this reason

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -67,8 +69,16 @@ void EquationInputArea::AddNewEquation()
...
+    {
+        return;
+    }
+
+    m_lastLineColorIndex = (m_lastLineColorIndex + 1) % colorAssignmentMapping.size();
+
+    auto eq =
+        ref new EquationViewModel(ref new Equation(), ++m_lastFunctionLabelIndex, AvailableColors->GetAt(colorAssignmentMapping[m_lastLineColorIndex])->Color);
```
rudyhuyn: I'm afraid it will launch a IndexOutOfRangeException in High Contrast because of the limited number of colors?

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
rudyhuyn: Especially in High Contrast (only 4 colors/equations)

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
rudyhuyn: I confirm, it launches an exception after you add the third equation

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
```diff
@@ -115,13 +115,8 @@
...
                         <Setter Target="OperatorPanelRow.MinHeight" Value="{StaticResource OperatorPanelButtonRowSizeLarge}"/>
 
-                        <Setter Target="BitwiseButton.ChevronFontSize" Value="{ThemeResource OperatorPanelChevronFontSizeLarge}"/>
-                        <Setter Target="BitwiseButton.GlyphFontSize" Value="{ThemeResource OperatorPanelGlyphFontSizeLarge}"/>
-                        <Setter Target="BitwiseButton.FontSize" Value="{ThemeResource OperatorPanelFontSizeLarge}"/>
-
-                        <Setter Target="BitShiftButton.ChevronFontSize" Value="{ThemeResource OperatorPanelChevronFontSizeLarge}"/>
-                        <Setter Target="BitShiftButton.GlyphFontSize" Value="{ThemeResource OperatorPanelGlyphFontSizeLarge}"/>
-                        <Setter Target="BitShiftButton.FontSize" Value="{ThemeResource OperatorPanelFontSizeLarge}"/>
+                        <Setter Target="BitwiseButton.Style" Value="{ThemeResource OperatorPanelButtonLargeStyle}"/>
```
rudyhuyn: You can use a StaticResource instead (and same for all other changes), the style as well as the resource used are not themed.

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
```diff
@@ -152,7 +152,7 @@
             <RowDefinition Height="Auto"/>
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
-        <Grid>
+        <Grid Margin="0,0,1,0" Background="{ThemeResource TextControlBackground}">
```
rudyhuyn: the background looks great! But I would prefer without the margin so the "header bar" goes up to the side of the window.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
joseartrivera: Will defer to @grochocki as this was a change he wanted. I don't think the goal is to have unique colors for each equation. @grochocki this is the change to limit the number of equations to 14.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
joseartrivera: Good catch!

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
joseartrivera: The margin is there to match the margin of the original EquationTextBox, without it looks like the textbox gets slightly bigger for no reason when transitioning between the two screens.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
joseartrivera: But it was not intended to limit to only 4 in high contrast, will fix that

src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
rudyhuyn: Let's use `1,0` in this case to match the textboxes

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -67,8 +69,27 @@ void EquationInputArea::AddNewEquation()
         Equations->GetAt(Equations->Size - 1)->IsLastItemInList = false;
     }
 
+    // Cap equations at 14
+    if (Equations->Size >= 14)
```
rudyhuyn: can we define a constant for this number?

src/Calculator/Views/GraphingCalculator/GraphingNumPad.xaml
```diff
@@ -112,17 +112,9 @@
...
-                        <Setter Target="TrigButton.GlyphFontSize" Value="{ThemeResource OperatorPanelGlyphFontSizeLarge}"/>
-                        <Setter Target="TrigButton.FontSize" Value="{ThemeResource OperatorPanelFontSizeLarge}"/>
-
-                        <Setter Target="InequalityButton.ChevronFontSize" Value="{ThemeResource OperatorPanelChevronFontSizeLarge}"/>
-                        <Setter Target="InequalityButton.GlyphFontSize" Value="{ThemeResource OperatorPanelGlyphFontSizeLarge}"/>
-                        <Setter Target="InequalityButton.FontSize" Value="{ThemeResource OperatorPanelFontSizeLarge}"/>
-
-                        <Setter Target="FuncButton.ChevronFontSize" Value="{ThemeResource OperatorPanelChevronFontSizeLarge}"/>
-                        <Setter Target="FuncButton.GlyphFontSize" Value="{ThemeResource OperatorPanelGlyphFontSizeLarge}"/>
-                        <Setter Target="FuncButton.FontSize" Value="{ThemeResource OperatorPanelFontSizeLarge}"/>
```
rudyhuyn: great, we had no reasons to use ThemeResource here!

rudyhuyn:
Great job!!! Thank you Pepe!

PR #886, opened 2019-12-20
Minor changes to share output

joseartrivera:
- Addresses UI issues identified in UX review for share output

- Manual tests

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
@@ -258,7 +258,7 @@ void GraphingCalculator::OnDataRequested(DataTransferManager ^ sender, DataReque
 
             rawHtml << L"<br><span style=\"color: rgb(68, 114, 196); font-style: bold; font-size: 13pt;\">";
             rawHtml << resourceLoader->GetString(L"VariablesShareHeader")->Data();
-            rawHtml << L"</span><br><div style=\"margin: 8pt 0pt 0pt 0pt;\"><span>";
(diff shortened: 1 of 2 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -256,9 +256,9 @@ void GraphingCalculator::OnDataRequested(DataTransferManager ^ sender, DataReque
         {
             auto localizedSeperator = LocalizationSettings::GetInstance().GetListSeparator() + L" ";
 
-            rawHtml << L"<span style=\"color: rgb(68, 114, 196); font-style: bold; font-size: 13pt;\">";
+            rawHtml << L"<br><span style=\"color: rgb(68, 114, 196); font-style: bold; font-size: 13pt;\">";
             rawHtml << resourceLoader->GetString(L"VariablesShareHeader")->Data();
-            rawHtml << L"</span><br><span>";
+            rawHtml << L"</span><br><div style=\"margin: 4pt 0pt 0pt 0pt;\"><span>";
```
EriWong: ><div style=\"margin: 4pt 0pt 0pt 0pt;\"> [](start = 36, length = 40)

does this do anything beyond putting a margin around the span?  We could just put this style tag in the span itself right?

joseartrivera: That may work normally, but the mail app seems to ignore a lot of css styling unless it is on specific elements for some reason, so I had to hack whatever I could so that the mail app also respects the margins and other styling.

joseartrivera: Removed the unnecessary span as discussed!

PR #890, opened 2019-12-23
Change active tracing to use correct glyph instead of dot

EriWong:
## Fixes part of #338.
Replaces the red dot used in active tracing with a red cursor icon

manually


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -424,6 +424,18 @@
...
                 </StackPanel>
             </Border>
+            <Border x:Name="TracePointer"
+                    Width="16"
+                    Height="18"
+                    Margin="50,50,0,0"
+                    HorizontalAlignment="Left"
+                    VerticalAlignment="Top"
+                    Visibility="Collapsed">
+                <FontIcon Foreground="Red"
```
EriWong: >Foreground="Red" [](start = 26, length = 16)

The dot was red before, so I made the cursor red...  I tried going with the theme text color instead, but that resulted in something that didn't pop.  It was hard to see where the pointer was at a glance

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -424,6 +424,18 @@
                     </Button>
                 </StackPanel>
             </Border>
+            <Border x:Name="TracePointer"
+                    Width="16"
+                    Height="18"
+                    Margin="50,50,0,0"
```
EriWong: >50,50 [](start = 28, length = 5)

For some reason, David set the initial pointer to be at 50, 50.  Following along here.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.h
```diff
@@ -55,6 +55,7 @@ public ref class GraphingCalculator sealed : public Windows::UI::Xaml::Data::INo
 
         void OnShowTracePopupChanged(bool newValue);
         void OnTracePointChanged(Windows::Foundation::Point newPoint);
+        void OnPointerPointChanged(Windows::Foundation::Point newPoint);
```
EriWong: >OnPointerPointChanged [](start = 13, length = 21)

This name is kind of confusing...  that said, I couldn't come up with a better name that wasn't confusing with TracePointChanged.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -177,6 +180,11 @@ void GraphingCalculator::OnTracePointChanged(Windows::Foundation::Point newPoint
     PositionGraphPopup();
 }
 
+void CalculatorApp::GraphingCalculator::OnPointerPointChanged(Windows::Foundation::Point newPoint)
+{
+    TracePointer->Margin = Thickness(newPoint.X, newPoint.Y, 0, 0);
```
EriWong: >TracePointer->Margin = Thickness(newPoint.X, newPoint.Y, 0, 0); [](start = 4, length = 63)

Moving the pointer by setting margins on it...  Not sure this is a good way to go about it, but didn't see any absolute positioning apis...

src/GraphControl/Control/Grapher.h
```diff
@@ -20,13 +20,16 @@ public
...
+public
+    delegate void PointerValueChangedEventHandler(Windows::Foundation::Point value);
 
     [Windows::UI::Xaml::Markup::ContentPropertyAttribute(Name = L"Equations")] public ref class Grapher sealed
         : public Windows::UI::Xaml::Controls::Control,
           public Windows::UI::Xaml::Data::INotifyPropertyChanged
     {
     public:
         event TracingValueChangedEventHandler ^ TracingValueChangedEvent;
+        event PointerValueChangedEventHandler ^ PointerValueChangedEvent;
```
EriWong: >PointerValueChangedEvent [](start = 48, length = 24)

passing the trace point back via event.  Does this have perf implications?  It seems ok on my box, but if there's a better way to do this, I'm all ears :)

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: Maybe using a [canvas](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.canvas) could help, but the margins seem to work fine.

PR #895, opened 2020-01-03
Update build year in Aboutflyout

### LINKED ISSUE #894: Change the build year in Aboutflyout
It's 2020 now, and according to #320 the `BUILD_YEAR` remains 2019.
Although none of the pending PR is merged until now, it's better to let the `BUILD_YEAR` be 2020 now so we don't need to force someone to change it in an unrelated PR.
*possibly we can have an automatic script that changes this macro to next year every Jan. 1st, but manually change it is also OK.*

**Requested Assignment**

dovisutu:
- Update `BUILD_YEAR` in `src/Calculator/Aboutflyout.xaml.cpp` according to #320

- manual

*typo in branch name, don't mind*


dovisutu:
@rudyhuyn Tests and builds are done

PR #896, opened 2020-01-03
Focus ClearEntry button after Clear button is used and vice versa

### LINKED ISSUE #860: Pressing enter in scientific mode opens navigation instead of displaying the result
In scientific mode, after pressing the Enter button, navigation menu opens instead of displaying the result.

**Steps To Reproduce**

1. Go to Scientific mode.
2. Click on C.
3. Enter 2+2 on keyboard.
4. Press Enter on keyboard.

**Expected behavior**
Result is displayed.

**Actual behavior**
Navigation menu is opened.

 - OS Build: 10.0.17134.0
 - Architecture: X64
 - Region: en-US
 - Dev Version Installed: -

**Requested Assignment**

joseartrivera:
- In scientific and programmer modes we hide/show either the Clear or ClearEntry button depending on if there is input to clear. When focus is lost, focus can go to the backspace button or as shown in #860 perhaps the navigation button. This PR fixes the focus issue, which will hopefully also fix the bug reported in #860.

- Manual tests

### DIFF
--- a/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
+++ b/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
@@ -490,6 +490,7 @@
                                        AutomationProperties.AutomationId="clearButton"
                                        ButtonId="Clear"
                                        Content="C"
+                                       LostFocus="ClearButton_LostFocus"
                                        Visibility="{x:Bind Model.IsInputEmpty, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}"/>
 
             <controls:CalculatorButton x:Name="ClearEntryButton"
@@ -499,6 +500,7 @@
                                        AutomationProperties.AutomationId="clearEntryButton"
                                        ButtonId="ClearEntry"
                                        Content="CE"
+                                       LostFocus="ClearEntryButton_LostFocus"
                                        Visibility="{x:Bind Model.IsInputEmpty, Mode=OneWay, Converter={StaticResource BooleanToVisibilityNegationConverter}}"/>
 
             <controls:CalculatorButton x:Name="BackSpaceButton"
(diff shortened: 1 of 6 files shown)


src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml
```diff
@@ -490,6 +490,7 @@
                                        AutomationProperties.AutomationId="clearButton"
                                        ButtonId="Clear"
                                        Content="C"
+                                       LostFocus="ClearButton_LostFocus"
                                        Visibility="{x:Bind Model.IsInputEmpty, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}"/>
```
grochocki: Ideally, you want to avoid programmatically handling focus whenever you can. Instead of adding focus management code, can we combine these buttons into one element in XAML that switches between Clear and ClearEntry functionality based on IsInputEmpty?

joseartrivera: We could do that, it does open up a couple more questions.

Should we combine the two shortcut keys also? ClearEntry has an additional 'delete' shortcut key.
Should we also swap the automation names or combine them into one?

grochocki: > Should we combine the two shortcut keys also?

No, I think we would want to keep separate shortcuts.

> Should we also swap the automation names or combine them into one?

I imagine it would be a single automation name.

> What about in standard mode when they are still two different buttons?

Two separate buttons here, but then you also don't have any funkiness with focus.

> If we want to keep those two things separate I think it is simpler to have two buttons instead of trying to swap out every property.

IMO, swapping out some of the display properties + shortcut is cleaner than wading into focus management. Focus issues are a lot trickier to repro/debug.

joseartrivera: I agree that is probably a more robust fix. I tried working on your solution by making a general purpose ternary to two generic object converter but hit some issues implementing in C++. I will try again later but I think we should get this fix in soon since it is a big annoyance to customers.

PR #903, opened 2020-01-07
Remove ARM64 from the package since Graphing mode does not support AR…

sanderl:
Update the package yaml to remove ARM64 from the package since Graphing mode does not support ARM64.

### DIFF
--- a/build/pipelines/templates/package-appxbundle.yaml
+++ b/build/pipelines/templates/package-appxbundle.yaml
@@ -7,14 +7,12 @@ jobs:
     - Buildx64
     - Buildx86
     - BuildARM
-    - BuildARM64
   condition: |
     and
     (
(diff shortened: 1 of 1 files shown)


build/pipelines/templates/package-appxbundle.yaml
```diff
@@ -7,14 +7,12 @@ jobs:
     - Buildx64
     - Buildx86
     - BuildARM
-    - BuildARM64
   condition: |
     and
     (
       in(dependencies.Buildx64.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
       in(dependencies.Buildx86.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
       in(dependencies.BuildARM.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
```
EriWong:
```suggestion
      in(dependencies.BuildARM.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')
```

PR #904, opened 2020-01-07
Fix crash when setting graph options to very large numbers

### LINKED ISSUE #905: Graphing mode crashes when large numbers are used in the graph settings
**Describe the bug**
Graphing mode crashes when large numbers are used in the graph settings

**Steps To Reproduce**
1. Go to graph mode
2. Put in large numbers for graph settings
3. Hover over a graphed line
4. Crashes

**Expected behavior**
Crash does not happen

**Requested Assignment**

joseartrivera:
- Update the API to the new API that uses doubles instead of floats

- Manual tests


joseartrivera:
> Make sure to update the internal build version to whatever version this PR will be in the Calculator-Internal pipeline. You'll need to update the ..\build\pipelines\templates\build-app-internal.yaml file. I had a recent PR with a similar change

Updated

build/pipelines/templates/build-app-internal.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       downloadDirectory: $(Build.SourcesDirectory)
       vstsFeed: WindowsApps
       vstsFeedPackage: calculator-internals
-      vstsPackageVersion: 0.0.31
+      vstsPackageVersion: 0.0.32
```
sanderl: I just checked your change is in version 0.0.33

```suggestion
      vstsPackageVersion: 0.0.33
```

joseartrivera: You are right! I must have resolved the merge conflicts wrong. Thanks!

PR #910, opened 2020-01-08
Use path instead of glyph for active tracing

EriWong:
- Updates the glyph in active tracing to use a path instead.
- Moves the initial position closer to the center of the graph surface
- Resets the position of the pointer when the calculator is resized

Manually

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -516,8 +516,8 @@
                 <Path x:Name="CursorPath"
                       Width="18"
                       Height="18"
-                      Fill="White"
-                      Stroke="Black"
+                      Fill="{ThemeResource SystemColorHighlightTextColor}"
+                      Stroke="{ThemeResource SystemColorWindowTextColor}"
                       StrokeThickness="1"
                       Data="M0 0 l1371 1371 H538 l-538 538 Z"
                       Stretch="Uniform"/>
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -339,6 +339,7 @@
         <Grid x:Name="LeftGrid"
               Grid.Row="1"
               Padding="0,4,0,0"
+              SizeChanged="LeftGrid_SizeChanged"
               Visibility="{x:Bind ShouldDisplayPanel(IsSmallState, SwitchModeToggleButton.IsChecked.Value, x:True), Mode=OneWay}">
             <Grid.Resources>
                 <ResourceDictionary>
@@ -434,16 +435,18 @@
                 </StackPanel>
             </Border>
             <Border x:Name="TracePointer"
-                    Width="16"
+                    Width="18"
                     Height="18"
-                    Margin="48,50,0,0"
                     HorizontalAlignment="Left"
                     VerticalAlignment="Top"
                     Visibility="Collapsed">
-                <FontIcon Foreground="Red"
-                          FontFamily="{StaticResource CalculatorFontFamily}"
-                          FontSize="18"
-                          Glyph="&#xE3B3;"/>
+                <Path Width="18"
+                      Height="18"
+                      Fill="White"
+                      Stroke="Black"
+                      StrokeThickness="1"
+                      Data="M0 0 l1371 1371 H538 l-538 538 Z"
+                      Stretch="Uniform"/>
             </Border>
             <Border MinWidth="36"
                     Margin="0,0,12,12"
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -508,20 +508,19 @@
                     </Button>
                 </StackPanel>
             </Border>
-            <Grid x:Name="TracePointer"
-                  HorizontalAlignment="Left"
-                  VerticalAlignment="Top"
-                  Visibility="Collapsed">
-                <Border x:Name="CursorShadow"/>
-                <Path x:Name="CursorPath"
-                      Width="18"
-                      Height="18"
-                      Fill="{ThemeResource SystemColorHighlightTextColor}"
-                      Stroke="{ThemeResource SystemColorWindowTextColor}"
-                      StrokeThickness="1"
-                      Data="M0 0 l1371 1371 H538 l-538 538 Z"
-                      Stretch="Uniform"/>
-            </Grid>
+            <Canvas>
+                <Grid x:Name="TracePointer" Visibility="Collapsed">
+                    <Border x:Name="CursorShadow"/>
+                    <Path x:Name="CursorPath"
+                          Width="18"
+                          Height="18"
+                          Fill="{ThemeResource SystemColorHighlightTextColor}"
+                          Stroke="{ThemeResource SystemColorWindowTextColor}"
+                          StrokeThickness="1"
+                          Data="M0 0 l1371 1371 H538 l-538 538 Z"
+                          Stretch="Uniform"/>
+                </Grid>
+            </Canvas>
             <Border MinWidth="36"
                     Margin="0,0,12,12"
                     HorizontalAlignment="Right"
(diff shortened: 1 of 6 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -82,6 +82,13 @@ GraphingCalculator::GraphingCalculator()
     virtualKey->Key = (VirtualKey)187; // OemAdd key
     virtualKey->Modifiers = VirtualKeyModifiers::Control;
     ZoomInButton->KeyboardAccelerators->Append(virtualKey);
+
+    // add shadow to the trace pointer when not in high contrast mode
+    auto accessibilitySettings = ref new Windows::UI::ViewManagement::AccessibilitySettings();
+    if (!accessibilitySettings->HighContrast)
```
joseartrivera: I don't think this takes into account switching to high contrast while the app is running

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -537,9 +543,30 @@ void GraphingCalculator::DisplayGraphSettings()
     flyoutGraphSettings->ShowAt(GraphSettingsButton);
 }
 
+void CalculatorApp::GraphingCalculator::AddShadow()
```
joseartrivera: Really liking how this looks! I think a drop shadow to the trace value popup would also look good, but maybe we can do this later.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -507,18 +508,20 @@
...
-                    Margin="48,50,0,0"
-                    HorizontalAlignment="Left"
-                    VerticalAlignment="Top"
-                    Visibility="Collapsed">
-                <FontIcon Foreground="Red"
-                          FontFamily="{StaticResource CalculatorFontFamily}"
-                          FontSize="18"
-                          Glyph="&#xE3B3;"/>
-            </Border>
+            <Grid x:Name="TracePointer"
```
joseartrivera: Minor bug, but if I have the TracePointer over a line, the trace value pop up disappears when my real mouse leaves the graph area. We can probably tackle it later if it isn't straightfoward, probably something to do with the state that is handled in grapher.cpp

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -537,9 +543,30 @@ void GraphingCalculator::DisplayGraphSettings()
...
 {
     auto flyout = static_cast<Flyout ^>(sender);
     auto graphingSetting = static_cast<GraphingSettings ^>(flyout->Content);
     args->Cancel = graphingSetting->CanBeClose();
 }
+
+void GraphingCalculator::LeftGrid_SizeChanged(Object ^ /*sender*/, SizeChangedEventArgs ^ e)
+{
+    // Initialize the pointer to the correct location to match initial value in GraphControl\DirectX\RenderMain.cpp
+    TracePointer->Margin = Thickness(e->NewSize.Width/2 + 40, e->NewSize.Height/2 - 40, 0, 0);
```
joseartrivera: Minor bug, but if I maximize the window or go back to windowed, the trace value pop up ends up in a weird location for a brief moment if it was showing before the window size change.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -507,18 +508,20 @@
...
+                  HorizontalAlignment="Left"
+                  VerticalAlignment="Top"
+                  Visibility="Collapsed">
+                <Border x:Name="CursorShadow"/>
+                <Path x:Name="CursorPath"
+                      Width="18"
+                      Height="18"
+                      Fill="White"
```
joseartrivera: I assume this is fine in high contrast since it is the same as mouse color?

EriWong: Well, I was thinking 2 fold...  1 is that this is somewhat temporary as we'll be changing the color to match the line as a future update, and 2, white fill with black stroke will always provide contrast.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
EriWong: I think this would be a good followup piece.  The popup is a rectangle, so themeshadow is a much easier way to get that to happen.  That said, I want to keep changes in this PR related to the trace pointer.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
EriWong: Yeah... I think this isn't straightforward.  The root cause of this is because we share a popup for the trace pointer and the mouse pointer.  We'd either have to write some complex logic on when to dismiss it, or disable the mouse trace cursor logic when active tracing is off.  Alternatively, we could have 2 popups and only show one at a time (with separate visual logic for each).

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
EriWong: I'm not seeing this behavior, but maybe I'm having a different repo (might be dependent on window size?)

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
EriWong: good catch.  I'll fix that.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
rudyhuyn: Fill and Stroke should support high contrast (Suggestion: Fill = SystemColorHighlightTextColor and Stroke = SystemColorWindowText to contrast with the background)

rudyhuyn: (Windows doesn't change the color of the cursor when you switch from normal mode to high contrast mode, but I suppose users also change the size and the color of their cursors to better see it)

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -537,9 +545,35 @@ void GraphingCalculator::DisplayGraphSettings()
...
```
rudyhuyn: Why not set TracePointer.VerticalAlignment and TracePointer.HorizontalAlignment to Center and use the margin `40, -40, 0 0) to not have to manage that manually?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
EriWong: fair.  will update the colors.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
EriWong: We use the margin to reposition the cursor when arrow keys are pressed.  I think it's clearer to do the math here and have it mirror the math in RenderMain.cpp instead of having to recalculate when the point moves.  That said, I would really like to only have the math be done in one place (which is how it happens for everything except initialization), but I couldn't find a nice way to do that.

EriWong: Using a canvas to position the pointer now.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -507,18 +508,19 @@
...
-                          FontSize="18"
-                          Glyph="&#xE3B3;"/>
-            </Border>
+            <Canvas>
+                <Grid x:Name="TracePointer" Visibility="Collapsed">
+                    <Border x:Name="CursorShadow"/>
+                    <Path x:Name="CursorPath"
+                          Width="18"
+                          Height="18"
+                          Fill="{ThemeResource SystemColorHighlightTextColor}"
```
rudyhuyn: we should use custom brush here

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
EriWong: Ah.  Yeah.  This is something that I think should probably be addressed as a separate commit as it isn't the result of this change, nor is it related to the pointer code.

PR #918, opened 2020-01-09
Update group policy check to look in HKCU instead of HKLM for allowing graphing mode

### LINKED ISSUE #919: Graphing Calculator can still be accessed on desktop after disabling it via Group Policy and/or InTune policy

sanderl:
- Update the registry path to HKCU instead of HKLM when it checks the group policy for allowing graphing mode

- Manually tested using the Group Policy editor to set the AllowGraphingMode policy.

### DIFF
--- a/src/CalcViewModel/Common/NavCategory.cpp
+++ b/src/CalcViewModel/Common/NavCategory.cpp
@@ -89,7 +89,7 @@ bool IsGraphingModeEnabled()
     DWORD bufferSize{ sizeof(allowGraphingCalculator) };
     // Make sure to call RegGetValueW only on Windows 10 1903+
     if (RegGetValueW(
-            HKEY_LOCAL_MACHINE,
+            HKEY_CURRENT_USER,
             L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\Calculator",
             L"AllowGraphingCalculator",
             RRF_RT_REG_DWORD | RRF_RT_REG_BINARY,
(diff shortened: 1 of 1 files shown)


src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -89,7 +89,7 @@ bool IsGraphingModeEnabled()
     DWORD bufferSize{ sizeof(allowGraphingCalculator) };
     // Make sure to call RegGetValueW only on Windows 10 1903+
     if (RegGetValueW(
-            HKEY_LOCAL_MACHINE,
+            HKEY_CURRENT_USER,
             L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\Calculator",
             L"AllowGraphingCalculator",
             RRF_RT_REG_DWORD | RRF_RT_REG_BINARY,
```
mcooley: nit: while you're here, this can be RRF_RT_DWORD

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -89,10 +89,10 @@ bool IsGraphingModeEnabled()
     DWORD bufferSize{ sizeof(allowGraphingCalculator) };
     // Make sure to call RegGetValueW only on Windows 10 1903+
     if (RegGetValueW(
-            HKEY_LOCAL_MACHINE,
+            HKEY_CURRENT_USER,
```
rudyhuyn: We should probably check both, a group policy can be defined per machine or per user. I suppose in the case of shared devices, machine policy can be useful.

sanderl: Our policy can only be set per user. I tried to set it per machine using the Group Policy editor and it was not available.

PR #922, opened 2020-01-09
Making a few improvements in Grapher

gsfreema:
- Removed unused variable
- Used copy_if in GetGraphableEquations
- Using ArrayReference in GetGraphBitmapStream so bytes aren't copied an extra time

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -198,7 +198,6 @@ namespace GraphControl
 
     KeyGraphFeaturesInfo ^ Grapher::AnalyzeEquation(Equation ^ equation)
     {
-        auto result = ref new KeyGraphFeaturesInfo();
         if (auto graph = GetGraph(equation))
         {
             if (auto analyzer = graph->GetAnalyzer())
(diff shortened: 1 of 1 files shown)


src/GraphControl/Control/Grapher.cpp
```diff
@@ -198,7 +198,6 @@ namespace GraphControl
 
     KeyGraphFeaturesInfo ^ Grapher::AnalyzeEquation(Equation ^ equation)
     {
-        auto result = ref new KeyGraphFeaturesInfo();
```
rudyhuyn: 👍

src/GraphControl/Control/Grapher.cpp
```diff
@@ -465,15 +464,8 @@ namespace GraphControl
     vector<Equation ^> Grapher::GetGraphableEquations()
     {
         vector<Equation ^> validEqs;
-
-        for (Equation ^ eq : Equations)
```
rudyhuyn: The change is correct, but because `std::copy_if` provides the same performances than the existing loop, I would suggest not to change this part. The original code is straightforward and easier to understand for contributors with little C++ experience.

gsfreema: That's fair.  I've reached a point where I look at loops and see if it's an existing algorithm.  Sometimes I go overboard.  In C#, I probably would have done `Equations.Where(eq => eq.IsGraphableEquation()).ToList()` and that's not even as efficient as a raw loop due to Linq overheads.

PR #930, opened 2020-01-13
Remove ARM64 from Release builds

sanderl:
- Remove ARM64 from Release builds since Graphing mode does not support ARM64

### DIFF
--- a/build/pipelines/azure-pipelines.release.yaml
+++ b/build/pipelines/azure-pipelines.release.yaml
@@ -30,11 +30,6 @@ jobs:
     platform: ARM
     condition: not(eq(variables['Build.Reason'], 'PullRequest'))
 
-- template: ./templates/build-app-internal.yaml
-  parameters:
-    platform: ARM64
-    condition: not(eq(variables['Build.Reason'], 'PullRequest'))
-
 - template: ./templates/run-ui-tests.yaml
   parameters:
     platform: x64


EwoutH:
@sanderl Do you by any chance know why the graphing mode isn't supported on Arm64 platforms? Is there a specif package or API that blocks it?

HowardWolosky:
@EwoutH - At present time, the [Microsoft-internal library](https://github.com/microsoft/calculator#graphing-mode) that the release builds consume as a pre-compiled .dll does not produce an Arm64 package.  If/when it does, this change can be reversed.

rudyhuyn:
I would also like to add that in the meantime, 
users using Windows 10 on ARM (ARM64 PCs) will receive the ARM version of the calculator, providing great performance without emulation.

EwoutH:
@HowardWolosky Thanks for the explanation. Can you pass a request upstream to compile Arm64 packages? Especially with [Microsoft pushing](https://educationblog.microsoft.com/en-us/2020/01/our-vision-to-close-the-broadband-gap-improve-teacher-training-and-boost-accessibility-in-the-classroom/) Arm64 for the education market, this looks like low-hanging fruit.

grochocki:
> Can you pass a request upstream to compile Arm64 packages?

Yep, we have already shared this request with our partner team.

PR #958, opened 2020-01-17
Fix graphing crash when plotting equations with errors

### LINKED ISSUE #957: Graphing mode crashes on some equations that contain errors
**Describe the bug**
Graphing mode crashes on some equations that contain errors

**Steps To Reproduce**
1. Go to graphing mode
2. Type sqrt(y)
3. See error, hover mouse over graph, see crash

**Expected behavior**
No crash

joseartrivera:
- The call to initialize an empty graph was replaced with a call that tries to graph the error equation

- Type sqrt(y) and hover mouse over graph

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -162,14 +162,11 @@ namespace GraphControl
 
     void Grapher::OnEquationChanged(Equation ^ equation)
     {
-        // If the equation was previously valid, we should try to graph again in the event of the failure
-        bool shouldRetry = equation->IsValidated;
-
         // Reset these properties if the equation is requesting to be graphed again
         equation->HasGraphError = false;
         equation->IsValidated = false;
 
-        TryPlotGraph(false, shouldRetry);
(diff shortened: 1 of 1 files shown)


joseartrivera:
> I was still able to repro a crash when I inputted a valid equation, then an invalid one.

Good catch, I was able to fix this on our side but it will result in an empty graph, temporarily not rendering existing valid equations until the equation with an error is fixed by the user. Better than crashing though.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -314,14 +314,18 @@ namespace GraphControl
                         initResult = nullopt;
                     }
                 }
+                else
+                {
+                    shouldKeepPreviousGraph = false;
```
sanderl: Can we remove the equation with an error and try to graph again?

joseartrivera: As discussed, this logic was already in place but was not triggering correctly, probably because the way equations get added was slightly changed at some point. Fixed now!

PR #967, opened 2020-01-22
Update mocks to allow for more UI to appear

joseartrivera:
# Description of the changes:
- Currently the community cannot access a lot of the features for the Graphing UI because most equations will return as errors, this change fleshes out the mocks so that more of the UI can be accessed (analyzer, equation colors, variables).
- Would be nice in the future to add mock data to the analyzer for a full view instead of an error page

- Manual tests


joseartrivera:
> The code looks good. Testing it, I wasn't expected the variables UI with my equation. You might want to add a section to the readme to describe the mock behavior.

There actually already is a section referencing the mock implementation!

PR #1014, opened 2020-01-31
Allow rendering the graph on a background thread

### LINKED ISSUE #980: Graphing engine rendering should occur on background thread
**Describe the bug**
There are known performance issues with rendering equations on the graph due to how we currently process input.

**Expected behavior**
We should move graphing engine rendering to a background thread.

joseartrivera:
- Move graph rendering to the background thread
- Allow variable values larger than the min/max to change the min/max instead of resetting
- Fix a bug where the equation box was no longer clearing bold/underline formatting

- Manual tests
-Render m*sin(x*m)+m and drag the variable slider. Notice that the graph renders while the slider moves without stuttering.

### DIFF
--- a/src/Calculator/Controls/MathRichEditBox.cpp
+++ b/src/Calculator/Controls/MathRichEditBox.cpp
@@ -12,6 +12,7 @@ using namespace std;
 using namespace Windows::ApplicationModel;
 using namespace Windows::UI::Xaml;
 using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Text;
 using namespace Windows::Foundation::Collections;
 using namespace Windows::System;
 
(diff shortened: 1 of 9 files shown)


src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -178,6 +179,15 @@ void MathRichEditBox::BackSpace()
 
 void MathRichEditBox::SubmitEquation(EquationSubmissionSource source)
 {
+    // Clear formatting since the graph control doesn't work with bold/underlines
+    auto range = this->TextDocument->GetRange(0, this->TextDocument->Selection->EndPosition);
+
+    if (range != nullptr)
+    {
+        range->CharacterFormat->Bold = FormatEffect::Off;
```
sanderl: Can we clear all character formatting? Copy/Paste will keep the paragraph formatting (i.e. spacing above or below the text)

PR #1021, opened 2020-02-04
Enables analysis on functions with variables

### LINKED ISSUE #990: Function analysis is not working for equations with variables
**Describe the bug**
Function analysis is not working for equations with variables.

**Expected behavior**
Function analysis should work for equations with variables, leveraging the value of the variable at the time of analysis.

**Screenshots**
image

sanderl:
- Set the graph option to allow analysis for functions with variables to true

- manually

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -199,6 +199,8 @@ namespace GraphControl
         {
             if (auto analyzer = graph->GetAnalyzer())
             {
+                graph->GetOptions().SetAllowKeyGraphFeaturesForFunctionsWithParameters(true);
+                auto isVariableAnalysisAllowed = graph->GetOptions().GetAllowKeyGraphFeaturesForFunctionsWithParameters();
                 if (analyzer->CanFunctionAnalysisBePerformed())
                 {
                     if (S_OK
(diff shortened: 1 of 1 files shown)


src/GraphControl/Control/Grapher.cpp
```diff
@@ -199,6 +199,8 @@ namespace GraphControl
         {
             if (auto analyzer = graph->GetAnalyzer())
             {
+                graph->GetOptions().SetAllowKeyGraphFeaturesForFunctionsWithParameters(true);
```
joseartrivera: I think we should consider replacing this with a call to UpdateGraphOptions, that way we can keep this graph creation consistent with the regular graph creation.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -199,6 +199,8 @@ namespace GraphControl
...
+
```
joseartrivera: Not related to this PR, but since this is creating a new graph, are we not passing along the units (Radians, degrees, gradians)? Would this alter the output of the analysis?

sanderl: The analysis updates according to the trig units.

joseartrivera: Is that correct? If the graph is set to degrees shouldn't the analysis also be in degrees?

joseartrivera: nvm, as we discussed this is done in the MathSolver not in the Graph!

PR #1022, opened 2020-02-06
Make unit converters stil working when users opened 'Currency' while offline

### LINKED ISSUE #973: Other converter modes stop working if we open 'Currency' in offline mode
**Describe the bug**

If we switch from the Currency Mode to any other converter mode while we are offline, the input box is either not visible or is unavailable.

**Steps To Reproduce**

1. Go to the 'Currency' mode of the calculator while you are offline. Make sure that the 'Currency' Mode doesn't have any previously fetched rates (You may reset this app from App Settings to clear its data). It should only show:
    
> Offline. Please check your Network Settings.

2. Then switch to any other mode in the 'Converter' section.
3. Notice that the input box is not available.

**Expected behavior**

The other modes should have been working even while we are offline.

**Screenshots**

Issue

Issue (2)

See that no input panel is available.

 - OS Build: 10.0.17134.619
 - Architecture: x64
 - Region: New Delhi, India

rudyhuyn:
- Modify how `UnitConverter.xaml` manages its visual state
- Modify how `UnitConverterViewModel` manages `IsCurrencyCurrentCategory` to be sure to always match be up to date
- Fix crashes related to SelectedUnitTemplate when no units are selected.

- manually

### DIFF
--- a/src/CalcViewModel/UnitConverterViewModel.cpp
+++ b/src/CalcViewModel/UnitConverterViewModel.cpp
@@ -171,9 +171,6 @@ void UnitConverterViewModel::OnCategoryChanged(Object ^ parameter)
 
 void UnitConverterViewModel::ResetCategory()
 {
-    UCM::Category currentCategory = CurrentCategory->GetModelCategory();
-    IsCurrencyCurrentCategory = currentCategory.id == NavCategory::Serialize(ViewMode::Currency);
-
     m_isInputBlocked = false;
     SetSelectedUnits();
 
--- a/src/CalcViewModel/UnitConverterViewModel.h
+++ b/src/CalcViewModel/UnitConverterViewModel.h
@@ -143,7 +143,6 @@ namespace CalculatorApp
             OBSERVABLE_OBJECT_CALLBACK(OnPropertyChanged);
 
             OBSERVABLE_PROPERTY_R(Windows::Foundation::Collections::IObservableVector<Category ^> ^, Categories);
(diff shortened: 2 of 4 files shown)


lychichem:
> The change looks good to me. I tested this out and it fixes the issue where the other converters are empty when the currency converter is empty. But it does not fix the issue with currency converter, which is expected.

Again, I have to say issue with currency converter is totally the fault of MSN API returns empty value. You can visit https://api.msn.com/Finance/Calculator/CurrenciesStaticData?apikey=qg3rEHwuH1g4nNX8ENJrcnWSl9Gk93GdQvsx2Vs7Zn&localizeFor=zh-Hans-CN (What is used in distributed version of Windows Calculator). So the best and fastet way is to call MSN team and say "hey guys, your API have bug(s) which breakes our App".

rudyhuyn:
@lychichem Hi. We are well aware of the, we've contacted the team in charge of this endpoint last week and they are currently working on resolving it. 

Currency converter had 2 issues:
- the endpoint returning an error: tracked by #1050 and handled by the MSN API team.
- other converters were non-functional if the user tried to use the currency converter and the endpoint returning an error. This bug was tracked by #973 and resolved by this PR.

The purpose of this fix wasn't to solve the issue with the endpoint but to fix other converters when the endpoint isn't accessible (including when you are offline)

Rudy

lychichem:
@rudyhuyn Yes, you are right. I've just be annoyed by the problem that I can't use currency converter, and said some rude words to your fellow. Sorry for that.

PR #1023, opened 2020-02-10
Make it clear to both compilers and programmers the end result of ResolveRuntimeWindowService is to return a service object

pi1024e:
- Stylistic change to match the fact the other functions have throw in the if block, not outside of it.

- Manual, automatic

### DIFF
--- a/src/Calculator/WindowFrameService.cpp
+++ b/src/Calculator/WindowFrameService.cpp
@@ -154,14 +154,12 @@ namespace CalculatorApp
     {
         auto service = TryResolveRuntimeWindowService(serviceId);
 
-        if (service)
-        {
-            return service;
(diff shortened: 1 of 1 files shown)


mcooley:
Hi @pi1024e, thanks for noticing that this code could be improved. It looks like this code is not actually called anywhere, so it should just be removed instead of modifying it further.

I started experimenting with removing dead code from WindowFrameService (see [this commit in my fork](https://github.com/mcooley/calculator/commit/5a6574b9d235981d0d6e4dfd8c0d126ce1be6ef3)). I didn't have time to finish, but I would be happy for someone else to take it further. You could remove the unused code, or go a step further and remove WindowFrameService completely after some refactoring.

PR #1024, opened 2020-02-10
Make it clear the point of the ResolveRuntimeWindowService is to return a service by returning at the end.

pi1024e:
## Fixes #.
Unit test errors

- Stylistic changes

- Manual
- Automatic


grochocki:
This appears to be a duplicate of #1023

PR #1041, opened 2020-02-18
Graphing Calculator Diagnostics Instrumentation

### LINKED ISSUE #1040: Refreshing Diagnostic Data to Include Graphing Mode
Updated diagnostic data to account for the new graphing mode.

sanderl:
- Add diagnostic data for graphing mode

- Manually, ensured diagnostic data is uploaded as expected


src/GraphControl/Control/Grapher.cpp
```diff
@@ -192,7 +192,16 @@ namespace GraphControl
             return;
         }
 
-        PlotGraph(true);
+        bool keepCurrentView = true;
+
+        // If the equation has changed, the IsLineEnabled state is reset.
```
joseartrivera: I don't see IsLineEnabled getting reset when the equation is changed

src/GraphControl/Control/Grapher.cpp
```diff
@@ -246,6 +255,26 @@ namespace GraphControl
...
+            if (eq->HasGraphError)
+            {
+                invalid++;
+            }
+            if (eq->IsValidated)
+            {
+                valid++;
+            }
+        }
+        if (!m_trigUnitsChanged)
```
joseartrivera: Maybe it makes more sense to only run a render pass when the trig units change instead of re-plotting, would that work?

src/GraphControl/Control/Grapher.cpp
```diff
@@ -246,6 +255,26 @@ namespace GraphControl
                 co_await TryUpdateGraph(keepCurrentView);
             }
         }
+
+        int valid = 0;
+        int invalid = 0;
+        for (Equation ^ eq : Equations)
```
joseartrivera: If you move this telemetry into SetEquationErrors and SetEquationsAsValid you can re-use the loops in there to get a count and only loop over the equations once. You also avoid the risk of miscounting since you are looping over all Equations while those methods loop only the graphable equations.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -425,6 +454,8 @@ namespace GraphControl
 
         Variables = updatedVariables;
         VariablesUpdated(this, Variables);
+
+        TraceLogger::GetInstance()->LogVariableCountChanged(Variables->Size);
```
joseartrivera: I don't think this necessarily fires only when the variable count changes

src/GraphControl/Control/Grapher.h
```diff
@@ -291,7 +293,9 @@ public
 
         void SetEquationsAsValid();
         void SetEquationErrors();
-        std::optional<std::vector<std::shared_ptr<Graphing::IEquation>>> TryInitializeGraph(bool keepCurrentView, _In_ const Graphing::IExpression* graphingExp = nullptr);
+        std::optional<std::vector<std::shared_ptr<Graphing::IEquation>>>
```
joseartrivera: We should move this back to 1 line, maybe got auto fixed?

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -234,6 +234,8 @@ void EquationTextBox::OnRemoveButtonClicked(Object ^ sender, RoutedEventArgs ^ e
         m_equationButton->IsChecked = false;
     }
 
+    TraceLogger::GetInstance()->LogGraphButtonClicked(L"RemoveFunction");
```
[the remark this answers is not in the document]
sanderl: Since we attempt to perform analysis every time the analysis button is clicked, I decided only to log the analysis since we can get a count to determine the number of times the button is clicked.

src/GraphControl/Control/Grapher.cpp
sanderl: This is fixed. It should be working now. Let me know what you think of the changes.

src/GraphControl/Control/Grapher.cpp
sanderl: I think we have to re-plot because the values of the points change when the trig units change

src/GraphControl/Control/Grapher.cpp
sanderl: We want to count all equations even if they are hidden on the graph. So only logging the graphable equations could give us an unclear picture of how many equations are in the equation list.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -125,6 +125,15 @@ void EquationInputArea::EquationTextBox_Submitted(Object ^ sender, MathRichEditB
         || (submission->Source == EquationSubmissionSource::FOCUS_LOST && submission->HasTextChanged && eq->Expression != nullptr
             && eq->Expression->Length() > 0))
     {
+        auto eqButton = VisualTree::FindDescendantByName(tb, "EquationButton");
```
joseartrivera: The IsLineEnabled property should be bound to a property on the EquationTextBox that has a TemplateBinding to the EquationButton->IsChecked property. Using the VisualTree is a more likely to break silently in the future as people change the XAML and eqButton suddenly becomes a nullptr with no compile error.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -456,11 +455,7 @@ namespace GraphControl
         Variables = updatedVariables;
```
joseartrivera: You can see if the variable count changed here by checking Variables.Count against updatedVariables.Count before this assignment.

src/GraphControl/Models/KeyGraphFeaturesInfo.cpp
```diff
@@ -59,12 +59,28 @@ KeyGraphFeaturesInfo ^ KeyGraphFeaturesInfo::Create(IGraphFunctionAnalysisData d
...
 KeyGraphFeaturesInfo ^ KeyGraphFeaturesInfo::Create(CalculatorApp::AnalysisErrorType type)
 {
     auto res = ref new KeyGraphFeaturesInfo();
     res->AnalysisError = type;
+
+    String ^ errorMessage = ref new String();
+
+    if (res->AnalysisError == CalculatorApp::AnalysisErrorType::AnalysisCouldNotBePerformed)
+    {
+        errorMessage = L"Analysis could not be performed.";
```
joseartrivera: Why not just log the enum?

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -35,12 +35,22 @@ namespace CalculatorApp
...
+    constexpr auto EVENT_NAME_GRAPH_SETTINGS_CHANGED = L"GraphSettingsChanged";
 
     constexpr auto EVENT_NAME_EXCEPTION = L"Exception";
 
     constexpr auto PDT_PRIVACY_DATA_TAG = L"PartA_PrivTags";
     constexpr auto PDT_PRODUCT_AND_SERVICE_USAGE = 0x0000'0000'0200'0000u;
 
+    constexpr auto SessionGuid = L"SessionGuid";
+    constexpr auto CalcMode = L"CalcMode";
+    constexpr auto GraphingMode = L"Graphing";
```
rudyhuyn: Can we use upper cases to be consistent with other constants?

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -226,6 +227,7 @@ namespace CalculatorApp::ViewModel
                 if (value && m_Graph != nullptr && m_Graph->TrigUnitMode != (int)Graphing::EvalTrigUnitMode::Radians)
                 {
                     m_Graph->TrigUnitMode = (int)Graphing::EvalTrigUnitMode::Radians;
+                    TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::TrigUnits);
```
rudyhuyn: Very nit: let's call this method after the 3 RaisePropertyChanged to update the UI as fast as possible.

src/CalcViewModel/Common/TraceLogger.h
```diff
@@ -28,8 +25,37 @@ namespace CalculatorApp
...
+        Share,
+        ZoomIn,
+        ZoomOut,
+        ZoomReset
+    };
+
+    public enum class VisibilityButtonState
+    {
+        Show,
+        Hide
```
rudyhuyn: Except if we consider adding a 3rd state, let's use a Boolean instead

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -405,8 +417,49 @@ void EquationInputArea::VariableAreaTapped(Object ^ sender, TappedRoutedEventArg
...
-  
+
 void EquationInputArea::EquationTextBox_EquationFormatRequested(Object ^ sender, MathRichEditBoxFormatRequest ^ e)
 {
     EquationFormatRequested(sender, e);
 }
+
+void EquationInputArea::Slider_ValueChanged(Object ^ sender, RangeBaseValueChangedEventArgs ^ e)
+{
+    static Map<String ^, Delayer ^> ^ variableSliders = ref new Map<String ^, Delayer ^>();
```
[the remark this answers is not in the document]
rudyhuyn: It will also be an issue for multi instance. Let's use a field instead

src/GraphControl/Models/KeyGraphFeaturesInfo.cpp
```diff
@@ -59,12 +59,28 @@ KeyGraphFeaturesInfo ^ KeyGraphFeaturesInfo::Create(IGraphFunctionAnalysisData d
...
 }
 
 KeyGraphFeaturesInfo ^ KeyGraphFeaturesInfo::Create(CalculatorApp::AnalysisErrorType type)
 {
     auto res = ref new KeyGraphFeaturesInfo();
     res->AnalysisError = type;
+
+    String ^ errorMessage = ref new String();
+
+    if (res->AnalysisError == CalculatorApp::AnalysisErrorType::AnalysisCouldNotBePerformed)
```
rudyhuyn: Nit: let's use `if (type ==` instead

src/GraphControl/Models/KeyGraphFeaturesInfo.cpp
```diff
@@ -59,12 +59,28 @@ KeyGraphFeaturesInfo ^ KeyGraphFeaturesInfo::Create(IGraphFunctionAnalysisData d
...
+        res->TooComplexFeatures > 0 ? L"Some features were too complex to calculate. TooComplexFeatures Flag: " + res->TooComplexFeatures : ref new String());
     return res;
 }
 
 KeyGraphFeaturesInfo ^ KeyGraphFeaturesInfo::Create(CalculatorApp::AnalysisErrorType type)
 {
     auto res = ref new KeyGraphFeaturesInfo();
     res->AnalysisError = type;
+
+    String ^ errorMessage = ref new String();
```
rudyhuyn: Let's remove the `ref new String()` and add:
```
else
{
  errorMessage = L"unknown";
}
```

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -379,9 +389,86 @@ namespace CalculatorApp
...
+
+    void TraceLogger::LogShowHideButtonClicked(bool isHide)
+    {
+        if (!GetTraceLoggingProviderEnabled())
+            return;
+
+        LoggingFields fields{};
+        fields.AddGuid(SESSION_GUID, sessionGuid);
+        fields.AddString(CALC_MODE, GRPAHING_MODE);
+        fields.AddBoolean(L"ButtonState", isHide);
```
joseartrivera: It would be clearer from the data view if this were called something like “IsButtonHidden”, true/false rather than “ButtonState”, true/false

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -379,9 +389,86 @@ namespace CalculatorApp
...
         LoggingFields fields{};
-        fields.AddGuid(L"SessionGuid", sessionGuid);
-        fields.AddString(L"Mode", NavCategory::GetFriendlyName(mode)->Data());
+        fields.AddGuid(SESSION_GUID, sessionGuid);
+        fields.AddString(CALC_MODE, NavCategory::GetFriendlyName(mode)->Data());
         fields.AddUInt64(PDT_PRIVACY_DATA_TAG, PDT_PRODUCT_AND_SERVICE_USAGE);
         LogLevel2Event(EVENT_NAME_INPUT_PASTED, fields);
     }
+
+    void TraceLogger::LogShowHideButtonClicked(bool isHide)
```
joseartrivera: Parameter might better named as isHidden

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -244,6 +248,7 @@ namespace CalculatorApp::ViewModel
                 if (value && m_Graph != nullptr && m_Graph->TrigUnitMode != (int)Graphing::EvalTrigUnitMode::Degrees)
                 {
                     m_Graph->TrigUnitMode = (int)Graphing::EvalTrigUnitMode::Degrees;
+                    TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::TrigUnits);
```
joseartrivera: I think Rudy’s comment for moving under the RaisePropertyChanged applies here and below

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -382,3 +386,8 @@ void EquationTextBox::OnEquationFormatRequested(Object ^ sender, MathRichEditBox
 {
     EquationFormatRequested(this, args);
 }
+
+void EquationTextBox::OnIsEquationLineDisabledPropertyChanged(bool /*oldValue*/, bool newValue)
+{
+    m_equationButton->IsChecked = newValue;
```
joseartrivera: Can you templatebind this in xaml instead?

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -405,8 +414,46 @@ void EquationInputArea::VariableAreaTapped(Object ^ sender, TappedRoutedEventArg
...
-  
+
 void EquationInputArea::EquationTextBox_EquationFormatRequested(Object ^ sender, MathRichEditBoxFormatRequest ^ e)
 {
     EquationFormatRequested(sender, e);
 }
+
+void EquationInputArea::Slider_ValueChanged(Object ^ sender, RangeBaseValueChangedEventArgs ^ e)
+{
+    //variableSliders = ref new Map<String ^, Delayer ^>();
```
joseartrivera: You can remove this

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -405,8 +414,46 @@ void EquationInputArea::VariableAreaTapped(Object ^ sender, TappedRoutedEventArg
...
+    auto slider = static_cast<Slider ^>(sender);
+
+    if (slider->FocusState == Windows::UI::Xaml::FocusState::Unfocused)
```
joseartrivera: Would be good to add a comment as to why this is here so we don’t delete it later

src/CalcViewModel/Common/TraceLogger.cpp
sanderl: I updated it to isHideButton

src/CalcViewModel/Common/TraceLogger.cpp
sanderl: I updated it to isHIdeButton since this is logging if the Hide or Show function button clicked

src/GraphControl/Logger/TraceLogger.cpp
```diff
@@ -0,0 +1,154 @@
...
+        fields.AddUInt64(L"ConcurrentInvalidFunctions", currentInvalidEquations);
+        fields.AddUInt64(L"TotalValidFunctions", TotalValidEquations);
+        fields.AddUInt64(L"TotalInvalidFunctions", TotalInvalidEquations);
+        fields.AddUInt64(PDT_PRIVACY_DATA_TAG, PDT_PRODUCT_AND_SERVICE_USAGE);
+        LogLevel2Event(EVENT_NAME_EQUATION_COUNT_CHANGED, fields);
+    }
+
+    void TraceLogger::LogFunctionAnalysisPerformed(int analysisErrorType, uint32 tooComplexFlag)
+    {
+        LoggingFields fields{};
```
rudyhuyn: same question here

src/GraphControl/Logger/TraceLogger.cpp
```diff
@@ -0,0 +1,154 @@
...
+        }
+        else if (currentInvalidEquations > PreviousInvalidEquations)
+        {
+            TotalInvalidEquations++;
+        }
+
+        PreviousValidEquations = currentValidEquations;
+        PreviousInvalidEquations = currentInvalidEquations;
+
+        LoggingFields fields{};
```
rudyhuyn: Any reason why we don't check `GetTraceLoggingProviderEnabled` here?

src/GraphControl/Logger/TraceLogger.cpp
```diff
@@ -0,0 +1,154 @@
...
+    constexpr int64_t MICROSOFT_KEYWORD_LEVEL_1 = 0;
+    constexpr int64_t MICROSOFT_KEYWORD_LEVEL_2 = 0;
+    constexpr int64_t MICROSOFT_KEYWORD_LEVEL_3 = 0;
+#endif
+
+#pragma region TraceLogger setup and cleanup
+
+    TraceLogger::TraceLogger()
+        : g_calculatorProvider(
+            L"MicrosoftCalculator",
```
rudyhuyn: Should we use a more explicit name to distinct these logs from https://github.com/microsoft/calculator/blob/577aafb3f4b405c5cf64444a3f9e8e147259bf8f/src/CalcViewModel/Common/TraceLogger.cpp#L61? "MicrosoftGraphingCalculator"?

src/GraphControl/Logger/TraceLogger.h
```diff
@@ -0,0 +1,46 @@
...
+        void LogLevel2Event(std::wstring_view eventName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+        void LogLevel3Event(std::wstring_view eventName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+
+        std::unique_ptr<TraceActivity> CreateTraceActivity(std::wstring_view activityName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+
+        winrt::Windows::Foundation::Diagnostics::LoggingChannel g_calculatorProvider;
+
+        GUID sessionGuid;
+
+        winrt::Windows::Foundation::Diagnostics::LoggingActivity m_appLaunchActivity;
```
mcooley: Looks like this is unused?

src/GraphControl/Logger/TraceLogger.h
```diff
@@ -0,0 +1,46 @@
...
+        // The lower the level of the keyword, the higher the possibility that the corresponding event may be sampled.
+        void LogLevel1Event(std::wstring_view eventName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+        void LogLevel2Event(std::wstring_view eventName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+        void LogLevel3Event(std::wstring_view eventName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+
+        std::unique_ptr<TraceActivity> CreateTraceActivity(std::wstring_view activityName, winrt::Windows::Foundation::Diagnostics::LoggingFields fields);
+
+        winrt::Windows::Foundation::Diagnostics::LoggingChannel g_calculatorProvider;
+
+        GUID sessionGuid;
```
mcooley: The GraphControl has a different session ID than the rest of the app? That's weird. They should be shared.

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -25,6 +25,8 @@ using namespace Windows::UI::Xaml::Input;
 using namespace GraphControl;
 using namespace Calculator::Utils;
 
+using nanoseconds = std::chrono::nanoseconds;
```
mcooley: Unused?

src/Calculator/Utils/Delayer.h
```diff
@@ -0,0 +1,27 @@
...
+        void Start();
+        void ResetAndStart();
+        void Stop();
+
+        event Windows::Foundation::EventHandler<Platform::Object ^> ^ Action;
+
+     private:
+        void Timer_Tick(Platform::Object ^ sender, Platform::Object ^ e);
+
+        Windows::UI::Xaml::DispatcherTimer ^ m_timer;
```
mcooley: Is this the right timer to use? Probably you don't care which thread the action fires on, so a threadpool timer would be a better choice. Also, can we avoid creating one timer per slider and instead use one per instance of EquationInputArea?

src/GraphControl/Logger/TraceActivity.h
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+#include <winmeta.h>
+
+namespace GraphControl
+{
+    // RAII wrapper that automatically sends the Stop event when the class gets destructed.
+    class TraceActivity
```
mcooley: Unused?

src/CalcViewModel/Common/TraceLogger.h
```diff
@@ -28,8 +25,31 @@ namespace CalculatorApp
...
+        ZoomOut,
+        ZoomReset
+    };
+
+    public enum class LineStyleType
+    {
+        Color
+    };
+
+public ref class TraceLogger sealed
```
mcooley: nit: indent seems off

src/GraphControl/Logger/TraceLogger.cpp
sanderl: This name corresponds to the event name that is used in the diagnostics, I don't think we should change it because all of these events are from the calculator app.

src/TraceLogging/pch.cpp
```diff
@@ -0,0 +1 @@
+﻿#include "pch.h"
```
rudyhuyn: very very nit:

Add 
```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.
```

src/Calculator/Calculator.vcxproj
```diff
@@ -16,8 +16,7 @@
     <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
     <AppxDefaultResourceQualifierUAP_Contrast>black</AppxDefaultResourceQualifierUAP_Contrast>
     <AppxBundle>Always</AppxBundle>
-    <PackageCertificateKeyFile>TemporaryKey.pfx</PackageCertificateKeyFile>
```
joseartrivera: Deleted by accident?

src/Calculator/Calculator.vcxproj
```diff
@@ -227,6 +226,12 @@
   </PropertyGroup>
   <PropertyGroup>
     <GenerateProjectSpecificOutputFolder>true</GenerateProjectSpecificOutputFolder>
+    <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
```
joseartrivera: Probably left over from generating the appx, can we remove this block of new code?

src/GraphControl/Logger/TraceLogger.cpp
```diff
@@ -0,0 +1,89 @@
...
+        else if (currentInvalidEquations > PreviousInvalidEquations)
+        {
+            TotalInvalidEquations++;
+        }
+
+        PreviousValidEquations = currentValidEquations;
+        PreviousInvalidEquations = currentInvalidEquations;
+
+        auto fields = ref new LoggingFields();
+        fields->AddString(ref new String(CalcMode), ref new String(GraphingMode));
```
joseartrivera: Use StringReference instead of ref new string here and throughout the PR

src/GraphControl/Logger/TraceLogger.h
sanderl: Good catch. I got this fixed.

src/Calculator/Utils/Delayer.h
sanderl: Right now a threadpool timer works as well since the current logic doesn't access a xaml object. However, in the future, we'll be able to use the Delayer to access xaml objects if we leave the DispatcherTimer instead of updating it to a threadpool timer.

For the number of timers, we want to be able to log when each variable is used, so we want them to have their own timers. I have updated the event to log the variable name as well.

src/CalcViewModel/Common/TraceLogger.h
mcooley: indent still seems off

src/CalcViewModel/Common/TraceLogger.h
```diff
@@ -48,34 +67,22 @@ public
...
-    internal :
-        void LogStandardException(CalculatorApp::Common::ViewMode mode, std::wstring_view functionName, _In_ const std::exception& e);
-        void LogWinRTException(CalculatorApp::Common::ViewMode mode, std::wstring_view functionName, _In_ winrt::hresult_error const& e);
+        void LogShowHideButtonClicked(bool isHideButton);
+        void LogGraphButtonClicked(GraphButton buttonName);
+        void LogGraphLineStyleChanged(LineStyleType style);
+        void LogVariableChanged(Platform::String ^ inputChangedType, Platform::String ^ variableName);
+        void LogVariableSettingsChanged(Platform::String ^ setting);
+        void LogGraphSettingsChanged(GraphSettingsType settingsType);
+        internal : void LogStandardException(CalculatorApp::Common::ViewMode mode, std::wstring_view functionName, _In_ const std::exception& e);
```
mcooley: nit: looks like clang-format messed this up

src/Calculator/Calculator.vcxproj
```diff
@@ -17,7 +17,7 @@
     <AppxDefaultResourceQualifierUAP_Contrast>black</AppxDefaultResourceQualifierUAP_Contrast>
     <AppxBundle>Always</AppxBundle>
     <PackageCertificateKeyFile>TemporaryKey.pfx</PackageCertificateKeyFile>
-    <AppxPackageSigningEnabled>true</AppxPackageSigningEnabled>
+    <AppxPackageSigningEnabled>True</AppxPackageSigningEnabled>
```
mcooley: revert this

src/Calculator/Utils/Delayer.h
mcooley: DispatcherTimers come with significant overhead; we shouldn't pay for that until we actually need it.

src/Calculator/Utils/Delayer.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "pch.h"
+
+namespace CalculatorApp
+{
+public
```
mcooley: nit: weird formatting

src/GraphControl/Logger/TraceLogger.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "Common.h"
+
+
+// A trace logging provider can only be instantiated and registered once per module.
+// This class implements a singleton model ensure that only one instance is created.
+namespace GraphControl
+{
+
+public
```
mcooley: nit: format

src/GraphControl/Logger/TraceLogger.h
```diff
@@ -0,0 +1,30 @@
...
+    {
+    internal:
+        static TraceLogger ^ GetInstance();
+
+        void LogEquationCountChanged(int currentValidEquations, int currentInvalidEquations);
+        void LogFunctionAnalysisPerformed(int analysisErrorType, uint32 tooComplexFlag);
+        void LogVariableCountChanged(int variablesCount);
+
+    private:
+        // Create an instance of TraceLogger
```
mcooley: nit: comment not necessary

src/Calculator/Calculator.vcxproj
rudyhuyn: I would recommend to not revert it. "true" was a bug, VS will always try to replace it with "True" when you edit the vcproj

src/Calculator/Utils/Delayer.h
rudyhuyn: I would agree with you in general, but not especially in this particular case because we stop/restart the timer a lot (ThreadpoolTimer does not provide a way to reset a timer and you are forced to create a new timer, which will create a huge overload, especially when associated with a Slider that will spam events) and because the action is not a time-consuming task.

mcooley: Resetting the dispatchertimer will probably do just as much work as creating a new threadpool timer. I agree that the API is harder to use, though. So I guess I'm fine keeping this as written, for this particular instance.

The reason I'm being picky here is that I have tried and failed to write "generic"/"reusable" timer helpers before. They tend to become timebombs--someone else inevitably comes along and uses it without understanding the tradeoffs the original author made and gets into trouble. Maybe naming this class "DispatcherTimerDelayer" would help discourage that outcome.

sanderl: I renamed the class to DispatcherTimerDelayer

mcooley:
No further feedback from me except the things which are causing the build failures.

sanderl:
> No further feedback from me except the things which are causing the build failures.

I'm already fixing those. Thanks.

src/GraphControl/Logger/TraceLogger.h
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "Common.h"
+
+
```
rudyhuyn: Extra empty line to remove

PR #1046, opened 2020-02-24
No need for else after break

pi1024e:
- There is no need to have an else statement after an if statement that only contains a break command.

- Manual
- Automated


mcooley:
Thanks for taking a look at this. There's no difference in the behavior of the generated code, and I don't think the new style is significantly better than the old style, so we'll keep the existing code as-is.

PR #1049, opened 2020-02-25
Fix usability issues with settings flyout

### LINKED ISSUE #992: X/Y-min/max values in graph options should not change automatically when manually setting
**Describe the bug**
In attempting to lock aspect ratio of the graph window, we are adjusting X/Y-min/max values automatically. This behavior is not straightforward to users attempting to set a specific window.

**Expected behavior**
Users can edit all values without others changing automatically, even if that breaks aspect ratio.

**Screenshots**
tILdCPWS9A

### LINKED ISSUE #1019: Graph options error state should not prevent flyout closure
**Describe the bug**
When there is an error in graph options, the flyout is not light-dismissible.

**Expected behavior**
Flyout should always be dismissible. If in error, we should revert to previous good values (this is the behavior today, if you click outside of Calculator window).

**Screenshots**
image

joseartrivera:
- Set ForceProportionalAxes to false
- Remove not being able to close flyout when there is an error

- Manual tests

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -472,7 +472,7 @@
             </Grid.Resources>
             <graphControl:Grapher Name="GraphingControl"
                                   Style="{ThemeResource ThemedGrapherStyle}"
-                                  ForceProportionalAxes="True"
+                                  ForceProportionalAxes="False"
                                   LosingFocus="GraphingControl_LosingFocus"
                                   LostFocus="GraphingControl_LostFocus"
                                   RequestedTheme="Light"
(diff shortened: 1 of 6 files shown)


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -472,7 +472,7 @@
             </Grid.Resources>
             <graphControl:Grapher Name="GraphingControl"
                                   Style="{ThemeResource ThemedGrapherStyle}"
-                                  ForceProportionalAxes="True"
+                                  ForceProportionalAxes="False"
```
sanderl: Do we want to remove the logic In GraphingSettingsViewModel.cpp, UpdateDisplayRange() where ForceProportionalAxes is used?

joseartrivera: removed!

PR #1054, opened 2020-02-27
Check for ctrl when navigating to handle AltGr

### LINKED ISSUE #1009: 'Alt' key on French keyboard has conflicts in graphing mode
**Describe the bug**
When using azerty (French) keyboard layout, the right "Alt" key is replaced with an "Alt Gr" key, which is used to type some special chars such as '\~', '#', etc. when combined with other keys, typically the digit keys. When typing an equation in the graphing window, if you type a special character with Alt Gr and a digit key, the character will be typed, but the command will also be interpreted as using an Access Key with Alt, causing the app to navigate to the corresponding page. For example, typing '~' with Alt Gr + 4 navigates to the Scientific page.

**Expected behavior**
Alt Gr should not be mistaken with Alt and should not trigger navigation.

EriWong:
## Fixes #1009 .

Detect Ctrl when handling accelerator shortcuts to avoid Alt Gr triggering navigation
- Checks for ctlr being pressed when handling nav bar shortcut

manually

### DIFF
--- a/src/Calculator/Common/KeyboardShortcutManager.cpp
+++ b/src/Calculator/Common/KeyboardShortcutManager.cpp
@@ -642,6 +642,13 @@ void KeyboardShortcutManager::OnAcceleratorKeyActivated(CoreDispatcher ^, Accele
             return;
         }
 
+        // Ctrl is pressed in addition to alt, this means Alt Gr is intended.  do not navigate.
+        if ((static_cast<short>(Window::Current->CoreWindow->GetKeyState(VirtualKey::Control)) & static_cast<short>(CoreVirtualKeyStates::Down))
+            == static_cast<short>(CoreVirtualKeyStates::Down))
+        {
+            return;
+        }
+
         const auto& lookupMap = GetCurrentKeyDictionary(static_cast<MyVirtualKey>(key), altPressed);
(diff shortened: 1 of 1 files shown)


src/Calculator/Common/KeyboardShortcutManager.cpp
```diff
@@ -642,6 +642,13 @@ void KeyboardShortcutManager::OnAcceleratorKeyActivated(CoreDispatcher ^, Accele
             return;
         }
 
+        // Ctrl is pressed in addition to alt, this means Alt Gr is intended.  do not navigate.
+        if ((static_cast<short>(Window::Current->CoreWindow->GetKeyState(VirtualKey::Control)) & static_cast<short>(CoreVirtualKeyStates::Down))
```
sanderl: Is there a way for us to detect the keyboard type and add that to the if condition?

rudyhuyn: Even if qwerty keyboards don't have a `Altr Gr` button, Ctrl and Alt will also do some special actions. A good example is Ctrl+Alt+C, it won't   copy but type `©`. We should not try to check keyboard layouts.

sanderl: That makes sense to me. I just wanted to make sure we weren't blocking user intent in other cases.

PR #1055, opened 2020-02-27
Persist variable settings after graph is plotted

### LINKED ISSUE #1017: Graphing mode equation variables reset if you modify equation list
**Describe the bug**
After entering an equation with variables, if you add/edit/remove an equation from the list, variable value + settings (min/max/step) reset.

**Expected behavior**
Any changes a user makes to variables should persist unless the variable is removed from use entirely.

**Screenshots**
S1FNtqGyZA

joseartrivera:
- Creates a model for the variable that stores the values after the graph is re-plotted
- Updated VariableViewModel and other code to contain the model

- Manual tests


src/Settings.XamlStyler
```diff
@@ -19,7 +19,8 @@
         "*:*, *",
         "PageSource, PageIndex, Offset, Color, TargetName, Property, Value, StartPoint, EndPoint",
         "mc:Ignorable, d:IsDataSource, d:LayoutOverrides, d:IsStaticText",
-		"IsEnabled, x:Load, Load"
+		"IsEnabled, x:Load, Load",
+        "Value, Maximum, Minimum"
```
joseartrivera: This is one of those rare instances where the ordering of the binding actually matters (the order of binding is determined by the literal order the properties are defined in XAML). This change puts Value above Maximum and Minimum, otherwise if Value was higher than Maximum in the UI prior to the underlying viewmodel change, it would get reset to be the Maximum value and trigger a two-way binding, setting the default int for Value to be 2 or 0 instead of 1.

src/GraphControl/GraphControl.vcxproj.filters
```diff
@@ -71,5 +74,7 @@
   <ItemGroup>
     <CopyFileToFolders Include="$(GraphingImplDll)" />
     <CopyFileToFolders Include="$(GraphingEngineDll)" />
+    <CopyFileToFolders Include="$(GraphingImplDll)" />
```
sanderl: nit: you can remove the extra copy lines

joseartrivera:
> Every time a variable is added, the whole control refreshes. Can we somehow prevent the UI from refreshing with each variable add?

Not part of this PR but I added a fix.

PR #1056, opened 2020-02-27
Update Analysis after variables change

### LINKED ISSUE #1031: Equation analysis doesn't update when variables change
**Describe the bug**
Equation analysis doesn't update when variables change. Need to update the variables on the graph when doing analysis.

**Steps To Reproduce**
1. Graph y=mx+b
2. Change b=2
3. Click analysis equation

**Expected behavior**
y-intercept should be 2

**Screenshots**
image

sanderl:
- Updates the variable values for the variables in the function being analyzed when the analysis button is clicked

- manually
-
-

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -199,6 +199,16 @@ namespace GraphControl
     {
         if (auto graph = GetGraph(equation))
         {
+            for (auto graphVar : graph->GetVariables())
+            {
+                auto key = ref new String(graphVar->GetVariableName().data());
+                if (Variables->HasKey(key))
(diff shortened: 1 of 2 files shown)


src/GraphControl/Control/Grapher.cpp
```diff
@@ -199,6 +199,16 @@ namespace GraphControl
     {
         if (auto graph = GetGraph(equation))
         {
+            for (auto graphVar : graph->GetVariables())
```
joseartrivera: You should modify SetGraphArgs(); to take a graph as a variable and then call that function here instead. Then update other calls to pass in m_graph so that we don't duplicate code.

PR #1060, opened 2020-02-28
Support other locales for decimal and list separators

### LINKED ISSUE #1052: Graphing mode needs to support commas as decimals for different locales
**Describe the bug**
Graphing mode needs to be able to use commas as decimals in locales where that is appropriate

joseartrivera:
- Updates our logic to account for locales that use a comma as a decimal and semi-colon as a list separator 

- Manual tests

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -518,7 +518,7 @@ namespace GraphControl
         else
         {
             m_solver->ParsingOptions().SetLocalizationType(::LocalizationType::DecimalPointAndListComma);
-            m_solver->FormatOptions().SetLocalizationType(::LocalizationType::DecimalCommaAndListSemicolon);
+            m_solver->FormatOptions().SetLocalizationType(::LocalizationType::DecimalPointAndListComma);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/GraphControl/Control/Grapher.cpp
```diff
@@ -500,6 +508,20 @@ namespace GraphControl
...
+    {
+        if (newValue)
+        {
+            m_solver->ParsingOptions().SetLocalizationType(::LocalizationType::DecimalCommaAndListSemicolon);
+            m_solver->FormatOptions().SetLocalizationType(::LocalizationType::DecimalCommaAndListSemicolon);
+        }
+        else
+        {
+            m_solver->ParsingOptions().SetLocalizationType(::LocalizationType::DecimalPointAndListComma);
+            m_solver->FormatOptions().SetLocalizationType(::LocalizationType::DecimalCommaAndListSemicolon);
```
mcooley: Is this supposed to be LocalizationType::DecimalPointAndListComma?

src/GraphingInterfaces/IMathSolver.h
```diff
@@ -50,6 +51,7 @@ namespace Graphing
 
         virtual void SetFormatType(FormatType type) = 0;
 		virtual void SetMathMLPrefix(const std::wstring& value) = 0;
```
mcooley: nit: fix weird indent while you're here?

src/GraphControl/Control/Grapher.cpp
joseartrivera: Good catch!

PR #1064, opened 2020-03-04
Support mouse gestures when active tracing is enabled

### LINKED ISSUE #989: Mouse gestures should be supported when active tracing is enabled

joseartrivera:
- Trace cursor should follow the mouse on the graph

- Manual testing


joseartrivera:
> This isn't a smooth experience. I'm using a touch pad, and I can move fast enough to make the mouse cursor show up instead of the tracing icon. I find the tracing cursor lags behind the mouse movement as well.

Thanks for confirming, I noticed it too on remote desktop but wasn't sure if it was just input lag. Looks like it's a bug with Active Tracing always triggering a render even when the not necessary. This is a problem for tracing in general, I included a fix with this PR.

PR #1065, opened 2020-03-04
Support regular textbox context menu functions in the EquationTextBox

### LINKED ISSUE #978: Equation input textbox should support cut/copy/paste in context menu
**Describe the bug**
Common text actions are missing from context menu.

**Expected behavior**
Cut/copy/paste are available.

image

**Screenshots**
image

joseartrivera:
Since [TextCommandBarFlyout](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/command-bar-flyout#command-bar-flyouts-for-text-controls) is not customizable and we have added custom menu items, we need to manually wire up all of the basic text commands.

Added these items that are normally present in the RichEditBox context menu: cut, copy, paste, undo, select all.

- Manual tests


sanderl:
The current implementation always shows all of the menus. Can we have it behave more like the default control where cut and copy are not available when there is no text selected and undo is not available when there is nothing to undo?

joseartrivera:
> The current implementation always shows all of the menus. Can we have it behave more like the default control where cut and copy are not available when there is no text selected and undo is not available when there is nothing to undo?

Undo already behaves that way, I added an extra check for cut/copy

PR #1073, opened 2020-03-06
Add reset button to settings flyout

### LINKED ISSUE #1069: Graph window options should have reset button in the flyout

joseartrivera:
- Add reset button to settings flyout

- Manual tests

### DIFF
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -71,6 +71,26 @@ void GraphingSettingsViewModel::InitRanges()
     m_dontUpdateDisplayRange = false;
 }
 
+void GraphingSettingsViewModel::ResetView()
+{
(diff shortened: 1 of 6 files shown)


src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
```diff
@@ -71,6 +71,26 @@ void GraphingSettingsViewModel::InitRanges()
...
+    if (m_Graph != nullptr)
+    {
+        m_Graph->ResetGrid();
+        InitRanges();
+        m_XMinError = false;
+        m_XMaxError = false;
+        m_YMinError = false;
+        m_YMaxError = false;
+
+        RaisePropertyChanged("XError");
```
joseartrivera: For some reason this controller is written in a weird way such that it handles it's own RaisePropertyChanged, we will need to rethink this and the GraphControl implementation, not sure why the GraphControl doesn't just call SetDisplayRanges when the X/Y properties are changed there.

PR #1075, opened 2020-03-07
Improve error handling by displaying an error message

### LINKED ISSUE #985: Improve error handling for equation input to help users understand what is wrong with their input

joseartrivera:
- Displays an error message when hovering over the error icon.

- Manual tests, currently some of the errors do not seem to get returned by the graph engine.


src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
     <value>Unable to calculate the range for this function.</value>
     <comment>Error displayed when Range is not returned from the analyzer.</comment>
   </data>
-  <data name="equationAnalysisBack.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
+  <data name="Overflow" xml:space="preserve">
```
joseartrivera: @grochocki would be nice to have you take a second pass at some of these strings if you have time!

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <value>Overflow, the number is too large.</value>
+    <comment>Error that occurs during graping when the number is too large. To see this error, assign a large number to variable a, then keep doing "a:=a*a" until it happens.</comment>
+  </data>
+  <data name="RequireRadiansMode" xml:space="preserve">
+    <value>Radians mode is required to graph this equation.</value>
+    <comment>Error that occurs during graping when radians is required.</comment>
+  </data>
+  <data name="TooComplexToSolve" xml:space="preserve">
+    <value>The equation is too complex to graph.</value>
```
sanderl: I think we should use function instead of equation. Here and below.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+  <data name="DecimalPointWithoutDigits" xml:space="preserve">
+    <value>Decimal point is missing digits.</value>
+    <comment></comment>
+  </data>
+  <data name="UnexpectedEndOfExpression" xml:space="preserve">
+    <value>Unexpected end of expression.</value>
+    <comment>Error that occurs during graping when the expression ends unexpectedly. Ex: 3-4*</comment>
+  </data>
+  <data name="UnexpectedToken" xml:space="preserve">
+    <value>Unexpected token in the expression.</value>
```
sanderl: I don't think most users would know how to fix the expression because I don't think they will know what a token is. Can we give an example of what might be the problem or give a different error message?

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+  <data name="InvalidNumberDigit" xml:space="preserve">
+    <value>The digit could not be resolved for the given base.</value>
+    <comment>Error that occurs during graping when trying to use bases incorrect. Ex: base(2,1020).</comment>
+  </data>
+  <data name="InvalidNumberBase" xml:space="preserve">
+    <value>The base must be greater than 2 or less than 36.</value>
+    <comment>Error that occurs during graping when the base is out of range.</comment>
+  </data>
+  <data name="InvalidVariableSpecification" xml:space="preserve">
+    <value>A function in the equation requres a variable in a certain argument position.</value>
```
sanderl: I find this error message confusing. Maybe something like "Required variable is missing."

src/Calculator/Resources/en-US/Resources.resw
sanderl: does it make sense to give examples of the problem in the error message itself not just in the comment?

sanderl:
I found a couple issues where the error is not showing. There could be more since I only did a few tests.

Testing results:
y= ; results in the appropriate error
x-x=3 ; no error shown, doesn't graph
ix < 3; results in appropriate error
i < 3; results in appropriate error
xi builds up to a complex character; results in expected error
sin(x=y) ; results in appropriate error
1 % 2.5 ; no error shown, doesn't graph

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
     <value>Unable to calculate the range for this function.</value>
     <comment>Error displayed when Range is not returned from the analyzer.</comment>
   </data>
-  <data name="equationAnalysisBack.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
+  <data name="Overflow" xml:space="preserve">
+    <value>Overflow, the number is too large.</value>
```
grochocki: Suggested Change: "Overflow (the number is too large)"

src/Calculator/Resources/en-US/Resources.resw
grochocki: I don't think we want to provide examples in the error message. It starts to become pretty verbose. Let's start with the simple error messages, and add more detail as necessary based on user feedback.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+  <data name="FactorialCannotPerformOnLargeNumber" xml:space="preserve">
+    <value>The factorial function has an argument that is too large to graph.</value>
+    <comment>Error that occurs during graping when a factorial has a large n</comment>
+  </data>
+  <data name="ModuloCannotPerformOnFloat" xml:space="preserve">
+    <value>Modulo can only be used with whole numbers.</value>
+    <comment>Error that occurs during graping when modulo is used with a float.</comment>
+  </data>
+  <data name="EquationTooComplexToSolve" xml:space="preserve">
+    <value>The equation is too complex to graph.</value>
```
grochocki: Is this error duplicate of above? Why two?

src/Calculator/Resources/en-US/Resources.resw
grochocki: Also, I think we may want to say "This function..." here and below.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+  <data name="ParenthesisMismatch" xml:space="preserve">
+    <value>The equation is missing an opening paranthesis.</value>
+    <comment>Error that occurs during graping when the equation is missing a (</comment>
+  </data>
+  <data name="UnmatchedParenthesis" xml:space="preserve">
+    <value>The equation is missing a closing paranthesis.</value>
+    <comment>Error that occurs during graping when the equation is missing a )</comment>
+  </data>
+  <data name="TooManyDecimalPoints" xml:space="preserve">
+    <value>Too many decimal points in a number.</value>
```
grochocki: "There are too many decimal points in a number"

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when a number has too many decimals. Ex: 1.2.3</comment>
+  </data>
+  <data name="DecimalPointWithoutDigits" xml:space="preserve">
+    <value>Decimal point is missing digits.</value>
```
grochocki: "A decimal point is missing digits"

src/Calculator/Resources/en-US/Resources.resw
grochocki: How about "Unexpected characters in the expression"

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+  <data name="UnexpectedEndOfExpression" xml:space="preserve">
+    <value>Unexpected end of expression.</value>
+    <comment>Error that occurs during graping when the expression ends unexpectedly. Ex: 3-4*</comment>
+  </data>
+  <data name="UnexpectedToken" xml:space="preserve">
+    <value>Unexpected token in the expression.</value>
+    <comment>Error that occurs during graping when there is an unexpected token.</comment>
+  </data>
+  <data name="InvalidToken" xml:space="preserve">
+    <value>Invalid token in the expression.</value>
```
grochocki: Just flagging to fix the same way that we fix above

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when there is an invalid token.</comment>
+  </data>
+  <data name="TooManyEquals" xml:space="preserve">
+    <value>Too many equal signs.</value>
```
grochocki: "There are too many equal signs"

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when there are too many equals.</comment>
+  </data>
+  <data name="EqualWithoutGraphVariable" xml:space="preserve">
+    <value>Missing x or y variable.</value>
```
grochocki: "The function must contain at least one x or y variable."

But is this entirely true? I can plot just "6", and that works because the "y=" is implicit. When do we run into this error?

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when the equation is missing x or y.</comment>
+  </data>
+  <data name="InvalidEquationSyntax" xml:space="preserve">
+    <value>Invalid equation syntax.</value>
```
grochocki: I don't think we should use the word "syntax". Perhaps: "Invalid expression"

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <data name="EqualWithoutEquation" xml:space="preserve">
+    <value>Equal was used without an equation.</value>
+    <comment>Error that occurs during graping when equal is used without an equation. Ex: sin(x=y)</comment>
+  </data>
+  <data name="ExpectParenthesisAfterFunctionName" xml:space="preserve">
+    <value>Paranthesis missing after function name.</value>
+    <comment>Error that occurs during graping when paranthesis are missing after a function.</comment>
+    </data>
+  <data name="IncorrectNumParameter" xml:space="preserve">
+    <value>A function has the incorrect number of parameters.</value>
```
grochocki: We should reconcile how to refer to part of a function, since we are calling the entire thing a function above.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when a function has the wrong number of parameters</comment>
+  </data>
+  <data name="InvalidVariableNameFormat" xml:space="preserve">
+    <value>The variable name is invalid.</value>
```
grochocki: Perhaps: "A variable name is invalid"

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+  <data name="GeneralError" xml:space="preserve">
+    <value>The equation could not be graphed.</value>
+    <comment>General error that occurs during graping.</comment>
+  </data>
+  <data name="InvalidNumberDigit" xml:space="preserve">
+    <value>The digit could not be resolved for the given base.</value>
+    <comment>Error that occurs during graping when trying to use bases incorrect. Ex: base(2,1020).</comment>
+  </data>
+  <data name="InvalidNumberBase" xml:space="preserve">
+    <value>The base must be greater than 2 or less than 36.</value>
```
grochocki: and

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when the base is out of range.</comment>
+  </data>
+  <data name="InvalidVariableSpecification" xml:space="preserve">
+    <value>A function in the equation requres a variable in a certain argument position.</value>
+    <comment>Error that occurs during graping when a function requires a variable in a particular position. Ex: 2nd argument of deriv.</comment>
+  </data>
+  <data name="ExpectingLogicalOperands" xml:space="preserve">
+    <value>Equation is mixing logical and scalar operands.</value>
```
grochocki: 😱

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when operands are mixed. Such as true and 1.</comment>
+  </data>
+  <data name="CannotUseIndexVarInOpLimits" xml:space="preserve">
+    <value>x or y cannot be used in the upper or lower limits.</value>
+    <comment>Error that occurs during graping when x or y is used in integral upper limits.</comment>
+  </data>
+  <data name="CannotUseIndexVarInLimPoint" xml:space="preserve">
+    <value>x or y variable cannot be used in the limit point.</value>
```
grochocki: remove "variable" for consistency with error immediately before

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,167 @@
...
+    <comment>Error that occurs during graping when x or y is used in the limit point.</comment>
+  </data>
+  <data name="CannotUseComplexInfinityInReal" xml:space="preserve">
+    <value>Cannot use complex infinity.</value>
+    <comment>Error that occurs during graping when complex infinity is used</comment>
+  </data>
+  <data name="CannotUseIInInequalitySolving" xml:space="preserve">
+    <value>Cannot use complex numbers in inequalities.</value>
+    <comment>Error that occurs during graping when complex numbers are used in inequalities.</comment>
```
grochocki: I did not notice any errors that were longer than one sentence. Can we remove all periods?

src/Calculator/Resources/en-US/Resources.resw
joseartrivera: Something like 6=1 would throw it iirc

src/Calculator/Resources/en-US/Resources.resw
joseartrivera: Good point, maybe "Mathematical operation"?

src/Calculator/Resources/en-US/Resources.resw
sanderl: Maybe something more generic like "Invalid function" since 6=1 and 6=A are not valid functions.

joseartrivera: I think when we can be more specific about an error message we should do so, it makes it clearer to the user what exactly is wrong with their input.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,163 @@
...
+    <value>The equation is missing a closing paranthesis</value>
+    <comment>Error that occurs during graping when the equation is missing a )</comment>
+  </data>
+  <data name="TooManyDecimalPoints" xml:space="preserve">
+    <value>There are too many decimal points in a number</value>
+    <comment>Error that occurs during graping when a number has too many decimals. Ex: 1.2.3</comment>
+  </data>
+  <data name="DecimalPointWithoutDigits" xml:space="preserve">
+    <value>A decimal point is missing digits</value>
+    <comment></comment>
```
sanderl: Comment missing

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,163 @@
     <value>Unable to calculate the range for this function.</value>
     <comment>Error displayed when Range is not returned from the analyzer.</comment>
   </data>
-  <data name="equationAnalysisBack.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
+  <data name="Overflow" xml:space="preserve">
+    <value>Overflow (the number is too large)</value>
+    <comment>Error that occurs during graping when the number is too large. To see this error, assign a large number to variable a, then keep doing "a:=a*a" until it happens.</comment>
```
sanderl: spelling: should be graphing, here and below

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,163 @@
...
+    <value>The expression is empty</value>
+    <comment>Error that occurs during graping when the expression is empty</comment>
+  </data>
+    <data name="EqualWithoutEquation" xml:space="preserve">
+    <value>Equal was used without an equation</value>
+    <comment>Error that occurs during graping when equal is used without an equation. Ex: sin(x=y)</comment>
+  </data>
+  <data name="ExpectParenthesisAfterFunctionName" xml:space="preserve">
+    <value>Paranthesis missing after function name</value>
+    <comment>Error that occurs during graping when paranthesis are missing after a function.</comment>
```
ItzLevvie:
```suggestion
    <comment>Error that occurs during graping when parenthesis are missing after a function.</comment>
```

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,163 @@
...
+    <data name="EmptyExpression" xml:space="preserve">
+    <value>The expression is empty</value>
+    <comment>Error that occurs during graping when the expression is empty</comment>
+  </data>
+    <data name="EqualWithoutEquation" xml:space="preserve">
+    <value>Equal was used without an equation</value>
+    <comment>Error that occurs during graping when equal is used without an equation. Ex: sin(x=y)</comment>
+  </data>
+  <data name="ExpectParenthesisAfterFunctionName" xml:space="preserve">
+    <value>Paranthesis missing after function name</value>
```
ItzLevvie:
```suggestion
    <value>Parenthesis missing after function name</value>
```

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4106,7 +4106,163 @@
...
+    <data name="OutOfDomain" xml:space="preserve">
+    <value>Equation is out of domain</value>
+    <comment>Error that occurs during graping when the equation is out of domain.</comment>
+  </data>
+    <data name="GE_NotSupported" xml:space="preserve">
+    <value>Graphing this equation is not supported</value>
+    <comment>Error that occurs during graping when the equation is not supported.</comment>
+  </data>
+  <data name="ParenthesisMismatch" xml:space="preserve">
+    <value>The equation is missing an opening paranthesis</value>
```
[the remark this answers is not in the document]
joseartrivera: Good catch!

PR #1027, opened 2020-02-12
Add dual-screen support to Calculator

rudyhuyn:
## Fixes #1029

Support Dual-landscape and Dual-portrait modes.

image
image
image
image

### Description of the changes:
- Add new StateTrigger detecting the different view modes
- Add a fork of TwoPaneView in C++CX for 2 reasons:
   - The fix for C++ apps (crash when navigating) isn't available for the moment
   - Add `Pane1MinLength`, `Pane1MaxLength`, `Pane2MinLength`, `Pane2MaxLength` properties
- Merge Visual State Group of Unit Converter, causing some issues due to conflicts
- Add `DeviceFamilyHelper`

### Other changes
Visual states of `Calculator.xaml` had some collisions, the states related to AOT (https://github.com/microsoft/calculator/blob/master/src/Calculator/Views/Calculator.xaml#L521) and the states related to CalculationResult (https://github.com/microsoft/calculator/blob/master/src/Calculator/Views/Calculator.xaml#L544) were both triggered by the same type of trigger (window resizing), without checking if the app was AOT or not, and were modifying the same properties which caused some layouts issues.

In order to solve this problem, which also affects dual screens, we have replaced the StateTriggers used in AOT visual states with `ApplicationViewModeTrigger` in order to activate them only when the application is actually AOT.

### How changes were validated:
- manually with 10X 10.0.19563

PR #1078, opened 2020-03-09
Patch TitleBar to be correctly displayed on Windows 10X Preview

### LINKED ISSUE #1077: The custom titlebar isn't correctly displayed on Windows 10X
**Describe the bug**
Windows 10X preview has currently an issue with custom titlebars (`CoreApplicationViewTitleBar` returns `Height = 0px` and `IsVisible = Collapsed` instead of `32px` and `True`)

**Expected behavior**

A future preview of Windows 10X will fix this problem, but in the meantime, we should display the title bar correctly on Windows 10X.

**Requested Assignment**

rudyhuyn:
- Force the visibility (visible) and the height (32px) of the Titlebar when running on Windows 10X
- Update `MaxVersionTested` to get access to the real DeviceFamily

- Tested with 10X Emulators


rudyhuyn:
Merged with #1027

PR #1079, opened 2020-03-09
Update styling of graph settings textbox and add shadow

### LINKED ISSUE #986: Add shadow elements to graphing calculator wherever we have custom popups
**Describe the bug**
Some custom popups do not have shadows.

**Expected behavior**
Audit all custom popups added for graphing mode and ensure shadows are enabled for consistency with platform.

**Screenshots**
image

### LINKED ISSUE #1018: Graph window options x/y min/max textbox styling does not match variable min/max styling
**Describe the bug**
Graph window options x/y min/max textbox styling does not match variable min/max styling.

**Expected behavior**
Styling should be consistent with variable editing. As we add more graph options, we will also benefit from the reduced visual footprint of this styling.
image

**Screenshots**
image

joseartrivera:
- Updates text boxes in settings flyout to be consistent with variable text boxes
- Adds shadow to trace value pop up

- Manual tests


src/Calculator/App.xaml
```diff
@@ -85,6 +85,22 @@
                     <SolidColorBrush x:Key="EquationBrush13" Color="#FF8E562E"/>
                     <SolidColorBrush x:Key="EquationBrush14" Color="#FF000000"/>
 
+                    <SolidColorBrush x:Key="GraphSettingsErrorBackgroundBrush" Color="#33EB5757"/>
+                    <SolidColorBrush x:Key="GraphSettingsErrorBorderBrush" Color="#FFEB5757"/>
+                    <SolidColorBrush x:Key="TextControlBackgroundFocused" Color="Transparent"/>
```
rudyhuyn: These brushes override system brushes and will modify all textboxes in the app. Their scopes were limited when they were in `EquationInputArea` but not here.

src/Calculator/App.xaml
```diff
@@ -197,6 +228,29 @@
                     <SolidColorBrush x:Key="EquationBrush2" Color="{ThemeResource SystemColorHighlightColor}"/>
                     <SolidColorBrush x:Key="EquationBrush3" Color="{ThemeResource SystemColorHotlightColor}"/>
                     <SolidColorBrush x:Key="EquationBrush4" Color="{ThemeResource SystemColorWindowTextColor}"/>
+
+                    <SolidColorBrush x:Key="GraphSettingsErrorBackgroundBrush" Color="{ThemeResource SystemColorButtonFaceColor}"/>
+                    <SolidColorBrush x:Key="GraphSettingsErrorBorderBrush" Color="Red"/>
+                    <SolidColorBrush x:Key="TextControlBackgroundFocused" Color="{StaticResource SystemColorButtonFaceColor}"/>
+                    <SolidColorBrush x:Key="TextControlBackgroundPointerOver" Color="{StaticResource SystemColorButtonFaceColor}"/>
+                    <SolidColorBrush x:Key="TextControlForegroundFocused" Color="{StaticResource SystemColorButtonTextColor}"/>
+                    <SolidColorBrush x:Key="VariablesBackgroundBrush" Color="{StaticResource SystemColorBackgroundColor}"/>
```
rudyhuyn: Not related to your PR, but this brush is not used, let's remove it

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -1,6 +1,7 @@
 <UserControl x:Class="CalculatorApp.GraphingCalculator"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:Windows10version1903="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 8)"
```
rudyhuyn: We should not associate UniversalApiContract and windows versions and be consistent with other namespaces.

`xmlns:contract8Present="..."`

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -87,6 +87,8 @@ GraphingCalculator::GraphingCalculator()
     virtualKey->Key = (VirtualKey)187; // OemAdd key
     virtualKey->Modifiers = VirtualKeyModifiers::Control;
     ZoomInButton->KeyboardAccelerators->Append(virtualKey);
+
+    SharedShadow->Receivers->Append(GraphingControl);
```
rudyhuyn: This code will crash on older version of windows

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml.cpp
```diff
@@ -37,11 +37,11 @@ Style ^ GraphingSettings::SelectTextBoxStyle(bool incorrectRange, bool error)
 {
     if (incorrectRange || error)
     {
-        return static_cast<::Style ^>(this->Resources->Lookup(L"ErrorTextBoxStyle"));
+        return static_cast<::Style ^>(Application::Current->Resources->Lookup(L"VariableTextErrorBoxStyle"));
     }
     else
     {
-        return nullptr;
+        return static_cast<::Style ^>(Application::Current->Resources->Lookup(L"ThemedVariableTextBoxStyle"));
```
rudyhuyn: The textbox won't be updated when the user will modify the app theme. 

We should try something like:
`<Style x:Key="GraphSettingsVariableTextBoxStyle" BasedOn="{ThemeResource ThemedVariableTextBoxStyle}" />` (in GraphingSettings.xaml)
and write
`return static_cast<::Style ^>(this->Resources->Lookup(L"GraphSettingsVariableTextBoxStyle"));`

src/Calculator/App.xaml
joseartrivera: Good catch!

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: This code shouldn’t run on older versions since it’s Graphing Calc but it’s probably good to add an explicit check anyway.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
joseartrivera: Is there a reason we shouldn’t or is it just for consistency? It is what VS auto-inserted for me, should we send feedback for them to change that?

src/Calculator/App.xaml
```diff
@@ -85,6 +85,22 @@
                     <SolidColorBrush x:Key="EquationBrush13" Color="#FF8E562E"/>
                     <SolidColorBrush x:Key="EquationBrush14" Color="#FF000000"/>
 
+                    <SolidColorBrush x:Key="GraphSettingsErrorBackgroundBrush" Color="#33EB5757"/>
```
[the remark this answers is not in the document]
joseartrivera: They are already removed!

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml.cpp
joseartrivera: Hmm, are you sure? It seems to be working for me

rudyhuyn: it works because GraphSettings is automatically closed when the window loses the focus. But if we want this popup to persist or when the theme is changed automatically (for example one of the many apps switching your theme automatically at night), the style used here won't be updated.

joseartrivera: I made the change, but I couldn't use ThemeResource because I got this error:
Style BasedOn property must be a Style, not 'ThemeResourceExtension' object

Still works for me, but unsure if the automatic theme changing will function correctly.

src/Calculator/App.xaml
```diff
@@ -1766,6 +1810,29 @@
...
+                <Setter Property="Padding" Value="2,6,2,2"/>
+                <Setter Property="TextAlignment" Value="Center"/>
+                <Setter Property="Background" Value="Transparent"/>
+                <Setter Property="BorderThickness" Value="0,0,0,1"/>
+                <Setter Property="FontSize" Value="14"/>
+                <Setter Property="InputScope" Value="Number"/>
+                <Setter Property="MinWidth" Value="30"/>
+            </Style>
+
+            <Style x:Key="VariableTextErrorBoxStyle" TargetType="TextBox">
```
rudyhuyn, line 1825: Except the background and the borderbrush, this style is exactly the same than VariableTextBoxStyle.

What do you think about replacing it with:
```
<Style x:Key="VariableTextErrorBoxStyle"
                           BasedOn="{StaticResource VariableTextBoxStyle}"
                           TargetType="TextBox">
    <Setter Property="BorderBrush" Value="{ThemeResource GraphSettingsErrorBorderBrush}"/>
    <Setter Property="Background" Value="{ThemeResource GraphSettingsErrorBackgroundBrush}"/>
</Style>
```
?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -619,7 +624,7 @@ void CalculatorApp::GraphingCalculator::AddTracePointerShadow()
     dropShadow->Mask = CursorPath->GetAlphaMask();
 
     auto shadowSpriteVisual = compositor->CreateSpriteVisual();
-    shadowSpriteVisual->Size = ::Numerics::float2(static_cast<float>(CursorPath->ActualWidth), static_cast<float>(CursorPath->ActualHeight));
+    shadowSpriteVisual->Size = ::Numerics::float2(18, 18);
```
rudyhuyn, line 627: Why this change?

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -223,8 +226,7 @@
                        AutomationProperties.HeadingLevel="Level2"/>
 
             <StackPanel Orientation="Vertical">
-                <RadioButton x:Uid="AlwaysLightTheme"
-                             IsChecked="{x:Bind IsMatchAppTheme, Mode=TwoWay, Converter={StaticResource BooleanNegationConverter}}"/>
+                <RadioButton x:Uid="AlwaysLightTheme" IsChecked="{x:Bind IsMatchAppTheme, Mode=TwoWay, Converter={StaticResource BooleanNegationConverter}}"/>
```
rudyhuyn, line 229: Can we use more explicit name? `ThemeOptionAlwaysLight`/ `ThemeOptionMatchApp`

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: ActualWidth and ActualHeight are usually 0 at the time this code runs

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
joseartrivera: Hmm agree, but not part of this PR, this is just XAML styler rearranging things

PR #1081, opened 2020-03-10
Don't display the history button when the panel disappears

### LINKED ISSUE #1080: History button can be visible in some conditions when in AOT mode.
image

**Describe the bug**
In some conditions, the history button can still be visible when entering compact-overlay mode.

**Requested Assignment**

rudyhuyn:
- Don't force the visibility of the History button when the history panel is closed

- manually

### DIFF
--- a/src/Calculator/Views/Calculator.xaml
+++ b/src/Calculator/Views/Calculator.xaml
@@ -659,8 +659,7 @@
                     Content="&#xe81c;"
                     ExitDisplayModeOnAccessKeyInvoked="False"
                     TabIndex="2"
-                    Visibility="{x:Bind local:Calculator.ShouldDisplayHistoryButton(Model.IsAlwaysOnTop, Model.IsProgrammer, DockPanel.Visibility), Mode=OneWay}"
-                    IsEnabled="{x:Bind Model.IsAlwaysOnTop, Converter={StaticResource BooleanNegationConverter}, Mode=OneWay}">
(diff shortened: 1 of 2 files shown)


src/Calculator/Views/Calculator.xaml
```diff
@@ -659,8 +659,7 @@
                     Content="&#xe81c;"
                     ExitDisplayModeOnAccessKeyInvoked="False"
                     TabIndex="2"
-                    Visibility="{x:Bind local:Calculator.ShouldDisplayHistoryButton(Model.IsAlwaysOnTop, Model.IsProgrammer, DockPanel.Visibility), Mode=OneWay}"
-                    IsEnabled="{x:Bind Model.IsAlwaysOnTop, Converter={StaticResource BooleanNegationConverter}, Mode=OneWay}">
```
joseartrivera: I am guessing the IsEnabled was originally added to disable the hotkey from being accessed when AoT on top. We should test this and add a comment to clarify instead of removing it if it turns out to be true.

rudyhuyn: You are right, KeyboardShortcutManager doesn't check if a button is visible or not (on purpose?). The command associated to the button already checks if the dockpanel is visible or not, but I still will revert this code because not part of the original problem

joseartrivera: I believe it is on purpose so that some shortcuts are still accessible even when they are hidden behind the shift toggle.

PR #1087, opened 2020-03-15
Change service URL to expression body instead of property

pi1024e:
- The change to an expression body best shows the ServiceURL as being an expression, instead of an object with no get but a set.

- Automated tests
- Manual Tests
- Compiler refactor


mcooley:
Thanks for taking a look at this. There's no difference in the behavior of the generated code, and I don't think the new style is significantly better than the old style, so we'll keep the existing code as-is.

PR #1088, opened 2020-03-15
Merge initializers and assignments

pi1024e:
- In some functions, there are outer scope variables that are initialized but have no value, and only obtain a value inside an if statement or inner scope, and others that are initialized but have no value until the function is nearly over. These changes put the variables in the innermost possible scope while ensuring they go initialized even if the codebase evolves. 
- Made some variables const, those variables being mostly the ones in which the initialization and assignments were conjoined.

- Compiler Refactorings
- Automated Unit tests
- Manual tests to ensure no side effects happen

### DIFF
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -37,14 +37,14 @@ namespace
             IDC_MOD,3, IDC_DIV,3, IDC_MUL,3,
             IDC_PWR,4, IDC_ROOT,4, IDC_LOGBASEX,4 };
 
-        for (unsigned int iPrec = 0; iPrec < size(rgbPrec); iPrec +=2 )
+        for (unsigned int iPrec = 0; iPrec < size(rgbPrec); iPrec += 2)
         {
             if (nopCode == rgbPrec[iPrec]) 
             {
                 return rgbPrec[iPrec + 1];
             }
         }
-        return rgbPrec[1];
+        return 0;
     }
 }
 
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -37,16 +37,14 @@ namespace
             IDC_MOD,3, IDC_DIV,3, IDC_MUL,3,
             IDC_PWR,4, IDC_ROOT,4, IDC_LOGBASEX,4 };
 
-        unsigned int iPrec = 0;
-        while ((iPrec < size(rgbPrec)) && (nopCode != rgbPrec[iPrec]))
+        for (auto iPrec = 0; iPrec < size(rgbPrec); iPrec +=2 )
         {
-            iPrec += 2;
+            if (nopCode == rgbPrec[iPrec]) 
+            {
+                return rgbPrec[iPrec + 1];
+            }
         }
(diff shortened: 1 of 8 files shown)


src/CalcManager/Ratpack/fact.cpp
```diff
@@ -69,10 +69,9 @@ void _gamma(PRAT* pn, uint32_t radix, int32_t precision)
     PRAT mpy = nullptr;
     PRAT ratprec = nullptr;
     PRAT ratRadix = nullptr;
-    int32_t oldprec;
 
     // Set up constants and initial conditions
-    oldprec = precision;
+    const int32_t oldprec = precision;
     ratprec = i32torat(oldprec);
```
mcooley: I think you can make a similar refactor here with some of the PRAT variables.

src/CalcManager/CalculatorHistory.cpp
```diff
@@ -53,7 +52,7 @@ unsigned int CalculatorHistory::AddToHistory(
     // in the history doesn't get broken for RTL languages
     spHistoryItem->historyItemVector.expression = L'\u202d' + generatedExpression + L'\u202c';
     spHistoryItem->historyItemVector.result = wstring(result);
-    addedIndex = AddItem(spHistoryItem);
+    unsigned int addedIndex = AddItem(spHistoryItem);
```
rudyhuyn: Thank you for this one! We can even remove `addedIndex` and directly return addIndex

src/CalcManager/Ratpack/exp.cpp
```diff
@@ -448,10 +441,9 @@ void powratcomp(_Inout_ PRAT* px, _In_ PRAT y, uint32_t radix, int32_t precision
             {
                 // If power is an integer let ratpowi32 deal with it.
                 PRAT iy = nullptr;
-                int32_t inty;
                 DUPRAT(iy, y);
                 subrat(&iy, podd, precision);
-                inty = rattoi32(iy, radix, precision);
+                int32_t inty = rattoi32(iy, radix, precision);
```
rudyhuyn: It won't change the generated binary, but easier to read! 👍

src/CalcManager/CEngine/scicomm.cpp
```diff
@@ -214,7 +212,7 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
             // Here * is m_nPrevOpCode, m_currentVal is 2  (by 1*2), m_nLastCom is +, m_nOpCode is ^
             if (m_fPrecedence && 0 != m_nPrevOpCode)
             {
-                nPrev = NPrecedenceOfOp(m_nPrevOpCode);
+                int nPrev = NPrecedenceOfOp(m_nPrevOpCode);
                 nx = NPrecedenceOfOp(m_nLastCom);
```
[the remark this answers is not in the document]
pi1024e: It is a bit difficult since nx is used in the switch statement that comes later, which is not nested in any other statement, but I will see if that is possible without any drastic changes to the function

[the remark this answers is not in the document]
pi1024e: Update: I managed to do it

pi1024e:
@mcooley How do you feel about these changes?

PR #1089, opened 2020-03-16
Use static_cast instead of C-style cast in UnitConverter.cpp

pi1024e:
# Description of the changes:
- In UnitConverter, for the conversion to vector, the delimiter size is cast to an int using the C style casting, so this was changed to the c++ style.

- ad hoc testing
- automated

On a side note, I have to ask why even cast at all, considering the variables in the equation are all size_t anyway. Nevertheless, this is the only case in the entire solution in which a c style case is used, and it is in a cpp file, so I changed it to that. Please let me know if the cast is even necessary should I be able to improve this code further by removing the case.

Thank you!


rudyhuyn:
Thank you for your contribution, we really appreciate your hard work!

No problem with this PR but all PR must be associated with an open question, so feel free to create an issue before submitting a PR in the future! Here are our [contribution guidelines](https://github.com/microsoft/calculator/blob/master/CONTRIBUTING.md)

Thanks

Rudy

PR #1090, opened 2020-03-16
Unresolve call to virtual function in constructor by using the namespace.

pi1024e:
- There is a call to a virtual function in the constructor, which will be resolved statically. Because we do not want that, I put the namespace name so the compiler does not do that.

- Manual
- Automated


rudyhuyn:
Thanks!

I will unfortunately reject this change. `UnitConverter` isn't sealed and even if no classes are derived from _UnitConverter_ in this project, we aren't sure that this will not be the case of some third party developers who use CalcManager in their own project. Thanks again for your contributions!

PR #1106, opened 2020-03-23
Dark Theme For Graph Control

### LINKED ISSUE #997: Graph display should support dark theme with theme option in graph options
**Describe the bug**
Graph currently does not support dark theme.

**Expected behavior**
* Graph remains light theme by default (even when the rest of the app is dark theme).
* Graph options allows users to specify graph theme. This setting should be persistent. between sessions.
* Colors in style picker all need dark theme adjustments
* We may need to explore the button styling for graph buttons (e.g., zoom in) to ensure we meet contrast requirements.

**Design Details**

sanderl:
- Adds the them setting options
- Enables Dark theme on the graph control
- Adds diagnostics for the theme settings
- updated the internal build version and the calculator app version

- manually using both the always light and match app options
- manually tested changing the app theme in settings updates the graph control theme per the setting in graph settings

### DIFF
--- a/build/pipelines/azure-pipelines.release.yaml
+++ b/build/pipelines/azure-pipelines.release.yaml
@@ -9,8 +9,8 @@ pr: none
 
 variables:
   versionMajor: 10
-  versionMinor: 2001
-  versionBuild: $[counter('10.2001.*', 0)]
+  versionMinor: 2002
+  versionBuild: $[counter('10.2002.*', 0)]
   versionPatch: 0
 
 name: '$(versionMajor).$(versionMinor).$(versionBuild).$(versionPatch)'
--- a/build/pipelines/azure-pipelines.release.yaml
+++ b/build/pipelines/azure-pipelines.release.yaml
@@ -9,8 +9,8 @@ pr: none
 
 variables:
   versionMajor: 10
-  versionMinor: 2002
-  versionBuild: $[counter('10.2002.*', 0)]
+  versionMinor: 2001
+  versionBuild: $[counter('10.2001.*', 0)]
   versionPatch: 0
 
 name: '$(versionMajor).$(versionMinor).$(versionBuild).$(versionPatch)'
--- a/src/CalcViewModel/Common/TraceLogger.cpp
+++ b/src/CalcViewModel/Common/TraceLogger.cpp
@@ -41,6 +41,7 @@ namespace CalculatorApp
     constexpr auto EVENT_NAME_VARIABLE_CHANGED = L"VariableChanged";
     constexpr auto EVENT_NAME_VARIABLE_SETTING_CHANGED = L"VariableSettingChanged";
     constexpr auto EVENT_NAME_GRAPH_SETTINGS_CHANGED = L"GraphSettingsChanged";
+    constexpr auto EVENT_NAME_GRAPH_THEME = L"GraphTheme";
 
     constexpr auto EVENT_NAME_EXCEPTION = L"Exception";
 
@@ -303,12 +304,21 @@ namespace CalculatorApp
         TraceLoggingCommon::GetInstance()->LogLevel2Event(StringReference(EVENT_NAME_VARIABLE_SETTING_CHANGED), fields);
     }
 
-    void TraceLogger::LogGraphSettingsChanged(GraphSettingsType settingType)
+    void TraceLogger::LogGraphSettingsChanged(GraphSettingsType settingType, String ^ settingValue)
     {
         auto fields = ref new LoggingFields();
         fields->AddString(StringReference(CALC_MODE), StringReference(GRAPHING_MODE));
         fields->AddInt16(L"SettingType", static_cast<int16>(settingType));
+        fields->AddString(L"SettingValue", settingValue);
 
         TraceLoggingCommon::GetInstance()->LogLevel2Event(StringReference(EVENT_NAME_GRAPH_SETTINGS_CHANGED), fields);
+    }
+    void TraceLogger::LogGraphTheme(Platform::String ^ graphTheme)
+    {
+        auto fields = ref new LoggingFields();
+        fields->AddString(StringReference(CALC_MODE), StringReference(GRAPHING_MODE));
+        fields->AddString(L"GraphTheme", graphTheme);
+
+        TraceLoggingCommon::GetInstance()->LogLevel2Event(StringReference(EVENT_NAME_GRAPH_THEME), fields);
     }
 }
--- a/src/CalcViewModel/Common/TraceLogger.cpp
+++ b/src/CalcViewModel/Common/TraceLogger.cpp
@@ -313,7 +313,8 @@ namespace CalculatorApp
 
         TraceLoggingCommon::GetInstance()->LogLevel2Event(StringReference(EVENT_NAME_GRAPH_SETTINGS_CHANGED), fields);
     }
-    void TraceLogger::LogGraphTheme(Platform::String ^ graphTheme)
+
+    void TraceLogger::LogGraphTheme(String ^ graphTheme)
     {
         auto fields = ref new LoggingFields();
         fields->AddString(StringReference(CALC_MODE), StringReference(GRAPHING_MODE));
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -10,6 +10,7 @@ using namespace CalcManager::NumberFormattingUtils;
 using namespace GraphControl;
 using namespace std;
 using namespace Platform;
+using namespace Windows::Storage;
 using namespace Windows::UI::Xaml;
 
 GraphingSettingsViewModel::GraphingSettingsViewModel()
@@ -115,7 +116,7 @@ void GraphingSettingsViewModel::UpdateDisplayRange()
 
     m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue);
 
-    TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Grid);
+    TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Grid, L"");
 }
 
 bool GraphingSettingsViewModel::HasError()
@@ -127,16 +128,12 @@ void GraphingSettingsViewModel::OnPropertyChanged(Platform::String ^ propertyNam
 {
     if (propertyName == L"IsAlwaysLightTheme" && IsAlwaysLightTheme)
     {
-        Graph->AxesColor = static_cast<Windows::UI::Color>(Application::Current->Resources->Lookup(L"LightThemeAxisColor"));
-        Graph->GraphBackground = static_cast<Windows::UI::Color>(Application::Current->Resources->Lookup(L"LightThemeGraphBackgroundColor"));
-        Graph->GridLinesColor = static_cast<Windows::UI::Color>(Application::Current->Resources->Lookup(L"LightThemeGridLinesColor"));
-        Graph->GraphTheme = L"AlwaysLight";
+        GraphThemeSettingChanged(this, propertyName);
+        TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Theme, propertyName);
     }
     if (propertyName == L"IsMatchAppTheme" && IsMatchAppTheme)
     {
-        Graph->AxesColor = static_cast<Windows::UI::Color>(Application::Current->Resources->Lookup(L"DarkThemeAxisColor"));
-        Graph->GraphBackground = static_cast<Windows::UI::Color>(Application::Current->Resources->Lookup(L"DarkThemeGraphBackgroundColor"));
-        Graph->GridLinesColor = static_cast<Windows::UI::Color>(Application::Current->Resources->Lookup(L"DarkThemeGridLinesColor"));
-        Graph->GraphTheme = L"MatchApp";
+        GraphThemeSettingChanged(this, propertyName);
+        TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Theme, propertyName);
     }
 }
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -10,7 +10,6 @@ using namespace CalcManager::NumberFormattingUtils;
 using namespace GraphControl;
 using namespace std;
 using namespace Platform;
-using namespace Windows::Storage;
 using namespace Windows::UI::Xaml;
 
 GraphingSettingsViewModel::GraphingSettingsViewModel()
@@ -25,7 +24,6 @@ GraphingSettingsViewModel::GraphingSettingsViewModel()
     , m_dontUpdateDisplayRange()
     , m_XIsMinLastChanged(true)
     , m_YIsMinLastChanged(true)
-    , m_IsMatchAppTheme(false)
 {
 }
 
@@ -110,13 +108,4 @@ void GraphingSettingsViewModel::UpdateDisplayRange()
 bool GraphingSettingsViewModel::HasError()
 {
     return m_XMinError || m_YMinError || m_XMaxError || m_YMaxError || XError || YError;
-}
-
-void GraphingSettingsViewModel::SetGraphTheme(bool isMatchAppTheme)
-{
-    String ^ propertyName = isMatchAppTheme ? L"IsMatchAppTheme" : L"IsAlwaysLightTheme";
-    ApplicationDataContainer ^ localSettings = ApplicationData::Current->LocalSettings;
-    localSettings->Values->Insert(L"IsGraphThemeMatchApp", isMatchAppTheme);
-    GraphThemeSettingChanged(this, isMatchAppTheme);
-    TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Theme, propertyName);
 }
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -10,6 +10,7 @@ using namespace CalcManager::NumberFormattingUtils;
 using namespace GraphControl;
 using namespace std;
 using namespace Platform;
+using namespace Windows::UI::Xaml;
 
 GraphingSettingsViewModel::GraphingSettingsViewModel()
     : m_XMinValue(0)
@@ -23,6 +24,8 @@ GraphingSettingsViewModel::GraphingSettingsViewModel()
     , m_dontUpdateDisplayRange()
     , m_XIsMinLastChanged(true)
     , m_YIsMinLastChanged(true)
+    , m_IsAlwaysLightTheme(true)
+    , m_IsMatchAppTheme(false)
 {
 }
 
@@ -36,6 +39,18 @@ void GraphingSettingsViewModel::SetGrapher(Grapher ^ grapher)
         }
     }
     Graph = grapher;
+
+    if (Graph->GraphTheme == L"MatchApp")
+    {
+        IsAlwaysLightTheme = false;
+        IsMatchAppTheme = true;
+    }
+    else
+    {
+        IsAlwaysLightTheme = true;
+        IsMatchAppTheme = false;
+    }
+
     InitRanges();
     RaisePropertyChanged(L"TrigUnit");
 }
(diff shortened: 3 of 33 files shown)


build/pipelines/azure-pipelines.release.yaml
```diff
@@ -9,8 +9,8 @@ pr: none
 
 variables:
   versionMajor: 10
-  versionMinor: 2001
-  versionBuild: $[counter('10.2001.*', 0)]
+  versionMinor: 2002
```
mcooley: Let's increment this in a separate PR

src/CalcViewModel/Common/TraceLogger.cpp
```diff
@@ -303,12 +304,21 @@ namespace CalculatorApp
...
+    void TraceLogger::LogGraphSettingsChanged(GraphSettingsType settingType, String ^ settingValue)
     {
         auto fields = ref new LoggingFields();
         fields->AddString(StringReference(CALC_MODE), StringReference(GRAPHING_MODE));
         fields->AddInt16(L"SettingType", static_cast<int16>(settingType));
+        fields->AddString(L"SettingValue", settingValue);
 
         TraceLoggingCommon::GetInstance()->LogLevel2Event(StringReference(EVENT_NAME_GRAPH_SETTINGS_CHANGED), fields);
     }
+    void TraceLogger::LogGraphTheme(Platform::String ^ graphTheme)
```
mcooley: nit: add space above to match the style used in this file

mcooley: Also, I wish this file was more consistent about using "Platform::String" or "String". We should be consistent, at least for the code that's changing in this PR.

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -10,12 +10,14 @@ namespace CalculatorApp::ViewModel
     [Windows::UI::Xaml::Data::Bindable] public ref class GraphingSettingsViewModel sealed : public Windows::UI::Xaml::Data::INotifyPropertyChanged
     {
     public:
-        OBSERVABLE_OBJECT();
+        OBSERVABLE_OBJECT_CALLBACK(OnPropertyChanged);
```
mcooley: Classes which implement INotifyPropertyChanged shouldn't listen for their own property changes. It's inefficient, and the "stringly-typed" API is prone to break if someone renames properties in the future.

Instead, we should call GraphThemeSettingChanged in the _setter_ for the two properties you've added. We don't have a macro in Utils.h which makes this easy but you could add one along these lines:

```
// The callback specified in the macro is a method in the class that will be called every time the object changes
// The callback takes one argument, which is the new value of the property
#define OBSERVABLE_PROPERTY_RW_CALLBACK(t, n, c)                                                                                                                           \
    property t n                                                                                                                                               \
    {                                                                                                                                                          \
        t get()                                                                                                                                                \
        {                                                                                                                                                      \
            return m_##n;                                                                                                                                      \
        }                                                                                                                                                      \
        void set(t value)                                                                                                                                      \
        {                                                                                                                                                      \
            if (m_##n != value)                                                                                                                                \
            {                                                                                                                                                  \
                m_##n = value;                                                                                                                                 \
                c(value);
                RaisePropertyChanged(L#n);                                                                                                                     \
            }                                                                                                                                                  \
        }                                                                                                                                                      \
    }                                                                                                                                                          \
                                                                                                                                                               \
private:                                                                                                                                                       \
    t m_##n;                                                                                                                                                   \
                                                                                                                                                               \
public:
```

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -87,6 +89,31 @@ GraphingCalculator::GraphingCalculator()
...
     ZoomInButton->KeyboardAccelerators->Append(virtualKey);
+    m_uiSettings = ref new UISettings();
+    m_uiSettings->ColorValuesChanged += ref new TypedEventHandler<UISettings ^, Object ^>(this, &GraphingCalculator::OnColorValuesChanged);
+
+    ApplicationDataContainer ^ localSettings = ApplicationData::Current->LocalSettings;
+    if (localSettings == nullptr)
+    {
+        return;
+    }
+    if (localSettings->Values->HasKey(L"IsGraphThemeMatchApp") == true)
```
mcooley: no need for "== true"

mcooley: Add an empty line between the previous "}" and this if statement

src/GraphingImpl/Mocks/GraphingOptions.h
```diff
@@ -251,6 +252,19 @@ namespace MockGraphingImpl
             m_boxColor = Graphing::Color();
         }
 
+        virtual Graphing::Color GetGridColor() const
+        {
+            return m_gridColor;
+        }
+        virtual void SetGridColor(const Graphing::Color& value)
```
mcooley: nit: spaces between these methods

src/GraphingInterfaces/IGraphingOptions.h
```diff
@@ -52,9 +52,9 @@ namespace Graphing
         virtual void SetBackColor(const Graphing::Color& value) = 0;
         virtual void ResetBackColor() = 0;
 
-        virtual void SetAllowKeyGraphFeaturesForFunctionsWithParameters(bool kgf) = 0;
-        virtual bool GetAllowKeyGraphFeaturesForFunctionsWithParameters() const = 0;
-        virtual void ResetAllowKeyGraphFeaturesForFunctionsWithParameters() = 0;
+		virtual void SetAllowKeyGraphFeaturesForFunctionsWithParameters(bool kgf) = 0;
```
mcooley: nit: odd indents

src/GraphingImpl/Mocks/GraphingOptions.h
sanderl: The current style in this file has like methods grouped without spacing and the groups are spaced. I followed the spacing pattern that is already present.

src/CalcViewModel/Common/TraceLogger.cpp
joseartrivera: Style of the project is to omit the namespace in cpp files

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.h
```diff
@@ -274,12 +274,50 @@ namespace CalculatorApp::ViewModel
...
+                if (m_IsMatchAppTheme)
+                {
+                    GraphThemeSettingChanged(this, L"IsMatchAppTheme");
+                    TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Theme, L"IsMatchAppTheme");
                 }
             }
         }
 
     public:
+        event Windows::Foundation::EventHandler<Platform::String ^> ^ GraphThemeSettingChanged;
```
joseartrivera: Enum or bool might be better than a string here?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -87,6 +89,32 @@ GraphingCalculator::GraphingCalculator()
     virtualKey->Key = (VirtualKey)187; // OemAdd key
     virtualKey->Modifiers = VirtualKeyModifiers::Control;
     ZoomInButton->KeyboardAccelerators->Append(virtualKey);
+    m_uiSettings = ref new UISettings();
+    m_uiSettings->ColorValuesChanged += ref new TypedEventHandler<UISettings ^, Object ^>(this, &GraphingCalculator::OnColorValuesChanged);
+
+    ApplicationDataContainer ^ localSettings = ApplicationData::Current->LocalSettings;
+    if (localSettings == nullptr)
+    {
+        return;
```
joseartrivera: Since this method is a hodge podge of different initialization steps, I don't think we should return early. Someone else might tack on more initialization to the bottom and not realize that the code may never run. Either wrap the logic in the negative of this if statement, or keep the existing logic and move all of this into it's own initialization method that gets called from here.

sanderl: I updated this to check if localSettings != nullptr.

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml
```diff
@@ -183,6 +183,22 @@
...
+                       x:Uid="GraphThemeHeading"
+                       Margin="0,16,0,6"
+                       Style="{StaticResource SubTitleTextBoxStyle}"
+                       AutomationProperties.HeadingLevel="Level2"/>
+
+            <StackPanel Orientation="Vertical">
+                <RadioButton x:Name="AlwaysLightTheme"
+                             x:Uid="AlwaysLightTheme"
+                             IsChecked="{x:Bind ViewModel.IsAlwaysLightTheme, Mode=TwoWay}"/>
+                <RadioButton x:Name="MatchAppTheme"
```
joseartrivera: Are these x:Names ever used? If not we can remove them. Here and throughout the PR.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -688,3 +713,49 @@ void GraphingCalculator::OnVisualStateChanged(Object ^ sender, VisualStateChange
...
+}
+
+void GraphingCalculator::UpdateGraphTheme(ApplicationTheme appTheme)
+{
+    if (appTheme == ApplicationTheme::Dark)
+    {
+        VisualStateManager::GoToState(this, L"GrapherDarkTheme", true);
+    }
+
+    if (appTheme == ApplicationTheme::Light)
```
joseartrivera: How does this work with high contrast?

src/Calculator/App.xaml
```diff
@@ -70,20 +70,20 @@
...
-                    <SolidColorBrush x:Key="EquationBrush6" Color="#FF008055"/>
-                    <SolidColorBrush x:Key="EquationBrush7" Color="#FF58595B"/>
-                    <SolidColorBrush x:Key="EquationBrush8" Color="#FFE81123"/>
-                    <SolidColorBrush x:Key="EquationBrush9" Color="#FFE3008C"/>
-                    <SolidColorBrush x:Key="EquationBrush10" Color="#FFB31564"/>
-                    <SolidColorBrush x:Key="EquationBrush11" Color="#FFFFB900"/>
-                    <SolidColorBrush x:Key="EquationBrush12" Color="#FFF7630C"/>
-                    <SolidColorBrush x:Key="EquationBrush13" Color="#FF8E562E"/>
-                    <SolidColorBrush x:Key="EquationBrush14" Color="#FF000000"/>
+                    <SolidColorBrush x:Key="EquationBrush1" Color="#EF5865"/>
```
joseartrivera: Do the dark mode colors work during switching at run time? I forget if you need to call the function that refreshes them.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -1040,6 +1043,34 @@ void Grapher::OnGraphBackgroundPropertyChanged(Windows::UI::Color /*oldValue*/,
     }
 }
 
+
+void Grapher::OnGridLinesColorPropertyChanged(Windows::UI::Color /*oldValue*/, Windows::UI::Color newValue)
+{
+    if (m_renderMain != nullptr && m_graph != nullptr)
+    {
+        auto gridLinesColor = Graphing::Color(newValue.R, newValue.G, newValue.B, newValue.A);
+        m_graph->GetOptions().SetGridColor(gridLinesColor);
```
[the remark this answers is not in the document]
sanderl: This one has the namespaces because of the 2 Color types Windows::UI::Color and Graphing::Color. The other methods for background and axis color also have the namespaces.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -1040,6 +1043,34 @@ void Grapher::OnGraphBackgroundPropertyChanged(Windows::UI::Color /*oldValue*/,
     }
 }
 
+
+void Grapher::OnGridLinesColorPropertyChanged(Windows::UI::Color /*oldValue*/, Windows::UI::Color newValue)
```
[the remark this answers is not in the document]
sanderl: This one has the namespaces because of the 2 Color types Windows::UI::Color and Graphing::Color. The other methods for background and axis color also have the namespaces.

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
```diff
@@ -23,6 +25,8 @@ GraphingSettingsViewModel::GraphingSettingsViewModel()
     , m_dontUpdateDisplayRange()
     , m_XIsMinLastChanged(true)
     , m_YIsMinLastChanged(true)
+    , m_IsAlwaysLightTheme(true)
```
rudyhuyn: I don't think we need 2 properties, `IsAlwaysLightTheme` should be enough

src/GraphControl/Control/Grapher.h
```diff
@@ -50,6 +50,8 @@ public
         DEPENDENCY_PROPERTY_R_WITH_DEFAULT_AND_CALLBACK(GraphControl::EquationCollection ^, Equations, nullptr);
         DEPENDENCY_PROPERTY_WITH_DEFAULT_AND_CALLBACK(Windows::UI::Color, AxesColor, Windows::UI::Colors::Transparent);
         DEPENDENCY_PROPERTY_WITH_DEFAULT_AND_CALLBACK(Windows::UI::Color, GraphBackground, Windows::UI::Colors::Transparent);
+        DEPENDENCY_PROPERTY_WITH_DEFAULT_AND_CALLBACK(Windows::UI::Color, GridLinesColor, Windows::UI::Colors::Transparent);
+        DEPENDENCY_PROPERTY_WITH_DEFAULT_AND_CALLBACK(Platform::String ^, GraphTheme, L"Light");
```
rudyhuyn: We should use an enum here instead of a string

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -87,6 +89,28 @@ GraphingCalculator::GraphingCalculator()
     virtualKey->Key = (VirtualKey)187; // OemAdd key
     virtualKey->Modifiers = VirtualKeyModifiers::Control;
     ZoomInButton->KeyboardAccelerators->Append(virtualKey);
+    m_uiSettings = ref new UISettings();
+    m_uiSettings->ColorValuesChanged += ref new TypedEventHandler<UISettings ^, Object ^>(this, &GraphingCalculator::OnColorValuesChanged);
+
+    ApplicationDataContainer ^ localSettings = ApplicationData::Current->LocalSettings;
+
+    if (localSettings != nullptr && localSettings->Values->HasKey(L"IsGraphThemeMatchApp"))
```
rudyhuyn: `IsGraphThemeMatchApp` should be a const at the top of the file

src/GraphControl/Control/Grapher.cpp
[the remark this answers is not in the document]
sanderl: I tested this out and the gridlines color does not get cleared when a new equation is added.

src/Calculator/App.xaml
sanderl: It wasn't working, good catch. I got it fixed.

src/GraphControl/Control/Grapher.h
sanderl: GraphTheme has been removed from this file.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -592,8 +592,16 @@ void GraphingCalculator::GraphSettingsButton_Click(Object ^ sender, RoutedEventA
...
-    m_graphSettings = ref new GraphingSettings();
-    m_graphSettings->ViewModel->GraphThemeSettingChanged += ref new EventHandler<bool>(this, &GraphingCalculator::OnGraphThemeSettingChanged);
+    if (m_graphSettingsVM == nullptr)
+    {
+        m_graphSettingsVM = ref new GraphingSettingsViewModel();
+        m_graphSettingsVM->IsAlwaysLightTheme = !IsMatchAppTheme;
+        m_graphSettingsVM->IsMatchAppTheme = IsMatchAppTheme;
+    }
+
+    m_graphSettingsVM->GraphThemeSettingChanged += ref new EventHandler<bool>(this, &GraphingCalculator::OnGraphThemeSettingChanged);
```
joseartrivera: You want to move this into the above If statement otherwise you will keep assigning this event multiple times and the code will run multiple times whenever it fires.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -569,10 +593,18 @@ void GraphingCalculator::GraphSettingsButton_Click(Object ^ sender, RoutedEventA
 
 void GraphingCalculator::DisplayGraphSettings()
 {
-    auto graphSettings = ref new GraphingSettings();
-    graphSettings->SetGrapher(this->GraphingControl);
+    if (m_graphSettingsVM == nullptr)
```
rudyhuyn: I would prefer GraphSettings to continue to create its ViewModel and keep it private.

To sync IsMatchAppTheme, we should instead create a dependency property IsMatchAppTheme in GraphingSettings use a two-way binding created code behind to sync values

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -692,3 +729,54 @@ void GraphingCalculator::OnVisualStateChanged(Object ^ sender, VisualStateChange
...
+}
+
+void GraphingCalculator::UpdateGraphTheme(ApplicationTheme appTheme)
+{
+    if (m_accessibilitySettings->HighContrast)
+    {
+        VisualStateManager::GoToState(this, L"GrapherHighContrast", true);
+        return;
+    }
+    if (appTheme == ApplicationTheme::Dark)
```
rudyhuyn: Nit: `else if` or `switch`

rudyhuyn: Instead of calculating appTheme outside this method, for no reasons sometimes (high contrast), we should probably remove the argument and directly calculate the value to use in the function.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -692,3 +729,54 @@ void GraphingCalculator::OnVisualStateChanged(Object ^ sender, VisualStateChange
...
+    }
+}
+
+void GraphingCalculator::OnGraphThemeSettingChanged(Object ^ sender, bool isMatchAppTheme)
+{
+    if (isMatchAppTheme)
+    {
+        IsMatchAppTheme = true;
+        WeakReference weakThis(this);
+        this->Dispatcher->RunAsync(Windows::UI::Core::CoreDispatcherPriority::Normal, ref new Windows::UI::Core::DispatchedHandler([weakThis]() {
```
rudyhuyn: If we think that this method can be called outside the UI thread, we should also be do that when isMatvhAppTheme is false (wrap the conditional with the dispatcher)

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -53,6 +54,11 @@ void EquationInputArea::OnPropertyChanged(String ^ propertyName)
     {
         OnEquationsPropertyChanged();
     }
+
+    if (propertyName == IsMatchAppThemePropertyName)
```
rudyhuyn: Nit: `else if`

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: Nit: remove namespaces

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -569,10 +593,18 @@ void GraphingCalculator::GraphSettingsButton_Click(Object ^ sender, RoutedEventA
 
 void GraphingCalculator::DisplayGraphSettings()
 {
-    auto graphSettings = ref new GraphingSettings();
-    graphSettings->SetGrapher(this->GraphingControl);
+    if (m_graphSettingsVM == nullptr)
+    {
+        m_graphSettingsVM = ref new GraphingSettingsViewModel();
+        m_graphSettingsVM->IsMatchAppTheme = IsMatchAppTheme;
```
rudyhuyn: Would be great to always set it, not only when created (to be sure they stay in sync)

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -281,31 +287,43 @@ void EquationInputArea::FocusEquationIfNecessary(CalculatorApp::Controls::Equati
...
-        m_AvailableColors->Append(safe_cast<SolidColorBrush ^>(Application::Current->Resources->Lookup(L"EquationBrush12")));
-        m_AvailableColors->Append(safe_cast<SolidColorBrush ^>(Application::Current->Resources->Lookup(L"EquationBrush13")));
-        m_AvailableColors->Append(safe_cast<SolidColorBrush ^>(Application::Current->Resources->Lookup(L"EquationBrush14")));
+        Object ^ themeDictionaryName = L"Light";
+        if (IsMatchAppTheme && Application::Current->RequestedTheme == ApplicationTheme::Dark)
+        {
+            themeDictionaryName = L"Default";
+        }
+        auto themeDictionary = static_cast<ResourceDictionary ^>(Application::Current->Resources->ThemeDictionaries->Lookup(themeDictionaryName));
+        m_AvailableColors->Append(safe_cast<SolidColorBrush ^>(themeDictionary->Lookup(L"EquationBrush1")));
```
joseartrivera: I am seeing the dark mode colors show up in light mode when launching the app when I change the setting while in light mode.

PR #1148, opened 2020-04-02
Change trace point value precision to be dynamic based on graph scale

### LINKED ISSUE #998: Tracing coordinates should round dynamically based on graph scale
**Describe the bug**
We always round coordinates to one decimal place, which is problematic depending on graph scale.

**Expected behavior**
To ensure a great default experience, we should scale the precision of coordinates based on the graph viewing window. For this initial change, we want to avoid adding a setting, but may consider this in the future based on feedback.

* Desired precision for any one axis should be determined by: image
* This effectively produces the following results:  

| Range         | Precision |
|---------------|-----------|
| -0.1 to 0.1   | 10<sup>-4</sup>    |
| -1 to 1       | 10<sup>-3</sup>     |
| -10 to 10     | 10<sup>-2</sup>      |
| -100 to 100   | 10<sup>-1</sup>       |
| -1000 to 1000 | 10<sup>1</sup>         |
| -10000 to 10000 | 10<sup>2</sup>         |

* Note that we transition from one level of precision to the next based on range, not the value of min/max, so -500 to 500 is treated the same as -1000 to 1000 and has a precision of 10<sup>1</sup>.
* For prevision is very large (greater than 10<sup>7</sup>) or very small (less than 10<sup>-7</sup>), we should switch to scientific notation.
* For disproportional magnitudes (e.g., x is -1 to 1, but y is -1000 to 1000), take the higher precision between all axes.

**Screenshots**
image

joseartrivera:
Change trace point value to be dynamic based on display range

Manual tests


src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -197,12 +197,48 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
     GraphingControl->PlotGraph(false);
 }
 
+wstringstream GraphingCalculator::FormatTraceValue(double min, double max, float pointValue)
+{
+    wstringstream traceValueString;
+    auto roundingNumber = pow(10, floor(log10(max - min)) - 3);
```
rudyhuyn, line 203: We should add a comment explaining what we try to achieve here.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -197,12 +197,48 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+    {
+        traceValueString << fixed;
+    }
+
+    if (roundingNumber < 1)
+    {
+        auto roundingNumberStr = std::to_wstring(roundingNumber);
+        TrimTrailingZeros(roundingNumberStr);
+        auto precisionLength = roundingNumberStr.length() - 2;
+        traceValueString << setprecision(precisionLength) << pointValue;
```
rudyhuyn, line 219: Add comment explaining that we set the precision to the number of decimal

rudyhuyn:
Great job, but let's make sure to document enough so contributors can have a better understanding of our code

joseartrivera:
> Great job, but let's make sure to document enough so contributors can have a better understanding of our code

Fixed!

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: We should extract the precision:
`auto precision = floor(log10(...));`

And use it line 224

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -197,12 +197,55 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+    }
+    else
+    {
+        // If we are rounding to a place larger than 1, do that rounding here
+        if (roundingNumber > 1)
```
rudyhuyn, line 230: Why display an inaccurate number, knowing that we will use the same number of characters?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -197,12 +197,55 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+{
+    wstringstream traceValueString;
+
+    // This formula will give us the precision to use for the pointValue based on the min/max. For example: 
+    // If we want to round to two decimal places, this will return 0.01
+    // If we want to round to the nearest 100th this will return 100
+    auto roundingNumber = pow(10, floor(log10(max - min)) - 3);
+
+    // Determine if we want to show scientific notation instead
+    if (roundingNumber <= 0.0000001 || roundingNumber >= 10000000)
```
rudyhuyn, line 210: Why not rely on std::defaultfloat and use std::precision(min(-7, max(7, precision)) ?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: The reason is because it is hard to increment on smaller digits in the larger scales with your mouse. So to reduce noise while scanning with the mouse we round. I could be convinced to just display the full value (without decimals) if you feel strongly. We should consult with @grochocki too.

rudyhuyn: It's commonly accepted to simplify the mantissa of a number, but not the characteristic part (digits before decimal separator), even if the reasoning is the same,

joseartrivera: I don't feel too strongly either way. @grochocki thoughts?

grochocki: If I understand Rudy's feedback properly here, he is suggesting we never round digits before the decimal separator. The intended behavior here is that when it is no longer possible to make precise tracing movements at the appropriate power of ten, we should bump the rounding to the next power of ten. As the graph output window scales, we will become increasingly sporadic in displayed tracing values due to imprecise nature of pixel to coordinate space mapping. While it would be computationally accurate, is not pragmatic for users, because they are not able to move the mouse and see massive jumps in output. To see more precise x/y values, users can simply zoom in. Or perhaps a future improvement might be to surface a way for users to input an exact x or y value to see the other.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: Good idea

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: I don't know what the behavior is of std::defaultfloat but it wasn't respecting the precision that I passed to it.

I updated the code to use min and abs, it is unclear what a negative precision means so I avoided using that.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -207,12 +207,56 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+    // Extract precision we will round to 
+    auto precision = static_cast<int>(floor(log10(max - min)) - 3);
+
+    // This formula will give us the precision to use for the pointValue based on the min/max. For example:
+    // If we want to round to two decimal places, this will return 0.01
+    // If we want to round to the nearest 100th this will return 100
+    auto roundingNumber = pow(10, precision);
+
+        // Determine if we want to show scientific notation instead
+    if (roundingNumber <= 0.0000001 || roundingNumber >= 10000000)
```
rudyhuyn, line 223:
```suggestion
    if (precision <= -7 || precision  >= 7)
```

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -207,12 +207,56 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+        traceValueString << scientific;
+    }
+    else
+    {
+        traceValueString << fixed;
+    }
+
+
+    // If we are rounding to a decimal place, set the precision
+    if (roundingNumber < 1)
```
rudyhuyn, line 234:
```suggestion
    if (precision < 0)
```

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
rudyhuyn: @grochocki: I'm 100% OK with rounding X, but we should recalculate Y to be consistent with the new value of X and not round X **and** Y and display incorrect pair of values

rudyhuyn: X = 11111111111
Y = X*2 + 3

current behavior:
11111111111, 22222222225
this PR behavior:
11111110000, 22222220000 -> which is incorrect
Expected behavior:
11111110000, 22222220003

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -207,12 +207,45 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+    // Determine if we want to show scientific notation instead
+    if (precision <= -7 || precision  >= 7)
+    {
+        traceValueString << scientific;
+    }
+    else
+    {
+        traceValueString << fixed;
+    }
+
```
rudyhuyn, line 226: nit: extra line

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -207,12 +207,45 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
+
+    // If we are rounding to a decimal place, set the precision
+    if (precision < 0)
+    {
+        traceValueString << setprecision(::min(7, abs(precision))) << pointValue;
```
rudyhuyn, line 231: we can replace the conditional with `setprecision(::max(0, ::min(7, precision))`

joseartrivera: As discussed, it will be simpler to leave the logic as is since otherwise we will need to inverse the value of precision

PR #1149, opened 2020-04-03
Update Icons Implement the Graph View Button

### LINKED ISSUE #995: Refresh select graphing mode iconography
**Describe the bug**
Some graphing mode iconography is not as clear as we would like.

**Expected behavior**
Refresh the following icons:
* Hide Equation --> Explore "Closed Eye" instead
image
* Function Analysis --> Explore just lighting bolt instead
image
* Negate is not accurate for this mode --> switch to negative sign "(-)" instead
image
* Reset view icon could be clearer
image

### LINKED ISSUE #1070: Stop moving graph when adding new equation
Currently, when adding or removing equations, we _always_ move the graph in an attempt to provide users a "best fit" window. However, this can be disruptive for users that have carefully crafted a custom window view from graph settings or through pointer interactions and the on-screen view buttons.

To address this, we should replace the current "Reset View" button, which currently only allows the user to reset any manual changes back to "best fit", with a toggle button that allows users to switch between "automatic" and "manual" adjustment of the widow.

**Behavior**
* When starting a new graphing calculator session, we should be in automatic mode by default. If the user clicks the toggle button when we are in automatic mode, this should be a no-op.
* If at any point the user manually manipulates the graph output (e.g., setting x/y-min or x/y-max in graph options, zooming with mouse wheel, panning with touch, etc.), we enter manual mode.
* We remain in manual mode until either (a) the user clicks the toggle button or (b) the user deletes all equations from the list (effectively starting a new session).

**Keyboard Shortcuts**
The existing keyboard shortcut (Ctrl+0) should be maintained. Effectively, it should always put the user in automatic mode.

**Tooltips + Narrator**
* For Narrator, the toggle button should be "Graph View" button with the two reported states as "Automatic Best Fit" and "Manual Adjustment".
* Tooltip: "Refresh view automatically (Ctrl + 0)"

**Design**
This design mock shows the "automatic" (filled) and "manual" (unfilled) states of the button:
image

The icon can be constructed by [layering](https://docs.microsoft.com/en-us/windows/uwp/design/style/segoe-ui-symbol-font#layering-and-mirroring) these two new font icons:
image image

sanderl:
- Updated the icon font and updated the function analysis and negate icons
- Changed the zoom reset button to graph view with automatic best fit and manual adjustment
- Fixed bug where the window ranges in graph settings didn't update after panning or zooming

- manual


src/GraphControl/Control/Grapher.cpp
```diff
@@ -208,16 +209,7 @@ namespace GraphControl
...
-
-        // If the equation has changed, the IsLineEnabled state is reset.
-        // This checks if the equation has been reset and sets keepCurrentView to false in this case.
-        if (!equation->HasGraphError && !equation->IsValidated && equation->IsLineEnabled)
-        {
-            keepCurrentView = false;
-        }
-
-        PlotGraph(keepCurrentView);
+        PlotGraph();
```
joseartrivera, line 212: We should always keep the current view when enabling/disabling equations

src/GraphControl/Control/Grapher.cpp
```diff
@@ -784,6 +792,7 @@ namespace GraphControl
 
                 if (needsRenderPass)
                 {
+                    IsKeepCurrentView = true;
```
joseartrivera, line 795: IsKeepCurrentView should change in _response_ to the GraphViewChangedEvent. It doesn't make sense for this property to be changed from this function. GraphViewChanged can be fired from more than one location. Clicking the +/- buttons do not change IsKeepCurrentView currently, as an example.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -501,6 +585,7 @@
                                   ForceProportionalAxes="False"
                                   GraphPlottedEvent="GraphingControl_GraphPlottedEvent"
                                   GraphViewChangedEvent="GraphingControl_GraphViewChangedEvent"
+                                  IsKeepCurrentView="{x:Bind IsManualAdjustment, Mode=TwoWay}"
```
joseartrivera, line 588: This doesn't change to manual if a user changes the graph settings.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -792,6 +792,7 @@ namespace GraphControl
...
```
joseartrivera, line 795: Weird behavior in this scenario:
1. Make sure auto fit is on
2. Graph y=2 in f(1)
3. Graph y=300 in f(2)
4. Hide f(2) then hide f(1)
5. Notice graph snaps back to default position
6. Re-enable f(2) and f(1), graph is no longer correctly auto fitted even though it is toggled on

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
```diff
@@ -283,6 +283,79 @@
                     </Setter.Value>
                 </Setter>
             </Style>
+            <Style x:Name="GraphViewToggleButtonStyle"
```
joseartrivera, line 286: I liked the old style better, the toggle button is confusing imo, I am not sure when it is on/off base solely based on the font icon. It was clearer when the background was blue like the trace toggle button on the top of row.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -774,3 +773,24 @@ void GraphingCalculator::OnGraphThemeSettingChanged(Object ^ sender, bool isMatc
...
+    auto narratorNotifier = ref new NarratorNotifier();
+    String ^ announcementText;
+    if (IsManualAdjustment)
+    {
+        announcementText = AppResourceProvider::GetInstance()->GetResourceString(L"GraphViewManualAdjustmentAnnouncement");
+
+    }
+    else
+    {
+        GraphingControl->ResetGrid();
```
joseartrivera, line 788: Not sure if related to this PR, but it doesn't seem like ResetGrid is always going back to the original perfect auto fit after it was initially moved.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.h
```diff
@@ -25,11 +25,11 @@ public ref class GraphingCalculator sealed : public Windows::UI::Xaml::Data::INo
         DEPENDENCY_PROPERTY_OWNER(GraphingCalculator);
         COMMAND_FOR_METHOD(ZoomOutButtonPressed, GraphingCalculator::OnZoomOutCommand);
         COMMAND_FOR_METHOD(ZoomInButtonPressed, GraphingCalculator::OnZoomInCommand);
-        COMMAND_FOR_METHOD(ZoomResetButtonPressed, GraphingCalculator::OnZoomResetCommand);
         OBSERVABLE_PROPERTY_R(bool, IsKeyGraphFeaturesVisible);
         DEPENDENCY_PROPERTY(bool, IsSmallState);
         DEPENDENCY_PROPERTY(Platform::String ^, GraphControlAutomationName);
         OBSERVABLE_PROPERTY_R(bool, IsMatchAppTheme);
+        OBSERVABLE_PROPERTY_RW(bool, IsAutoBestFit);
```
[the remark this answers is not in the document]
joseartrivera: Clicking Reset view doesn't seem to reset it

src/GraphControl/Control/Grapher.cpp
sanderl: This issue was already existing before the button change.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
sanderl: The current design is to only fill in the circle with no background color change.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.h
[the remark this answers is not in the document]
sanderl: I think there are some issues with the ResetGrid API from the Graphing Engine. I've started a thread to investigate further.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
joseartrivera: In that case I think we should update the tool tip to let the user know which state the button will go to when clicked, it isn't clear which icon is on/off to me

PR #1154, opened 2020-04-07
Recycle equation colors when no longer in use

### LINKED ISSUE #991: Equation colors should be recycled as soon as they are no longer in use to maximize contrast
**Describe the bug**
Equation colors have a specified priority to maximize contrast between lines, but when removing an equation, the color is not recycled, leading to less than optimal contrast when entering new equations.

**Expected behavior**
Equation colors should be recycled as soon as they are no longer in use to maximize contrast

joseartrivera:
- Assign equation colors starting with the first unused color

- Manual tests


src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -90,17 +90,29 @@ void EquationInputArea::AddNewEquation()
...
 
     if (m_accessibilitySettings->HighContrast)
     {
+        m_lastLineColorIndex = (m_lastLineColorIndex + 1) % AvailableColors->Size;
         colorIndex = m_lastLineColorIndex;
     }
     else
     {
-        colorIndex = colorAssignmentMapping[m_lastLineColorIndex];
+        std::array<bool, maxEquationSize> colorAssignmentUsed = { false, false, false, false, false, false, false,
```
rudyhuyn, line 102: this is not necessary, C++ will zero-initialize all items if you write `std::array<bool, maxEquationSize> 
colorAssignmentUsed{}`

rudyhuyn: Even if `maxEquationSize` is equals to the number of colors available, it can change in the future (limit the number of equations because of performances on some devices, etc...).

We should create another constant, only used for colors

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -90,17 +90,29 @@ void EquationInputArea::AddNewEquation()
...
+            colorAssignmentUsed[equation->LineColorIndex] = true;
+        }
+
+        colorIndex = 0;
+        while (colorIndex < maxEquationSize - 1 && colorAssignmentUsed[colorAssignmentMapping[colorIndex]])
+        {
+            colorIndex++;
+        }
+
+        colorIndex = colorAssignmentMapping[colorIndex];
```
rudyhuyn, line 115: it should not happen but let's still make sure that this function correctly support the cause where we didn't find a color available (so colorIndex is out of array). Suggestion: return the last color of the array

joseartrivera: In that case we will hit the colorIndex < maxEquationSize case of the while loop and always return the last color (already tested it)

rudyhuyn: You are right, I missed the `<` instead of `<=`. Can you add a comment explaining it before someone tries to fix the code in a couple of years?

rudyhuyn: You are right, I missed the `<` instead of `<=`. Can you add a comment explaining it before someone tries to fix the code in a couple of years?

Issue #1153, opened 2020-04-07
Navigating back from function analysis or right-clicking an unfocused EquationEditbox results in invisible text on dark mode


MicrosoftIssueBot:
When navigating back from function analysis in dark theme, equation text is invisible until you click into the text edit area.

or 

Right-click on an equation edit box that does not currently have focus, the text is no longer visible.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

PR #1155, opened 2020-04-08
Updated EquationButton contrast fixed dark mode foreground color bugs

### LINKED ISSUE #1132: Luminosity ratio for "Hide" button is 2.5:1 which is less than 3:1
**Repro Steps**
1. Launch 'Calculator'.
2. Home screen will be displayed.​
3. Navigate through Menu using keyboard and select 'Graphing'. 
4. Navigate to hide button.
5. Check luminosity ratio for hide button using color contrast analyzer tool. ​
6. Observe the issue. 

**Actual Result:**
Luminosity ratio for "Hide" button is 2.5:1 which is less than 3:1.

**Expected Result:**
Luminosity ratio for "Hide" button should be minimum 3:1.

image

sanderl:
- Updates the foreground color on the equation button using a luminosity algorithm to create the correct contrast

- manually

### DIFF
--- a/src/Calculator/Controls/EquationTextBox.cpp
+++ b/src/Calculator/Controls/EquationTextBox.cpp
@@ -423,11 +423,6 @@ bool EquationTextBox::RichEditHasContent()
 
 void EquationTextBox::OnRichEditMenuOpened(Object ^ /*sender*/, Object ^ /*args*/)
 {
-    if (m_richEditBox != nullptr)
-    {
-        VisualStateManager::GoToState(m_richEditBox, "Focused", false);
-    }
-
     if (m_kgfEquationMenuItem != nullptr)
     {
         m_kgfEquationMenuItem->IsEnabled = m_HasFocus && !HasError && RichEditHasContent();
--- a/src/Calculator/Controls/EquationTextBox.cpp
+++ b/src/Calculator/Controls/EquationTextBox.cpp
@@ -423,7 +423,10 @@ bool EquationTextBox::RichEditHasContent()
 
 void EquationTextBox::OnRichEditMenuOpened(Object ^ /*sender*/, Object ^ /*args*/)
 {
-    VisualStateManager::GoToState(m_richEditBox, "Focused", false);
+    if (m_richEditBox != nullptr)
+    {
+        VisualStateManager::GoToState(m_richEditBox, "Focused", false);
+    }
 
     if (m_kgfEquationMenuItem != nullptr)
     {
--- a/src/Calculator/Controls/EquationTextBox.cpp
+++ b/src/Calculator/Controls/EquationTextBox.cpp
@@ -20,8 +20,10 @@ using namespace Windows::UI::Xaml::Automation;
 using namespace Windows::UI::Xaml::Controls;
 using namespace Windows::UI::Xaml::Input;
 using namespace Windows::UI::Xaml::Controls::Primitives;
+using namespace Windows::UI::Xaml::Media;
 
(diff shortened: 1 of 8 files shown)


src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -533,3 +533,17 @@ EquationViewModel ^ EquationInputArea::GetViewModelFromEquationTextBox(Object ^
 
     return eq;
 }
+
+ SolidColorBrush ^ EquationInputArea::GetForegroundColor(Color lineColor)
```
rudyhuyn, line 537: Nit: can you create an helper in Utils instead so we can use it in other controls if necessary?

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
```diff
@@ -533,3 +533,17 @@ EquationViewModel ^ EquationInputArea::GetViewModelFromEquationTextBox(Object ^
...
+ {
+     auto luminance = 0.2126 * lineColor.R + 0.7152 * lineColor.G + 0.0722 * lineColor.B;  
+
+     auto whiteLum = 0.2126 * Colors::White.R + 0.7152 * Colors::White.G + 0.0722 * Colors::White.B;
```
rudyhuyn, line 541: We don't have to calculate it, we can directly use `255`

src/Calculator/Views/GraphingCalculator/EquationInputArea.xaml.cpp
joseartrivera: Good idea, we may want to add a comment that we don't take alpha into account just in case

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -421,6 +423,8 @@ bool EquationTextBox::RichEditHasContent()
 
 void EquationTextBox::OnRichEditMenuOpened(Object ^ /*sender*/, Object ^ /*args*/)
 {
+    VisualStateManager::GoToState(m_richEditBox, "Focused", false);
```
joseartrivera, line 426: We should check if m_richEditBox is nullptr

joseartrivera, line 426: Would be better to set m_HasFocus = true and call UpdateCommonVisualState() so that the visual state logic can be centralized there.

src/Calculator/Controls/EquationTextBox.cpp
```diff
@@ -421,6 +423,11 @@ bool EquationTextBox::RichEditHasContent()
 
 void EquationTextBox::OnRichEditMenuOpened(Object ^ /*sender*/, Object ^ /*args*/)
 {
+    if (m_richEditBox != nullptr)
+    {
+        VisualStateManager::GoToState(m_richEditBox, "Focused", false);
```
joseartrivera, line 428: I can still repro #1153 even with this change

sanderl: huh...i can't repro it. Is it the same repro steps?

joseartrivera: Yeah it was the repro steps for right clicking an unfocused textbox but only repros in dark mode,  I can double check

sanderl: I removed the this bug from the PR and moved it to it's own PR.

PR #1183, opened 2020-04-14
Fixes the issues the focus state issues when users open the context menu on EquationEditBoxes

### LINKED ISSUE #1153: Navigating back from function analysis or right-clicking an unfocused EquationEditbox results in invisible text on dark mode

sanderl:
- Updated the AddNewEquation method to only set the focus state to the new equation if the submission source was not lost focus
- Updated the EquationEditBox focus states to also give focus to the RichEditBox when m_hasFocus is true.

- manually

### DIFF
--- a/src/Calculator/Controls/EquationTextBox.cpp
+++ b/src/Calculator/Controls/EquationTextBox.cpp
@@ -421,6 +421,11 @@ bool EquationTextBox::RichEditHasContent()
 
 void EquationTextBox::OnRichEditMenuOpened(Object ^ /*sender*/, Object ^ /*args*/)
 {
+    if (m_richEditBox != nullptr)
+    {
+        VisualStateManager::GoToState(m_richEditBox, L"Focused", false);
+    }
+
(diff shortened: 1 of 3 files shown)


sanderl:
> As I run through the issue more, I think there are a few more problems with submitting when we open the context menu than just this bug. For example, trying to copy/paste with an equation that has not yet been submitted is broken since when we submit the equation, the RichEditBox clears its selection and cursor state, so the intended copy/cut/paste the user wanted to do when they opened the menu is no longer possible.
> 
> Maybe the right fix here is not submitting at all when the context menu opens, thoughts?

We had an offline conversation and decided that the equation should not be submitted when the context menu is opened, but instead when the user clicks an action in the context menu that needs the equation to be graphed (i.e. analyze function)

sanderl:
Closing this PR since it was fixed in PR #1220

PR #1202, opened 2020-04-29
Fix styling bugs in graph settings and style panel

### LINKED ISSUE #1188: In graph options, x/y min/max should revert to the box style
**Describe the bug**
In graph options, x/y min/max should revert to the box style.

**Steps To Reproduce**
1. Open graph options and notice that the textboxes for x/y min/max do not work well in flyout window with variable text edit styling.

**Expected behavior**
These textboxes should retain default system styling.

### LINKED ISSUE #1190: In graph options, "Line Thickness" text should be same style as other L2 headers
In graph options, "Line Thickness" text should be same style as other L2 headers.

### LINKED ISSUE #1194: If you edit an equation to set an expression that does not support line style (e.g., inequality), the line option setting is still available
**Describe the bug**
If you edit an equation to set an expression that does not support line style (e.g., inequality), the line option setting is still available

**Steps To Reproduce**
1. Add a non-inequality expression as an equation (e.g., y = x)
2. Once submitted, edit the same equation to an inequality (e.g., x + y < 1)
3. Open up line style options, and notice that line style is still available, even though this setting should be disabled for inequalities.

**Expected behavior**
Line style option should be disable for inequalities (color should still work!).

joseartrivera:
- Reverts graph setting textbox to original style
- Fixes heading font sizes and spacing
- Hooks up line style combo box foreground color
- Make line style combo box binding mode to OneWay instead of OneTime

- Manual testing

### DIFF
--- a/src/Calculator/App.xaml
+++ b/src/Calculator/App.xaml
@@ -85,19 +85,6 @@
                     <SolidColorBrush x:Key="EquationBrush13" Color="#B0896D"/>
                     <SolidColorBrush x:Key="EquationBrush14" Color="#FFFFFF"/>
                     <SolidColorBrush x:Key="DividerBrush" Color="#60FFFFFF"/>
-                    <SolidColorBrush x:Key="GraphSettingsErrorBackgroundBrush" Color="#33EB5757"/>
-                    <SolidColorBrush x:Key="GraphSettingsErrorBorderBrush" Color="#FFEB5757"/>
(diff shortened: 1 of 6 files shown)


sanderl:
In Graph Settings, is the Line Thickness combo box supposed to align with the right side of the Trig Units buttons?

joseartrivera:
> It looks good. I was wondering if the LineThickness combobox was supposed to be wider. That can come in a different PR though.

Can you write a bug for it? I can do it in a new PR.

PR #1213, opened 2020-05-07
Improve launch performance by delay loading NavigationView

joseartrivera:
- This PR delay loads NavigationView until after the rest of the app is loaded and the UI is responsive.
- Removed Header and AoT button from NavView footer to allow it to be loaded independently of NavView
~~- Delay loads NavigationView by placing a placeholder Hamburger button that loads the full NavigationView UI when invoked.~~
- ~~Added logic to load NavView when hot keys are used to allow existing hot keys to still work (Alt + 1, 2, 3, 4 and Ctrl +E).~~
- ~~Delay loads GraphEngine DLL until the code needs it~~

I used the [MeasureAppStartupTime tool](https://gist.github.com/mcooley/86450e22ee0a8ffb6430ccec74317234) created by @mcooley to launch the app repeatedly and measure startup performance. 

There is roughly a ~16% decrease in start up time with these changes when measuring locally on my dev machine. The scenario tested is launching the Calculator app into standard mode at the default launch size (smallest possible).

Launch perf with the above circumstances before this PR:
Average: 455.8604766666667 ms

Launch perf with the above circumstances after this PR:
Average: 380.44501666666656 ms

### DIFF
--- a/src/Calculator/Resources/af-ZA/Resources.resw
+++ b/src/Calculator/Resources/af-ZA/Resources.resw
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
   <!-- 
     Microsoft ResX Schema 
--- a/src/Calculator/Resources/af-ZA/Resources.resw
+++ b/src/Calculator/Resources/af-ZA/Resources.resw
@@ -2699,10 +2699,6 @@
     <value>M</value>
     <comment>Access key for the Memory button. {StringCategory="Accelerator"}</comment>
   </data>
-  <data name="DummyNavViewButton.AccessKey" xml:space="preserve">
-    <value>HA</value>
-    <comment>Access key for the Hamburger button. {StringCategory="Accelerator"}</comment>
-  </data>
   <data name="NavView.AccessKey" xml:space="preserve">
     <value>H</value>
     <comment>Access key for the Hamburger button. {StringCategory="Accelerator"}</comment>
--- a/src/Calculator/Resources/af-ZA/Resources.resw
+++ b/src/Calculator/Resources/af-ZA/Resources.resw
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <root>
   <!-- 
     Microsoft ResX Schema 
@@ -2698,6 +2698,10 @@
   <data name="MemoryButton.AccessKey" xml:space="preserve">
     <value>M</value>
     <comment>Access key for the Memory button. {StringCategory="Accelerator"}</comment>
+  </data>
+  <data name="DummyNavViewButton.AccessKey" xml:space="preserve">
+    <value>HA</value>
+    <comment>Access key for the Hamburger button. {StringCategory="Accelerator"}</comment>
   </data>
   <data name="NavView.AccessKey" xml:space="preserve">
     <value>H</value>
--- a/src/Calculator/Views/MainPage.xaml
+++ b/src/Calculator/Views/MainPage.xaml
@@ -131,7 +131,7 @@
                     BorderThickness="0"
                     FontFamily="{ThemeResource SymbolThemeFontFamily}"
                     AutomationProperties.LandmarkType="Navigation"
-                    Click="DummyNavViewButton_Click"
+                    Click="DummyNavViewButtonClick"
                     Content="&#xE700;"/>
 
             <StackPanel Grid.Column="1"
--- a/src/Calculator/Views/MainPage.xaml
+++ b/src/Calculator/Views/MainPage.xaml
@@ -174,6 +174,7 @@
                              DataContext="{x:Bind Model}"
                              ExpandedModeThresholdWidth="Infinity"
                              IsBackButtonVisible="Collapsed"
+                             IsPaneOpen="False"
                              IsPaneToggleButtonVisible="{x:Bind Model.IsAlwaysOnTop, Converter={StaticResource BooleanNegationConverter}, Mode=OneWay}"
                              IsSettingsVisible="False"
                              ItemInvoked="OnNavItemInvoked"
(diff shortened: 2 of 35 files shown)


rudyhuyn:
Good improvement, but I'm curious if `Delay loads NavigationView by placing a placeholder Hamburger button that loads the full NavigationView UI when invoked.` is necessary and if we can only delay-load a sub-part of it.

Can you try without delay loading the navigationView but removing the check (read a registry key) related to Graphing Calculator mode and compare the perf? (I suppose it's the main factor)

joseartrivera:
> Good improvement, but I'm curious if `Delay loads NavigationView by placing a placeholder Hamburger button that loads the full NavigationView UI when invoked.` is necessary and if we can only delay-load a sub-part of it.
> 
> Can you try without delay loading the navigationView but removing the check (read a registry key) related to Graphing Calculator mode and compare the perf? (I suppose it's the main factor)

It's a good idea, I gave it a shot but didn't see too much of a difference, although I could see how it may add delay in certain machines. I think the main issue with perf related to NavView is the amount of UI elements it introduces to the visual tree.

Here are the measurements, re-ran all 3 scenarios, seems like my PC was slightly faster overall than when I first ran them in the PR, but the difference between the changes in this PR and master is still roughly ~75ms.

Master:
Average: 400.1869333333334 ms

Remove graph reg key check (Made IsGraphingModeEnabled and IsGraphingModeAvailable always return true);
Average: 396.1287066666669 ms

This PR:
Average: 326.93176666666665 ms

Can you elaborate on only delay loading a part of NavView? I wanted to see how these numbers look in the wild and perhaps propose an option in NavView itself to support this as an option if the numbers looked good. But maybe there is a way we can do it without modifying WinUI?

src/Calculator/Views/MainPage.xaml
```diff
@@ -130,15 +186,16 @@
...
 
             <muxc:NavigationView.PaneFooter>
                 <StackPanel HorizontalAlignment="Stretch" Orientation="Vertical">
                     <muxc:NavigationViewList x:Name="NavFooter"
-                                             Width="{x:Bind NavView.OpenPaneLength, Mode=Oneway}"
+                                             Width="{StaticResource SplitViewOpenPaneLength}"
                                              IsItemClickEnabled="True"
                                              ItemClick="OnAboutButtonClick"
-                                             x:Load="False">
+                                             x:Load="false">
```
rudyhuyn, line 198: nit:revert

src/Calculator/Views/MainPage.xaml
```diff
@@ -112,13 +112,69 @@
...
+                <ColumnDefinition Width="Auto"/>
+            </Grid.ColumnDefinitions>
+            <Button x:Name="DummyNavViewButton"
+                    x:Uid="DummyNavViewButton"
+                    Grid.Column="0"
+                    Width="{StaticResource HamburgerHeight}"
+                    Height="{StaticResource HamburgerHeight}"
+                    Style="{StaticResource SquareIconButtonStyle}"
+                    Background="Transparent"
+                    BorderThickness="0"
```
rudyhuyn, line 131: Should not be set when in high-contrast

src/GraphControl/GraphControl.vcxproj
```diff
@@ -148,6 +148,7 @@
       <IgnoreAllDefaultLibraries>false</IgnoreAllDefaultLibraries>
       <AdditionalLibraryDirectories>$(GraphingImplLibDir);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
       <AdditionalDependencies>$(GraphingImplLib);WindowsApp.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <DelayLoadDLLs>GraphingImpl.dll;%(DelayLoadDLLs)</DelayLoadDLLs>
```
rudyhuyn, line 151: Do you see any performance gain?

GraphControl is loaded only when the calculator displays the Graphing mode, requiring GraphingImpl.dll to be loaded too.

joseartrivera: The performance gain is minimal, I think it may be more noticeable in slower machines, but I don't think its true that it only loads when in graphing mode. I've seen it loaded when graphing mode has not yet been visited. I have also confirmed by looking at the modules window in VS while debbugging. I tried adding the GraphControl DLL to the delay loaded dlls for the main Calculator project but it wasn't working for some reason, probably because we use the viewmodels. But this prevents both the bulk of the GraphingImpl dll and internal Graphing Engine dll from loading when a user launches into other modes.

src/Calculator/Views/MainPage.xaml
joseartrivera: Good catch!

mcooley:
@joseartrivera Can we check in these improvements as separate commits, and measure the impact of each one separately? That way if any of them end up causing problems later on, we'll know whether reverting is a big deal or not.

joseartrivera:
> @joseartrivera Can we check in these improvements as separate commits, and measure the impact of each one separately? That way if any of them end up causing problems later on, we'll know whether reverting is a big deal or not.

Done!

src/Calculator/Resources/af-ZA/Resources.resw
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
rudyhuyn, line 1: can you rollback this change?

src/Calculator/Views/MainPage.xaml
```diff
@@ -118,6 +118,49 @@
...
+              HorizontalAlignment="Left"
+              VerticalAlignment="Top">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="{StaticResource HamburgerHeightGridLength}"/>
+                <ColumnDefinition Width="Auto"/>
+                <ColumnDefinition Width="Auto"/>
+            </Grid.ColumnDefinitions>
+
+            <StackPanel Grid.Column="1"
+                        Margin="10,0,0,0"
```
rudyhuyn, line 132:
```suggestion
                        Margin="8,0,0,0"
```

The original left margin was 52, so this margin should be `12,0,0,0` but we also removed [4 pixels recently ](https://github.com/microsoft/calculator/pull/1215/files), the new margin should be 8,0,0,0

src/Calculator/Resources/af-ZA/Resources.resw
joseartrivera: fixed!

src/Calculator/Views/MainPage.xaml
joseartrivera: Good catch

PR #1223, opened 2020-05-09
Simplify KeyboardShortcutManager and correctly support Alt-Gr key

### LINKED ISSUE #1222: 2 events are launched when typing a character using Alt-Gr
**Describe the bug**
The AltGr key available with many non-US keyboards is the equivalent of Ctrl+Alt and is used to access to alternative characters (AltGr+Q will generate a `@` with a German keyboard for example).

In this case, the calculator will see 2 different events: User typed @ (`sqrt` in scientific mode) and users typed Ctrl+Q (M-) and will virtual press both buttons.

**Steps To Reproduce**
- Switch to German keyboard
- Type AltGr+Q (or Ctrl+Alt+Q)

**Expected behavior**
Only `sqrt` should be called.

**Requested Assignment**
I will fix it

rudyhuyn:
- Make sure to not consider `Ctrl+Alt+key` as `Ctrl+key`
- Clean KeyboardShortcutManager
   - Remove `VirtualKeyInverseChord` and `VirtualKeyControlInverseChord`, `ShiftButtonChecked` is never called, so `s_ShiftButtonChecked.find(viewId)->second)` is always false.
   - Remove code to manage the Escape key for the About dialog (not necessary, the flyout already 
manages it)
   - Remove `UpdateDropDownState(Flyout ^ aboutPageFlyout)` to only use `UpdateDropDownState(bool)`
   - Refactor how we manage the `Ctrl`, `Alt`, `Shift` keys. We now check directly if the key is pressed instead of using 3 `std::map` to save their states with the risk of missing a KeyUp event and corrupting the "cache" (can happen when the application loses focus while the key is still pressed).
   - Refactor `GetCurrentKeyDictionary` to always return the correct list of shortcuts instead of defaulting to the `s_VirtualKeysForButtons` when we don't recognize the key combination.
   - Move `GetCurrentKeyDictionary` and global fields to private static method/fields (to make the conversion to C# easier)
   - Replace `multimap` with `map` when we didn't need them

- manually in Fr, En and De


src/Calculator/Common/KeyboardShortcutManager.h
```diff
@@ -60,20 +55,31 @@ namespace CalculatorApp
...
             static void
             OnVirtualKeyControlShiftChordPropertyChanged(Windows::UI::Xaml::DependencyObject ^ target, MyVirtualKey oldValue, MyVirtualKey newValue);
 
             static void OnCharacterReceivedHandler(Windows::UI::Core::CoreWindow ^ sender, Windows::UI::Core::CharacterReceivedEventArgs ^ args);
             static void OnKeyDownHandler(Windows::UI::Core::CoreWindow ^ sender, Windows::UI::Core::KeyEventArgs ^ args);
-            static void OnKeyUpHandler(Windows::UI::Core::CoreWindow ^ sender, Windows::UI::Core::KeyEventArgs ^ args);
             static void OnAcceleratorKeyActivated(Windows::UI::Core::CoreDispatcher ^, Windows::UI::Core::AcceleratorKeyEventArgs ^ args);
+            static const std::multimap<MyVirtualKey, Platform::WeakReference>* KeyboardShortcutManager::GetCurrentKeyDictionary(bool altPressed = false);
+
+        private:
```
mcooley, line 68: nit: for all of these members let's make the first character after "s_" lowercase. s_CharacterForButtons -> s_characterForButtons, etc.

src/Calculator/Common/KeyboardShortcutManager.h
```diff
@@ -60,20 +55,31 @@ namespace CalculatorApp
...
+            static std::map<int, std::multimap<MyVirtualKey, Platform::WeakReference>> s_VirtualKeyAltChordsForButtons;
+            static std::map<int, std::multimap<MyVirtualKey, Platform::WeakReference>> s_VirtualKeyControlShiftChordsForButtons;
+
+            static std::map<int, bool> s_IsDropDownOpen;
+            static std::map<int, bool> s_ignoreNextEscape;
+            static std::map<int, bool> s_keepIgnoringEscape;
+            static std::map<int, bool> s_fHonorShortcuts;
+            static std::map<int, bool> s_fDisableShortcuts;
+
+            static Concurrency::reader_writer_lock s_keyboardShortcutMapLock;
```
mcooley, line 82: Looks like this isn't used anymore?

src/Calculator/Common/KeyboardShortcutManager.cpp
```diff
@@ -421,215 +365,159 @@ void KeyboardShortcutManager::OnCharacterReceivedHandler(CoreWindow ^ sender, Ch
...
-        return s_VirtualKeyAltChordsForButtons.find(viewId)->second;
-    }
-    else if (
-        (s_ShiftKeyPressed.find(viewId)->second)
-        && ((Window::Current->CoreWindow->GetKeyState(VirtualKey::Control) & CoreVirtualKeyStates::Down) == CoreVirtualKeyStates::Down))
-    {
-        return s_VirtualKeyControlShiftChordsForButtons.find(viewId)->second;
-    }
-    else if (s_ShiftKeyPressed.find(viewId)->second)
+    bool isControlKeyPressed = (Window::Current->CoreWindow->GetKeyState(VirtualKey::Control) & CoreVirtualKeyStates::Down) == CoreVirtualKeyStates::Down;
```
mcooley, line 382: In OnKeyDownHandler in some cases we'll call CoreWindow->GetKeyState twice, and we could get into an unexpected state if CoreWindow doesn't return consistent results.

I think the caller should be responsible for passing in the state of all three modifier keys, not just alt.

src/Calculator/Common/KeyboardShortcutManager.h
rudyhuyn: `s_keyboardShortcutMapLock`? It's used in many methods in KeyboardShortcutManager.cpp

src/Calculator/Common/KeyboardShortcutManager.cpp
rudyhuyn: I'm fine with that!

src/Calculator/Common/KeyboardShortcutManager.h
mcooley: Oops, I was looking at the diff wrong.

PR #1226, opened 2020-05-11
Handle non breaking white spaces in copy paste manager

### LINKED ISSUE #969: Remove space when we copy/paste a number
**Describe the bug**

When we copy the result of a calc, we got a number with space character between numbers (Eg: "2 365 596")

**Steps To Reproduce**

1. Calc 568 * 568 * 568
2. Copy the result
3. Clic on "C"
4. Paste the previous number
5. Invalid number ("Entrée non valide" on a French app)

**Expected behavior**
Remove spaces when we copy it.

 - OS Build: 1903
 - Architecture: x64
 - Application Version: ?
 - Region: FR_fr

EriWong:
- adds the non-breaking white space character to the list of removed characters from pasted strings

manually


rudyhuyn:
We can add 0x20 to the list of characters to remove, but we should also call `LocalizationSettings::RemoveGroupSeparators` so it will also solve the issue for all other locales

PR #1229, opened 2020-05-12
Remove ICustomPropertyProviders

rudyhuyn:
- `HistoryItemViewModel` and `MemoryItemViewModel` implements `ICustomPropertyProvider` only to support UI tests while we can have access to the information without. We should minimize at the maximum code specific to UI tests in Calculator and instead modify `CalculatorUITestFramework`

- Manually


src/CalculatorUITestFramework/HistoryPanel.cs
```diff
@@ -34,10 +36,10 @@ public void OpenHistoryPanel()
         /// Gets all of the history items listed in the History Pane.
         /// </summary>
         /// <returns>A readonly collection of history items.</returns>
-        public ReadOnlyCollection<AppiumWebElement> GetAllHistoryListViewItems()
+        public List<HistoryItem> GetAllHistoryListViewItems()
         {
             OpenHistoryPanel();
-            return this.HistoryListView.FindElementsByClassName("ListViewItem");
+            return (from item in this.HistoryListView.FindElementsByClassName("ListViewItem") select new HistoryItem(item)).ToList();
```
mcooley, line 42: nit: I'd prefer that we use method syntax instead of query syntax for LINQ

PR #1231, opened 2020-05-12
Updates ResetGrid to account for when equations are added during Manual Adjustment mode

### LINKED ISSUE #1187: Turning on refresh view automatically does not update the view properly
**Describe the bug**
Turning on refresh view automatically does not update the view properly.

**Steps To Reproduce**
1. Enter one or more equations and notice that the graph widow updates for "best fit"
2. Pan the graph to explore
3. Enter another equation and notice the graph window does not automatically update
4. Click the "Refresh view automatically" button in the bottom right corner
5. Notice that the graph window does not always update to the proper "best fit" view.

**Expected behavior**
Turning on refresh view automatically should update the view properly to "best fit".

**Additional context**
Does not always repro 100%.

sanderl:
- Tracks when we add equations in Manual Adjustment mode
- Conditionally resets using TryPlotGraph or ResetRange depending on whether or not equations were added during Manual Adjustment mode

- manually

### DIFF
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -100,8 +100,7 @@ void GraphingSettingsViewModel::UpdateDisplayRange()
         return;
     }
 
-    m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue);
-    m_Graph->RangeUpdatedBySettings = true;
+    m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue, true);
 
     TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Grid, L"");
 }
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -101,6 +101,7 @@ void GraphingSettingsViewModel::UpdateDisplayRange()
     }
 
     m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue);
+    m_Graph->RangeUpdatedBySettings = true;
 
     TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Grid, L"");
 }
--- a/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
+++ b/src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
@@ -100,7 +100,7 @@ void GraphingSettingsViewModel::UpdateDisplayRange()
         return;
     }
 
-    m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue, true);
+    m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue);
 
     TraceLogger::GetInstance()->LogGraphSettingsChanged(GraphSettingsType::Grid, L"");
 }
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -124,13 +124,15 @@ namespace GraphControl
                 {
                     TryPlotGraph(false, false);
                     m_resetUsingInitialDisplayRange = false;
-                }
-                else if (SUCCEEDED(renderer->ResetRange()))
-                {
-                    m_renderMain->RunRenderPass();
+                    GraphViewChangedEvent(this, GraphViewChangedReason::Reset);
+                    return;
                 }
 
-                GraphViewChangedEvent(this, GraphViewChangedReason::Reset);
+                if (SUCCEEDED(renderer->ResetRange()))
+                {
+                    m_renderMain->RunRenderPass();
+                    GraphViewChangedEvent(this, GraphViewChangedReason::Reset);
+                }
             }
         }
     }
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -122,22 +122,15 @@ namespace GraphControl
             {
                 if (m_resetUsingInitialDisplayRange)
                 {
-                    renderer->SetDisplayRanges(m_initialDisplayRangeXMin, m_initialDisplayRangeXMax, m_initialDisplayRangeYMin, m_initialDisplayRangeYMax);
-                    auto xCenter = (m_initialDisplayRangeXMax - m_initialDisplayRangeXMin) / 2;
-                    auto yCenter = (m_initialDisplayRangeYMax - m_initialDisplayRangeYMin) / 2;
-                    renderer->ScaleRange(xCenter, yCenter, 1);
+                    TryPlotGraph(false, false);
                     m_resetUsingInitialDisplayRange = false;
-                    m_renderMain->RunRenderPass();
-                    GraphViewChangedEvent(this, GraphViewChangedReason::Reset);
-                    return;
                 }
-                if (SUCCEEDED(renderer->ResetRange()))
+                else if (SUCCEEDED(renderer->ResetRange()))
                 {
                     m_renderMain->RunRenderPass();
                 }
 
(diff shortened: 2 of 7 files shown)


src/GraphControl/Control/Grapher.cpp
```diff
@@ -1091,6 +1099,10 @@ optional<vector<shared_ptr<Graphing::IEquation>>> Grapher::TryInitializeGraph(bo
         double xMin, xMax, yMin, yMax;
         m_graph->GetRenderer()->GetDisplayRanges(xMin, xMax, yMin, yMax);
         auto initResult = m_graph->TryInitialize(graphingExp);
+        if (IsKeepCurrentView)
+        {
+            m_resetUsingInitialDisplayRange = true;
+        }
         m_graph->GetRenderer()->SetDisplayRanges(xMin, xMax, yMin, yMax);
```
joseartrivera, line 1106: Not this PR, but a crash will happen here if the above line `auto initResult = m_graph->TryInitialize(graphingExp);` fails to properly initialize the graph and returns nullopt to initResult.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -1088,10 +1099,21 @@ optional<vector<shared_ptr<Graphing::IEquation>>> Grapher::TryInitializeGraph(bo
...
     {
+        auto renderer = m_graph->GetRenderer();
         double xMin, xMax, yMin, yMax;
-        m_graph->GetRenderer()->GetDisplayRanges(xMin, xMax, yMin, yMax);
+        renderer->GetDisplayRanges(xMin, xMax, yMin, yMax);
         auto initResult = m_graph->TryInitialize(graphingExp);
-        m_graph->GetRenderer()->SetDisplayRanges(xMin, xMax, yMin, yMax);
+        if (IsKeepCurrentView)
+        {
+            renderer->PrepareGraph();
```
joseartrivera, line 1108: Not sure if related to this change or internal graph engine changes, but when switching to gradians with y=x, sin(x), and x^2+y^2=1 graphed I see a new crash

src/GraphControl/Control/Grapher.cpp
```diff
@@ -118,9 +118,20 @@ namespace GraphControl
...
-                if (SUCCEEDED(renderer->ResetRange()))
+                HRESULT hr;
+                if (m_resetUsingInitialDisplayRange)
+                {
+                    hr = renderer->SetDisplayRanges(m_initialDisplayRangeXMin, m_initialDisplayRangeXMax, m_initialDisplayRangeYMin, m_initialDisplayRangeYMax);
+                    m_resetUsingInitialDisplayRange = false;
+                }
+                else
+                {
+                    hr = renderer->ResetRange();
```
joseartrivera, line 131: I don't think this gets called in this case:

1. Graph anything
2. Open up graph settings, set max x-range to 3000
3. Move graph around, click reset, resets back to view to when you set x-range to 3000 instead of best fit

src/GraphControl/Control/Grapher.cpp
```diff
@@ -1088,10 +1099,21 @@ optional<vector<shared_ptr<Graphing::IEquation>>> Grapher::TryInitializeGraph(bo
 {
     if (keepCurrentView || IsKeepCurrentView)
     {
+        auto renderer = m_graph->GetRenderer();
         double xMin, xMax, yMin, yMax;
-        m_graph->GetRenderer()->GetDisplayRanges(xMin, xMax, yMin, yMax);
+        renderer->GetDisplayRanges(xMin, xMax, yMin, yMax);
         auto initResult = m_graph->TryInitialize(graphingExp);
-        m_graph->GetRenderer()->SetDisplayRanges(xMin, xMax, yMin, yMax);
+        if (IsKeepCurrentView)
```
joseartrivera, line 1106: The crashing issue I mentioned before still applies here, if initResult == nullopt we probably shouldn't try to call GetDisplayRanges in this block either.

src/GraphControl/Control/Grapher.cpp
joseartrivera, line 1108: Can this return an error HResult? Should we handle that?

sanderl: I can't repro this crash.

src/GraphControl/Control/Grapher.cpp
sanderl: In Graph Settings we directly call GetDisplayRanges and SetDisplayRanges. In the case where the range was modified via settings, I have added a flag in that case we now replot the graph.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -1088,10 +1108,23 @@ optional<vector<shared_ptr<Graphing::IEquation>>> Grapher::TryInitializeGraph(bo
...
         double xMin, xMax, yMin, yMax;
-        m_graph->GetRenderer()->GetDisplayRanges(xMin, xMax, yMin, yMax);
+        renderer->GetDisplayRanges(xMin, xMax, yMin, yMax);
         auto initResult = m_graph->TryInitialize(graphingExp);
-        m_graph->GetRenderer()->SetDisplayRanges(xMin, xMax, yMin, yMax);
+        if (initResult != nullopt)
+        {
+            if (IsKeepCurrentView)
+            {
+                if (SUCCEEDED(renderer->PrepareGraph()))
```
joseartrivera, line 1119: Do you know what this function is doing or when it can fail? A comment explaining would be useful. Also, is the code path for when this fails valid? Will the below call to renderer->SetDisplayRanges crash? Is an error shown on the last added equation?

src/GraphControl/Control/Grapher.cpp
joseartrivera: Okay, I checked master and it repros there too so likely not an issue with this PR.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -118,13 +118,33 @@ namespace GraphControl
...
+                    hr = renderer->SetDisplayRanges(m_initialDisplayRangeXMin, m_initialDisplayRangeXMax, m_initialDisplayRangeYMin, m_initialDisplayRangeYMax);
+                    m_resetUsingInitialDisplayRange = false;
+                }
+                else if (m_rangeUpdatedBySettings)
+                {
+                    IsKeepCurrentView = false;
+                    TryPlotGraph(false, false);
+                    m_rangeUpdatedBySettings = false;
+                    GraphViewChangedEvent(this, GraphViewChangedReason::Reset);
+                    return;
```
joseartrivera, line 135: This function is a little harder to follow, can we add a comment explaining when we would enter each block and noting that we return early in this one?

src/CalcViewModel/GraphingCalculator/GraphingSettingsViewModel.cpp
```diff
@@ -101,6 +101,7 @@ void GraphingSettingsViewModel::UpdateDisplayRange()
     }
 
     m_Graph->SetDisplayRanges(m_XMinValue, m_XMaxValue, m_YMinValue, m_YMaxValue);
+    m_Graph->RangeUpdatedBySettings = true;
```
joseartrivera, line 104: I am not a fan of the viewmodel keeping track of this state, can we set this to true/false in SetDisplayRanges instead and have it be a private member variable?

src/GraphingImpl/Mocks/GraphRenderer.h
```diff
@@ -98,6 +98,11 @@ namespace MockGraphingImpl
             return S_OK;
         }
 
+        virtual HRESULT PrepareGraph()
```
joseartrivera, line 101: Hovering over lines on the graph and variable editing are broken in this PR/new engine.

src/GraphControl/Control/Grapher.cpp
sanderl: I've asked the graphing engine team what happens. I'm waiting to hear back.

src/GraphingImpl/Mocks/GraphRenderer.h
sanderl: I can't repro these issues.

src/GraphControl/Control/Grapher.cpp
sanderl: The graphing engine team got back to me and said that PrepareGraph should have no affect on any other API calls, so we are safe to call SetDisplayRanges if PrepareGraph fails.

src/GraphingImpl/Mocks/GraphRenderer.h
sanderl: I was able to repro this for y=mx, I've reached out to the graphing team.

PR #1239, opened 2020-05-20
Adding Preview tag to Resources File

quentin987:
"Preview" is not getting translated to other languages. Added to resource file so it would.

Manual testing

### DIFF
--- a/src/Calculator/Resources/en-US/Resources.resw
+++ b/src/Calculator/Resources/en-US/Resources.resw
@@ -2729,6 +2729,10 @@
     <value>Microsoft Software License Terms</value>
     <comment>Displayed on a link to the Microsoft Software License Terms on the About panel</comment>
   </data>
+  <data name="PreviewTag.Text" xml:space="preserve">
+    <value>Preview</value>
+    <comment>Preview tag for upcoming features</comment>
+  </data>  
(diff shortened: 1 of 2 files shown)


src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -2729,6 +2729,10 @@
     <value>Microsoft Software License Terms</value>
     <comment>Displayed on a link to the Microsoft Software License Terms on the About panel</comment>
   </data>
+  <data name="PreviewTag.Text" xml:space="preserve">
+    <value>Preview</value>
+    <comment>Preview tag for upcoming features</comment>
```
quentin987, line 2734: If someone has an alternative wording that they'd prefer. Feel free to comment and I can update.

mcooley: I think that you've written is OK. I might write "Label displayed next to upcoming features"

The important thing about the comment is to help a human translator understand how the string will be used, since they probably won't see the string in context. In this case we'll want them to understand that we mean "preview" in the sense of "beta feature" rather than in the some other sense (like "print preview"). It's also helpful to describe whether the text is for display, a screen reader, or something else.

quentin987: Good suggestion, updated.

PR #1250, opened 2020-05-26
Updated APIs to use new GetClosePointData() from Graphing Engine.

### LINKED ISSUE #998: Tracing coordinates should round dynamically based on graph scale

quentin987:
## Fixes #998.

Note: Rounding/precision isn't 100% accurate at this point, addressing separately.

-Updated API endpoints to use new GetClosePointData() API. 
-Specifying precision before consuming the API. 
-Removed Trace formatting unused function.

- Manual Testing
- Graphed lines y = x and y = 1.54321 and zoomed out and in to test different views.


src/GraphControl/DirectX/RenderMain.cpp
```diff
@@ -157,19 +157,20 @@ namespace GraphControl::DX
             critical_section::scoped_lock lock(m_criticalSection);
 
             int formulaId = -1;
-            float nearestPointLocationX, nearestPointLocationY;
-            double nearestPointValueX, nearestPointValueY, rhoValueOut, thetaValueOut, tValueOut;
+            float nearestPointLocationX, nearestPointLocationY, nearestPointValueX, nearestPointValueY;
```
joseartrivera, line 160: Is there a version of the API where they are all doubles?

src/GraphingInterfaces/IGraphRenderer.h
```diff
@@ -24,16 +24,14 @@ namespace Graphing::Renderer
 
 		virtual HRESULT DrawD2D1(ID2D1Factory* pDirect2dFactory, ID2D1RenderTarget* pRenderTarget, bool& hasSomeMissingDataOut) = 0;
         virtual HRESULT GetClosePointData(
```
joseartrivera, line 26: I getting a read access violation when hovering over the graph. I think you may need to update this and merge it with the latest changes, including your changes to the graph engine wrapper.

src/GraphControl/DirectX/RenderMain.cpp
quentin987: there are, but they appear to be deprecated, the ones that have doubles as args don't have the precision parameter.

quentin987:
pushed updated changes, exception shouldn't throw when you use tracing

src/GraphingInterfaces/IGraphRenderer.h
quentin987: pushed updated changes. let me know if you still hit this issue

src/GraphControl/DirectX/RenderMain.cpp
```diff
@@ -323,11 +350,11 @@ namespace GraphControl::DX
...
                                 }
 
-                                m_TraceLocation = Point(nearestPointLocationX, nearestPointLocationY);
+                                m_TraceLocation = Point(outNearestPointLocationX, outNearestPointLocationY);
                                 m_nearestPointRenderer.Render(m_TraceLocation);
                                 m_Tracing = true;
-                                m_TraceLocation = Point(nearestPointLocationX, nearestPointLocationY);
-                                m_TraceValue = Point(nearestPointValueX, nearestPointValueY);
+                                m_TraceLocation = Point(outNearestPointLocationX, outNearestPointLocationY);
+                                m_TraceValue = Point(outNearestPointValueX, outNearestPointValueY);
```
joseartrivera, line 357: Not this PR, but it looks like there is a flaw here that is changing the result of the y value.

For the equation y=x*1.234567 when x=100200 the graphing engine correctly sets outNearestPointValueY = 123703.61340000000.

However, when we put this value into a Point, Point takes a float and converts this value to 123703.617

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -207,44 +207,14 @@ void GraphingCalculator::OnEquationsVectorChanged(IObservableVector<EquationView
...
 void GraphingCalculator::OnTracePointChanged(Point newPoint)
 {
     wstringstream traceValueString;
 
     double xAxisMin, xAxisMax, yAxisMin, yAxisMax;
     GraphingControl->GetDisplayRanges(&xAxisMin, &xAxisMax, &yAxisMin, &yAxisMax);
-
-    traceValueString << "(" << FormatTraceValue(xAxisMin, xAxisMax, newPoint.X).str() << ", " << FormatTraceValue(yAxisMin, yAxisMax, newPoint.Y).str() << ")";
+    
+    traceValueString << "(" << newPoint.X << ", " << newPoint.Y << ")";
```
joseartrivera, line 217: Another change occurs here when converting the float to a string, in the same scenario as my other comment the y value goes from 123703.617 to 123704. I think to fix this we may want to call SetPrecision and set it to the maximum amount of decimals allowed and perhaps run the string through TrimTrailingZeros in NumberFormattingUtils.

quentin987: Great catch, float was introducing some unnecessary rounding, pushed changed to convert to use doubles instead of built-in point class.

src/GraphControl/DirectX/RenderMain.cpp
quentin987: Fixed, pushed updates.

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
```diff
@@ -16,6 +16,8 @@
 #include "Calculator/Views/GraphingCalculator/EquationInputArea.xaml.h"
 #include "CalcViewModel/Common/Utils.h"
 #include "GraphingSettings.xaml.h"
+#include <iomanip>
```
joseartrivera, line 19: nit: We should put this and the below header file in pch.h

src/GraphingImpl/Mocks/GraphRenderer.h
```diff
@@ -40,23 +40,22 @@ namespace MockGraphingImpl
...
+            double precision,
             int& formulaIdOut,
             float& xScreenPointOut,
             float& yScreenPointOut,
             double& xValueOut,
             double& yValueOut,
             double& rhoValueOut,
             double& thetaValueOut,
-            double& tValueOut)
+            double& tValueOutt)
```
joseartrivera, line 51: nit: typo?

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
quentin987: done, is there a "best-practices" for what to include/not include in the pch.h files?

src/GraphingImpl/Mocks/GraphRenderer.h
quentin987: good catch :)

src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml.cpp
joseartrivera: The style of the project is if its a standard header library it goes into the pch.h, otherwise can go in the regular .h

quentin987: awesome, thanks.

PR #1253, opened 2020-05-29
Fixing TestHistoryClearCommand Test

quentin987:
We removed a bunch of unused history code and persistence settings logic which these tests were calling into their own implementation and consuming. This was causing a failing test _**TestHistoryClearCommand**_  in x86, however if you were to wipe out your settings file, you can't repro the test failure.

Essentially this is test debt at this point which causes failed runs with certain settings files.

Automated testing


mcooley:
@quentin987 it looks like the build agents were updated to VS 16.6, which gets our x64 tests running in CI again. It looks like we have some other problems with x64 tests not passing--I'll investigate those. This PR is OK to merge even though the x64 test pass isn't completing successfully.

PR #1257, opened 2020-06-02
Delay load graph engine dll

joseartrivera:
- Delay load graph engine DLLs until the code needs them.

- Manual tests, checking modules in VS to verify that the omsaut.dll and graphengine.dll are loaded only when the user enters graphing mode


mcooley:
@joseartrivera code looks good to me. Do we have before/after perf numbers for this change?

joseartrivera:
> @joseartrivera code looks good to me. Do we have before/after perf numbers for this change?

Last I checked the change in perf is not perceivable on my machine but I think this is a good change to have anyway for slower machines and reducing the amount loaded in memory.

mcooley:
OK, I figured the launch latency improvement would be small. Working set reduction seems like a worthy goal.

PR #1261, opened 2020-06-02
Modify how we manage round corners

rudyhuyn:
- Modify how we manage corner radius for trig modes and make it compatible with high-contrast

image

image

- Manually


mcooley:
Code looks OK. What's the reason for this change?

PR #1302, opened 2020-07-07
Narrator does not provide the confirmation when user deleted one history from the history list items #1172

### LINKED ISSUE #1172: Narrator does not provide the confirmation when user deleted one history from the history list items
**Repro Steps:** 
1. Launch 'Calculator'.
2. Home screen will be displayed.​
3. Navigate through Menu using keyboard and Choose either 'Standard' or 'Scientific Calculator'.
4. Navigate to History button and activate it.
5. Navigate to any list item and press context menu button.
6. Context menu will open.
7. Navigate to delete button and activate it. 
8. Observe the issue. 

**Actual Result:**
Narrator does not provide the confirmation when user deleted one history from the history list items.

**Expected Result:**
Narrator should provide the confirmation when user deleted one history from the history list items.

vishegn:
- Added a new resource string for History slot cleared.
- Created a new narrator announcement method which will handle history slot cleared announcements.
- In HistoryViewModel, added code to DeleteItem() method.  This code will construct the announcement string for history item deletions.
- By setting the HistoryAnnouncement variable, the string is announced by Narrator.

- Tested the changes manually.  Narrator announces 'History slot # cleared' when a history is deleted.
-
-

### DIFF
--- a/src/CalcViewModel/HistoryViewModel.cpp
+++ b/src/CalcViewModel/HistoryViewModel.cpp
@@ -131,6 +131,7 @@ void HistoryViewModel::DeleteItem(_In_ HistoryItemViewModel ^ e)
             RaisePropertyChanged(ItemsSizeKey);
         }
     }
+    // Adding 1 to the history item index to provide 1-based numbering on announcements.
     wstring localizedIndex = to_wstring(itemIndex + 1);
     LocalizationSettings::GetInstance().LocalizeDisplayValue(&localizedIndex);
     m_localizedHistorySlotCleared = AppResourceProvider::GetInstance()->GetResourceString(HistoryResourceKeys::HistorySlotCleared);
--- a/src/CalcViewModel/HistoryViewModel.cpp
+++ b/src/CalcViewModel/HistoryViewModel.cpp
@@ -23,14 +23,16 @@ using namespace Windows::Foundation::Collections;
 static StringReference HistoryVectorLengthKey{ L"HistoryVectorLength" };
 static StringReference ItemsSizeKey{ L"ItemsCount" };
 
-namespace CalculatorApp::ViewModel::HistoryResourceKeys
+namespace HistoryResourceKeys
 {
     StringReference HistoryCleared(L"HistoryList_Cleared");
+    StringReference HistorySlotCleared(L"Format_HistorySlotCleared");
 }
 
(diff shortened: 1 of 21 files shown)


src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4715,4 +4715,8 @@
     <value>Memory list</value>
     <comment>Automation name for the group of controls for memory list.</comment>
   </data>
+  <data name="Format_HistorySlotCleared" xml:space="preserve">
+    <value>History slot %1 cleared</value>
```
sanderl, line 4719: Consider changing slot to item.

vishegn: hi @sanderl.  I left it as 'slot' to have parity with Memory item announcements.  If you feel 'item' suits better I will change it.

src/CalcViewModel/HistoryViewModel.cpp
```diff
@@ -129,6 +131,11 @@ void HistoryViewModel::DeleteItem(_In_ HistoryItemViewModel ^ e)
             RaisePropertyChanged(ItemsSizeKey);
         }
     }
+    wstring localizedIndex = to_wstring(itemIndex + 1);
```
quentin987, line 134: Can you add a comment prefixing this line on why we're using itemIndex + 1?

src/Calculator/Resources/en-US/Resources.resw
quentin987: Makes sense to be synced with memory to stay consistent, will mark this thread as resolved with the current changes.

quentin987:
Testing looks really good! One line needs to be updated for developer documentation and I can approve and merge. Thanks again!

src/CalcViewModel/HistoryViewModel.cpp
vishegn: Added a small comment to indicate the motivation behind the change.  Essentially, I wanted to move away from zero based numbering to 1-based so it would provide a better experience for end users.   Please let me know if this is in the right direction.

quentin987: Great! This makes total sense.

vishegn:
> There appears to be some conflicts with an existing merge, can you address those and I can approve?
 Done.  Please do the needful.  Thanks.

Issue #1310, opened 2020-07-11
A crash occurs when running graph mode.


Lee-WonJun:
**Describe the bug**

I am trying to fix #1308 now.
But when I try to switch the mode to graph mode, I get a null pointer exception error.

**Steps To Reproduce**

 1. Clone this project
 2. Change Build Mode to x86
 3. Build and run

**Expected behavior**
Runs normally.

**Screenshots**

![image](https://user-images.githubusercontent.com/10369528/87227368-706f1580-c3d5-11ea-8097-c843a80903fc.png)
Note) RenderMain::CanRenderPoint()

**Device and Application Information**

 - OS Build: 10.0.19041.0
 - Architecture: X64
 - Application Version: 10.2005.23.0
 - Region: ko-KR
 - Dev Version Installed: True

**Additional context**

I'm not sure if I'm setting it up incorrectly or a code error.
Also, my architecture is x64, but when I try to debug in x64, Debug mode, as soon as I start it, I get an error in xmemory. (It seems to be a problem with stl, not a calculator.)
So I tested it with x86. 

**Requested Assignment**

If this is fixed by simply adding a null check, I would like to fix this.
If it is not an error, or if it is an architectural problem, I just want to report it.

Chips1234:
It also does it with x64

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

sanderl:
@Lee-WonJun, thanks for reporting this issue. I am looking into this. I don't think the fix is likely to be a null check.

sanderl:
@Lee-WonJun, I take my null check statement back, it can be resolved with a null check at the beginning of the CanRenderPoint(). You are welcome to fix it if you would like. You just need to return false in the case that m_graph is null.

Lee-WonJun:
> 
> 
> @Lee-WonJun, I take my null check statement back, it can be resolved with a null check at the beginning of the CanRenderPoint(). You are welcome to fix it if you would like. You just need to return false in the case that m_graph is null.

Thanks for the answer!
I'll create a pull request

Chips1234:
Thanks, @Lee-WonJun!

PR #1312, opened 2020-07-12
Change the initialization time of Radio button's isChecked

### LINKED ISSUE #1303: Graph mode theme is changed after running programmer mode.
**Describe the bug**
Graph mode theme is changed to Dark form White after running programmer mode.

**Steps To Reproduce**
0. Set Windows Theme to Dark mode
1. Run Calculator
2. Change mode to Graph
3. Change theme to same of windows theme
4. Close Calculator
5. Run Calculator
6. Change theme to withe
7. Change mode to programmer
8. Change mode to Graph

**Expected behavior**
1. I expect to maintain theme in Graph mode

**Screenshots**

ezgif com-video-to-gif

 - OS Build: 10.0.19041.0
 - Architecture: X64
 - Application Version: 10.2005.23.0
 - Region: ko-KR
 - Dev Version Installed: True

**Additional context**
I Find this bug in  Calculator in Windows Store.
And I try to find this bug in Dev Version, but now my notebook's Windows Version isn't 2004.
So I tried to check in Dev mode using mstsc. but error occurs because m_graph is null (now, I don't know why this is happening, is it a bug, what am I forgetting?), so I took a temporary action. Subsequently, the same problem occurs in Dev mode.

**Requested Assignment**

Lee-WonJun:
Change the initialization time of Radio button's isChecked in CalculatorProgrammerRadixOperators

- Remove checkDefaultBitShift
- Add a member variable for saving what is selected
- Modify XAML

- Manual tests / Check that graph theme is unchanged
- Default Unit Tests (Unit Test pass, UI Test Fail, but it seems to be a bug because the master branch fails too )

### Information
As a result of analyzing the call point of g.hpp made from C++, it was confirmed that the theme radio button in Graphing Setting is changed suddenly when CalcuatorProgrammerRadixOperators is initialized.
It was determined that there was something wrong with the components of the CalacuatorProgrammerRadixOperators, and after various tests, it was confirmed that there was a problem in the initialization of the Radio Button (BitshiftFlyout).

The problem I am assuming is as follows.
1. The radio box group does not exist.
2. Arithmetic Shift Button of Bitshift Flyout Radio Box is checked at startup.
4. When the view is converted and the CacluatorPrgrammerRadixOperators are created, the Graph Setting does not disappear and remains in the memory (Graph Setting seems to be generated by Code-base, not XAML).
5. Since other Radio Button is selected, Radio Button of Graphing Setting is Unchecked.

- Excepted
image

- Error
image

I chose a simple way to check the radio button when displaying the BitShiftFlyout.

So, I added a member variable for saving what is selected. (I just use the reference of radio directly, not use others like enum. this is dangerous?)

I don't think this is the best way. I think this is not a fundamental solution
However, to solve the problem without structural changes (or changing the internal operation), I selected this way.

### Additional context
#1310 issue has not been answered yet and has been excluded from the PR.

### DIFF
--- a/src/CalcViewModel/ApplicationViewModel.cpp
+++ b/src/CalcViewModel/ApplicationViewModel.cpp
@@ -125,14 +125,6 @@ bool ApplicationViewModel::TryRecoverFromNavigationModeFailure()
 void ApplicationViewModel::OnModeChanged()
 {
     assert(NavCategory::IsValidViewMode(m_mode));
-    if (NavCategory::IsValidViewMode(m_PreviousMode) && NavCategory::IsCalculatorViewMode(m_PreviousMode) && !NavCategory::IsCalculatorViewMode(m_mode))
-    {
-        if (m_CalculatorViewModel)
-        {
-            m_CalculatorViewModel->SetCalculatorType(m_mode);
-        }
-    }
-
     if (NavCategory::IsCalculatorViewMode(m_mode))
     {
         if (!m_CalculatorViewModel)
--- a/src/CalcViewModel/ApplicationViewModel.cpp
+++ b/src/CalcViewModel/ApplicationViewModel.cpp
@@ -125,6 +125,14 @@ bool ApplicationViewModel::TryRecoverFromNavigationModeFailure()
 void ApplicationViewModel::OnModeChanged()
 {
     assert(NavCategory::IsValidViewMode(m_mode));
+    if (NavCategory::IsValidViewMode(m_PreviousMode) && NavCategory::IsCalculatorViewMode(m_PreviousMode) && !NavCategory::IsCalculatorViewMode(m_mode))
+    {
+        if (m_CalculatorViewModel)
+        {
+            m_CalculatorViewModel->SetCalculatorType(m_mode);
+        }
+    }
+
     if (NavCategory::IsCalculatorViewMode(m_mode))
     {
         if (!m_CalculatorViewModel)
--- a/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
+++ b/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
@@ -47,7 +47,12 @@ void CalculatorProgrammerRadixOperators::FlyoutButton_Clicked(_In_ Platform::Obj
 
 void CalculatorProgrammerRadixOperators::checkDefaultBitShift()
 {
+    CollapseBitshiftButtons();
     m_selectedShiftButtonMode = BitShiftMode::Arithmetic;
+    LshButton->Visibility = ::Visibility::Visible;
+    RshButton->Visibility = ::Visibility::Visible;
+    LshButton->IsEnabled = true;
+    RshButton->IsEnabled = true;
 }
 
 void CalculatorProgrammerRadixOperators::BitshiftFlyout_Checked(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
--- a/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
+++ b/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
@@ -38,16 +38,12 @@ void CalculatorProgrammerRadixOperators::LoadResourceStrings()
     m_logicalShiftButtonContent = resProvider->GetResourceString(L"logicalShiftButtonSelected");
     m_rotateCircularButtonContent = resProvider->GetResourceString(L"rotateCircularButtonSelected");
     m_rotateCarryShiftButtonContent = resProvider->GetResourceString(L"rotateCarryShiftButtonSelected");
+    m_selectedShiftButtonContent = m_arithmeticShiftButtonContent;
 }
 
 void CalculatorProgrammerRadixOperators::FlyoutButton_Clicked(_In_ Platform::Object ^ /*sender*/, _In_ Windows::UI::Xaml::RoutedEventArgs ^ /*e*/)
 {
     this->BitwiseFlyout->Hide();
-}
-
-void CalculatorProgrammerRadixOperators::checkDefaultBitShift()
-{
-    this->ArithmeticShiftButton->IsChecked = true;
 }
 
 void CalculatorProgrammerRadixOperators::BitshiftFlyout_Checked(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
@@ -109,7 +105,12 @@ void CalculatorProgrammerRadixOperators::BitshiftFlyout_Checked(Platform::Object
         announcementString = m_rotateCarryShiftButtonContent;
     }
 
-    this->BitShiftFlyout->Hide();
+    if (announcementString != m_selectedShiftButtonContent)
+    {
+        this->BitShiftFlyout->Hide();
+        m_selectedShiftButtonContent = announcementString;
+    }
+
     Model->SetBitshiftRadioButtonCheckedAnnouncement(announcementString);
 }
 
@@ -174,5 +175,25 @@ void CalculatorProgrammerRadixOperators::ClearButton_LostFocus(Object ^ sender, 
     if (ClearEntryButton->Visibility == ::Visibility::Visible && ClearButton->Visibility == ::Visibility::Collapsed)
     {
         ClearEntryButton->Focus(::FocusState::Programmatic);
+    }
+}
+
+void CalculatorApp::CalculatorProgrammerRadixOperators::BitShiftFlyout_Opened(Platform::Object ^ sender, Platform::Object ^ e)
+{
+    if (m_selectedShiftButtonContent == m_arithmeticShiftButtonContent)
+    {
+        ArithmeticShiftButton->IsChecked = true;
+    }
+    else if (m_selectedShiftButtonContent == m_logicalShiftButtonContent)
+    {
+        LogicalShiftButton->IsChecked = true;
+    }
+    else if (m_selectedShiftButtonContent == m_rotateCircularButtonContent)
+    {
+        RotateCircularButton->IsChecked = true;
+    }
+    else if (m_selectedShiftButtonContent == m_rotateCarryShiftButtonContent)
+    {
+        RotateCarryShiftButton->IsChecked = true;
     }
 }
--- a/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
+++ b/src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
@@ -47,7 +47,15 @@ void CalculatorProgrammerRadixOperators::FlyoutButton_Clicked(_In_ Platform::Obj
 
 void CalculatorProgrammerRadixOperators::checkDefaultBitShift()
 {
+    LoadDeferredLoadButtons();
+
+    if (IsButtonLoaded())
+    {
+        return;
+    }
+
     CollapseBitshiftButtons();
+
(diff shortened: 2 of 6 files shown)


Lee-WonJun:
The code was modified as the master branch was changed.
I added a string type, instead of reference type, member variable for saving what is selected. Because some string member variable is in the master branch now.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -176,3 +177,23 @@ void CalculatorProgrammerRadixOperators::ClearButton_LostFocus(Object ^ sender,
         ClearEntryButton->Focus(::FocusState::Programmatic);
     }
 }
+
+void CalculatorApp::CalculatorProgrammerRadixOperators::BitShiftFlyout_Opened(Platform::Object ^ sender, Platform::Object ^ e)
+{
+    if (m_selectedShiftButtonContent == m_arithmeticShiftButtonContent)
```
joseartrivera, line 183: Its by design that arithmetic shift is selected when switching away & back from other modes, this no longer happens with the graphing mode, but does with other modes, not sure why this occurs but it seems related to the root cause of the issue here.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -38,18 +38,14 @@ void CalculatorProgrammerRadixOperators::LoadResourceStrings()
     m_logicalShiftButtonContent = resProvider->GetResourceString(L"logicalShiftButtonSelected");
     m_rotateCircularButtonContent = resProvider->GetResourceString(L"rotateCircularButtonSelected");
     m_rotateCarryShiftButtonContent = resProvider->GetResourceString(L"rotateCarryShiftButtonSelected");
+    m_selectedShiftButtonContent = m_arithmeticShiftButtonContent;
```
joseartrivera, line 41: Lets move this out of this function since this isn't a resource string

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -109,7 +105,12 @@ void CalculatorProgrammerRadixOperators::BitshiftFlyout_Checked(Platform::Object
         announcementString = m_rotateCarryShiftButtonContent;
     }
 
-    this->BitShiftFlyout->Hide();
+    if (announcementString != m_selectedShiftButtonContent)
+    {
+        this->BitShiftFlyout->Hide();
+        m_selectedShiftButtonContent = announcementString;
```
joseartrivera, line 111: I like the idea of using an enum instead of the string content here

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -109,7 +105,12 @@ void CalculatorProgrammerRadixOperators::BitshiftFlyout_Checked(Platform::Object
         announcementString = m_rotateCarryShiftButtonContent;
     }
 
-    this->BitShiftFlyout->Hide();
+    if (announcementString != m_selectedShiftButtonContent)
```
joseartrivera, line 108: I like the new behavior of not closing when the same button is selected. 👍

Lee-WonJun: thanks!

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
Lee-WonJun: OK!

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
Lee-WonJun: I'll roll back the checkDefaultBitShift function I deleted.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
Lee-WonJun: I'll roll back the check Default BitShift function and modify its implementation.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
joseartrivera: I still see the same issue after your change.

To be more specific here is what I am seeing:
1. Go to programmer mode, change bit shift radio button to be logical
2. Go to graphing mode
3. Go back to programmer mode, open the bit shift flyout and see that the radio button is still on logical.

Repeat the above steps with any other mode in Calculator and you will see that the bit shift flyout will reset back to arithmetic shift.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -47,7 +47,12 @@ void CalculatorProgrammerRadixOperators::FlyoutButton_Clicked(_In_ Platform::Obj
 
 void CalculatorProgrammerRadixOperators::checkDefaultBitShift()
```
Lee-WonJun, line 48: In previous versions, there was an issue where button availability was reset when opening BitShift Flyout not when the mode is changed

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -47,7 +47,12 @@ void CalculatorProgrammerRadixOperators::FlyoutButton_Clicked(_In_ Platform::Obj
...
 {
-    this->ArithmeticShiftButton->IsChecked = true;
+    CollapseBitshiftButtons();
+    m_selectedShiftButtonMode = BitShiftMode::Arithmetic;
+    LshButton->Visibility = ::Visibility::Visible;
+    RshButton->Visibility = ::Visibility::Visible;
+    LshButton->IsEnabled = true;
+    RshButton->IsEnabled = true;
 }
```
Lee-WonJun, line 56: In previous versions, there was an issue where button availability was reset when opening BitShift Flyout not when the mode is changed

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
Lee-WonJun: I understand what you say, but, in my opinion, this issue is not about my first issue.
I think this problem is about calling callback in Calculator ViewModel.
I'll comment more detail in related commit

src/CalcViewModel/ApplicationViewModel.cpp
```diff
@@ -125,6 +125,14 @@ bool ApplicationViewModel::TryRecoverFromNavigationModeFailure()
...
 {
     assert(NavCategory::IsValidViewMode(m_mode));
+    if (NavCategory::IsValidViewMode(m_PreviousMode) && NavCategory::IsCalculatorViewMode(m_PreviousMode) && !NavCategory::IsCalculatorViewMode(m_mode))
+    {
+        if (m_CalculatorViewModel)
+        {
+            m_CalculatorViewModel->SetCalculatorType(m_mode);
+        }
+    }
+
```
Lee-WonJun, line 135: if you switch mode in Calculator (Programmer, Scientific, Standard) then the callback about mode (in this case "OperatorsPanel::EnsureProgrammerRadixOps()" is called.  But if you switch mode, not in Calculator, then the callback is not called because mode in Calculator is not changed. so I control the previous mode.

In my memory, Calculator value and the setting isn't changed if you switch mode, not in Calculator

but If you apply this, Calculator value and the setting is changed no matter which the previous mode you selected even it is not graph mode.

To apply only graph and calculator mode, we can change this "if statement". But I don't know what's right, please reply, Thank you!

-Previous Version
previous-version

-Now
now

joseartrivera: Sorry, I don't think I was clear enough in my comment. I wasn't talking about the result or the state of the whole calculator resetting, I am talking about just the state of the BitShift flyout demonstrated in this GIF:

flyout

However, after trying it out locally I noticed this still happens in master **before** your changes so it actually isn't an issue with this PR. Sorry for sidetracking you.

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
joseartrivera: I just saw this before realizing it myself, you are right it is a pre-existing problem.

src/CalcViewModel/ApplicationViewModel.cpp
Lee-WonJun: I have reverted commit about this, Thanks for the comment!

src/Calculator/Views/CalculatorProgrammerRadixOperators.xaml.cpp
```diff
@@ -47,7 +47,12 @@ void CalculatorProgrammerRadixOperators::FlyoutButton_Clicked(_In_ Platform::Obj
 
 void CalculatorProgrammerRadixOperators::checkDefaultBitShift()
 {
-    this->ArithmeticShiftButton->IsChecked = true;
```
joseartrivera, line 50: I don't think we need to call CollapseBitShiftButtons or set visibility here, it may cause a crash since some of those buttons may not be loaded yet at this point since they defer their loading. I think the only code we need here is this line:

`m_selectedShiftButtonMode = BitShiftMode::Arithmetic;`

Unless I am missing something else here that this is working around?

Lee-WonJun: This function is also called when the mode is changed in the calculator.
When the flyout menu is opened due to delaying initialization, the related button is initialized. but the menu does not open, then the related button is the same as the previous state

Lee-WonJun: Now I think about it, it just can be replaced to "ArithmeticShiftButton->IsChecked = true;" because it is called when mode changed in calc.
I'll check it

joseartrivera: Okay yeah you are right, I can see that when using the shortcut keys to access them. Can we factor out the logic to check if these buttons are loaded into a helper function and check that before calling CollapseBitShiftButtons just to double check that the buttons are initialized?

```    // Load deferred load buttons
    if (RolButton == nullptr)
    {
        FindName("RolButton");
        FindName("RorButton");
        FindName("RolCarryButton");
        FindName("RorCarryButton");
        FindName("LshLogicalButton");
        FindName("RshLogicalButton");
    }

    // Since arithmeticShiftButton defaults to IsChecked = true, this event an fire before we can load the deferred loaded controls. If that is the case, just
    // return and do nothing.
    if (RolButton == nullptr || RorButton == nullptr || RolCarryButton == nullptr || RorCarryButton == nullptr || LshLogicalButton == nullptr
        || RshLogicalButton == nullptr)
    {
        return;
    }

Lee-WonJun: as you said,
I added a check before calling Collapse Bit Shift Buttons.
However, the same code was duplicated in "checkDefaultBitShift" and "BitshiftFlyout_Checked". So I extracted as a function. Does it look okay?

joseartrivera: Perfect, thank you for the fix!

PR #1334, opened 2020-08-05
Excessive check for the max value is removed

### LINKED ISSUE #1301: Copy-pasting result computed by the calculator gives "Invalid input"
Calculator 10.1910.0.0

Step 1.
image

Step 2.
image

Step 3.
image

Step 4.
image

Why the input is invalid if the calculator has computed it just before??

JackRainy:
## Fixes #1301 .
It reproduces both for QWORD numbers (as mentioned in the report) and for WORD numbers (e.g. -32768).

When user is trying to insert quite a big negative number, the validator `CopyPasteManager::ExpressionRegExMatch` calculates both the maximum bits length and the maximum unsigned long value for it.
However, casting signed negative long to the unsigned long by this check leads to exceeding of the maximum value.

I have removed the check for the maximum value since there are already precedence checks for the maximum bits length `OperandLength(...) > maxOperandLength` and `TryOperandToULL(...) != nullptr` which throw the error if "the operand was empty, received invalid_argument, or received out_of_range".

- The changes were tested by manual testing on 4 types of numbers: QWORD, DWORD, WORD, BYTE.
- Additional unit tests have been written to cover these scenarios. Now there are automatic tests for the copy/paste context menu which were absent by the moment.

### DIFF
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -331,12 +331,6 @@ bool CopyPasteManager::ExpressionRegExMatch(
                     expMatched = false;
                     break;
                 }
-
-                if (operandAsULL->Value > maxOperandLengthAndValue.maxValue)
-                {
-                    expMatched = false;
-                    break;
-                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


JackRainy:
This solution did not work as expected, I am going to return with something better that this one.

PR #1336, opened 2020-08-06
Take the boundary of signed negative values into account( Fix issue 1301)

### LINKED ISSUE #1301: Copy-pasting result computed by the calculator gives "Invalid input"

JackRainy:
## Fixes #1301 

It reproduces both for QWORD numbers (as mentioned in the report) and for WORD numbers (e.g. -32768).

When user is trying to insert quite a big negative number, the validator `CopyPasteManager::ExpressionRegExMatch` calculates both the maximum bits length and the maximum unsigned long value for it.
However, the sign of a signed negative long is ignored by `SanitizeOperand` thus it results in casting of the signed long to the unsigned long with the overflow.

The check for the maximum value is improved by additional check whether it has a deal with the maximum negative number for the given word size. Now the range of acceptable values is extended (e.g. for WORD from [-32767...32767] to [-32768...32767] )

## How changes were validated:
- The changes were tested by manual testing on 4 types of numbers: QWORD, DWORD, WORD, BYTE.
- Additional unit tests have been written to cover these scenarios.
- Also, now there are automatic UI tests for the copy/paste context menu which were absent by the moment.

### DIFF
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -311,7 +311,7 @@ bool CopyPasteManager::ExpressionRegExMatch(
         if (operandMatched)
         {
             // Remember the sign of the operand
-            bool isNegativeValue = operand->Data()[0] == L'-' || programmerNumberBase != NumberBase::DecBase;
+            bool isNegativeValue = operand->Data()[0] == L'-';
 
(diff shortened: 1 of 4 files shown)


EriWong (approved):
Aplogies for the delay in this review.  The changes look good and I really appreciate the tests!

Issue #582, opened 2019-07-13
Add link to GitHub repo on about flyout


grochocki:
**Problem Statement**
There is no way to navigate to the project from within the Calculator app.

**Evidence or User Insights**

**Proposal**
By adding a link to the GitHub repo on the about flyout, people will know that it's open source and that they can create issues, PRs and more.

**Goals**
Improve discoverability of the GitHub project from within Calculator.

**Non-Goals**

**Low-Fidelity Concept**
Add this string to the about dialog after the "Send feedback" button:
> To learn how you can contribute to Windows Calculator, check out the project on [GitHub](https://go.microsoft.com/fwlink/?linkid=2099939).

![image](https://user-images.githubusercontent.com/1414423/62022535-a776f780-b181-11e9-9c2b-49e0a40f4409.png)

**Requested Assignment**
I'm just suggesting this idea.  I don't want to implement it.

> Note: Copied/split from #543

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

Poopooracoocoo:
Would it be preferred if people submitted issues on GitHub or on the Feedback Hub?

grochocki:
If you want to quickly submit feedback, use Feedback Hub, but if you want to contribute to developing the idea further, submit it on GitHub. For some new features, Feedback Hub is a great source for evidence that the feature addresses a user need. We regularly look at top feedback when making product decisions. At the same time, there are lots of small changes, like this one, that would likely never make it to the top of Feedback Hub top feature request lists.

Poopooracoocoo:
I feel that GitHub is better overall. We can see it being triaged, have an actual conversation and we can see issues from across the world.

grochocki:
We reviewed the pitch and would love to move forward with it. Since the idea is fairly straightforward and is entirely captured by the pitch, we can move this straight into production.

rudyhuyn:
> Windows Calculator is made available under the MIT license. To learn more and contribute, check out the project on _**GitHub**_.

Interesting control to code.

@grochocki: 
Tell me what you think about these specifications?
- Only the word "GitHub" is clickable.
- "GitHub" should be part of the textblock and support text-wrapping (which means we can't use `<StackPanel />`)
- The hyperlink is not at the end of the sentence in some languages 
[...]

Issue #893, opened 2019-12-26
Rewrite or have a parallel calculator project in C#


PaulDotNet:
Calculator is a great project for developers starting with UWP. However, it is implemented in C++ with managed extensions which I would consider rather a niche language with a niche features in a niche area of UI development. It is a niche within a niche within a niche.

I know C++ but would never consider this language for UI development. This ship has sailed around year 2000. On the other hand if Microsoft would pay me a huge salary to develop UI apps in C++ I would seriously consider it. Then I would say no. 

Having calculator project in  C# would make this project accessible to a much wider audience and maybe would help adoption of UWP somewhat. 

Please consider converting this project to C#. Future generations would thank you for it.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

grochocki:
Thanks for the submission, @PaulDotNet! To be honest, this is something we had already been discussing internally, but it is great to see that there is community interest in this as well. While we agree that migrating the codebase to C# would offers lots of benefits, porting is no trivial task, so we'll want to consider the pros and cons and make sure we have a plan that minimizes disruptions to the project and the community. Moving to next stage to investigate some of these details and assigning to @rudyhuyn to help drive discussion.

mdtauk:
I think because this project originated within the Windows OS code base, and probably relies on code from the previous Win32 code - this is why it is as it is.

The team maintaining the project would be well versed in C++ but it does make it more difficult for others wishing to contribute.

If the core C++ project had an ability to "plug-in" C# modules, then adding new features becomes easier.  But then if you decide to go that way in refactoring, what are the Pros and Cons to moving everything to C#?

WinUI 3.0 also allows the breaking out of the sandbox.  Does the project remain a UWP app?  If C# WinUI 3.0 relies on using a .NET Core framework, does the app want to take a dependency on it?

That is without considering if there is merit to keeping core code in a cross platform compatible C language, with an option to build a MacOS or Linux UI fork, which the community could work on?

Titaniumtown:
I would love to see the windows 10 calculator on linux, hope that becomes a thing in the near future.

stevenbrix:
FWIW, the Uno platform ported the calculator to C# already:
https://github.com/unoplatform/calculator

Rahul-Dixit:
I have a question in my mind: won't this impact the performance? C++ binaries are much faster and lightweight as compared to those created using C#.

[...]

PR #1343, opened 2020-08-08
Add settings

### LINKED ISSUE #596: Replace current about dialog with settings page
**Problem Statement**
Calculator currently does not have a settings page, which will block future work. The current about dialog is also inconsistent with many other Microsoft apps.

**Evidence or User Insights**
Many suggested features have suggested adding options in settings (#538, #486, #485, #393, #382, #215), and while we have not accepted all of these pitches, it is inevitable that there will be functionality requiring settings UX.

**Proposal**
Replace current about dialog with settings page. 

**Goals**
* Replace current about dialog with settings page
* Be consistent with settings UX patterns established in other Microsoft apps
* Add basic theme selection option common across many other apps

**Non-Goals**
* Add any additional options at this time

**Low-Fidelity Concept**
image

**Requested Assignment**

Chips1234:
- Add in a new page, the "Settings" page.
- Changed the NavView footer to "Settings".
- Replaced the "About" flyout with the "Settings" page.

Calculator settings NavBar
_Change the "About" button in the Navigation Bar to "Settings"._

Calculator settings
_The new "Settings Page"_

Calculator settings 2
_When the user changes the mode, they are prompted to restart the app for changes to take effect._ 

Calculator settings 3
_When low on space._

Manual

P.S. thanks @sanderl for all the help!


Chips1234:
Based on [this document](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.application.requestedtheme?view=winrt-19041#remarks):
>The theme can only be set when the app is started, not while it's running. Attempting to set RequestedTheme while the app is running throws an exception (NotSupportedException for Microsoft .NET code). If you give the user an option to pick a theme that's part of app UI, you must save the setting in the app data and apply it when the app is restarted. (For more info about app settings, see Store and retrieve settings and other app data).
Do we need to have the app restart?

Chips1234:
Checks are taking a while...

Chips1234:
The map app requires a restart:
map theme change

grochocki:
> Checks are taking a while...

I don't think draft PRs trigger CI builds.

> Do we need to have the app restart?

Let's use the same string as Maps (and other apps like Movies & TV and Photos): "You'll see your change the next time you start the app."

Also, I noticed in your screenshot that the "About this app" section is pinned to the bottom. Can you update this to match the design spec [here](https://github.com/microsoft/calculator-specs/tree/master/active/settingsPage)?

Chips1234:
Alright

Chips1234:
@grochocki Here are some questions:
* In my latest commit, I added the ability to change the theme but the "System theme" radio button doesn't seem to work, could you maybe help me with that?
* I think that this is a good opportunity to fix #582. Could you tell me how we could work with .resw files to add links?

Thanks.

grochocki:
Will defer to @sanderl on technical questions.

> I think that this is a good opportunity to fix #582.

In general, we try to keep independent changes as separate PRs, so lets tackle this after the settings page is in.

Chips1234:
OK

Chips1234:
@sanderl? Thank you

sanderl:
@Chips1234, you'll want to use the RequestedTheme and save it to AppData. You can look at the GraphControl theme for a starting point: void GraphingSettings::SetGraphTheme(bool isMatchAppTheme)

Chips1234:
Okay. Thanks

Chips1234:
@sanderl could you explain to me how to get the system default theme? Thanks

Administaror:
Oh.really?

Administaror:
Oh Thinks

Chips1234:
@sanderl?

sanderl:
> @sanderl could you explain to me how to get the system default theme? And also, how do I make sure "Windows Default" is selected as default? Thanks

@Chips1234, sorry for the delayed response. I believe you just need to clear the RequestedTheme and it should follow the default.

Chips1234:
How should I clear it? @sanderl

sanderl:
I think you can set it back to null.

Chips1234:
Okay, thank you @sanderl

Rahul-Dixit:
> ## Fixes #596 .
> 
> ### Description of the changes:
> - Add in a new NavCategory.
> - Changed the footer to "settings".
> - Replaced about flyout with settings page.
> Calculator settings NavBar
> Calculator settings
> Calculator settings 2
> 
> ### How changes were validated:
> Manual
> 
> P.S. thanks @sanderl for all the help!

@Chips1234 In the screenshots that you've shared, the 'Statement' is spelt as 'Statment' by mistake. Could you please correct it?

Chips1234:
> @Chips1234 In the screenshots that you've shared, 'Statement' is spelt as 'Statment' by mistake. Could you please correct it?

@Rahul-Dixit thanks, solved.

Rahul-Dixit:
@Chips1234 A few more things...

- Don't you think that you've chosen a much bigger heading level for 'App Theme' and 'About this app' than what was specified in the spec. In the spec, they have chosen a smaller heading level with the bold attribute.
- Could you please make the 'S' in 'Windows color Settings' lowercase for consistency with the spec and other Microsoft apps.

Thanks. The rest is good to go!

Rahul-Dixit:
@Chips1234 Can you also update the Version info to match the most recent version of this app?

Chips1234:
> @Chips1234 Can you also update the Version info to match the most recent version of this app?

@Rahul-Dixit the version string is set automatically

Rahul-Dixit:
@Chips1234 OK. Why is the string 'You will see your changes the next time you start the app' below the heading 'App theme'?

Just have a look at the screenshot of the Maps app.

Chips1234:
The version string is set automatically

________________________________

Chips1234:
Look at the one for Movies and TV’s. They’re on top of all the radio buttons, and I think they look better that way

Rahul-Dixit:
Right you are! There is a lot of inconsistency among these apps.

Chips1234:
Yes, I’m sending feedback to maps right now via feedback hub

________________________________

Chips1234:
@grochocki what heading level and style did you use for your spec?

Chips1234:
@Rahul-Dixit I live in Canada, so I spelt it tu Canadian way. However, as the resource file is en-US, I will correct it.

Rahul-Dixit:
> @Rahul-Dixit I live in Canada, so I spelt it tu Canadian way. However, as the resource file is en-US, I will correct it.

@Chips1234 I think it's fine. This spelling is more widely accepted. See the License at the root of this repository. No need to change anything.

Chips1234:
Okay

________________________________

Chips1234:
@grochocki what heading level and style should I use?

Rahul-Dixit:
@Chips1234 In the meanwhile, you can check why the unit tests are failing.

Chips1234:
@Rahul-Dixit I don't get why it has failed

Rahul-Dixit:
@Chips1234 There is some problem with the test Deserialize_AllInvalid.

Chips1234:
@Rahul-Dixit So what I should do to fix this is...

Rahul-Dixit (changes requested):
@Chips1234, you did not update the tests after adding the Settings page, and that is why the unit tests are failing!

I think that with the Settings page, the number of ViewModes has increased, so you need to make some changes in [``src/CalculatorUnitTests/NavCategoryUnitTests.cpp``](https://github.com/microsoft/calculator/blob/master/src/CalculatorUnitTests/NavCategoryUnitTests.cpp):

In `void NavCategoryUnitTests::Deserialize_AllInvalid()`,
```c++
VERIFY_ARE_EQUAL(ViewMode::None, NavCategory::Deserialize(ref new Box<int>(18)));
```
Change this to 19, as the number of ViewModes has increased.

In `void NavCategoryUnitTests::IsValidViewMode_AllValid()`, append the following line
```c++
VERIFY_IS_TRUE(NavCategory::IsValidViewMode(ViewMode::Settings));
```

In `void NavCategoryUnitTests::IsValidViewMode_AllInvalid()`,
```c++
if (Windows::Foundation::Metadata::ApiInformation::IsMethodPresent("Windows.UI.Text.RichEditTextDocument", "GetMath"))
{
    // There are 19 total options so int 19 should be the first invalid
     VERIFY_IS_TRUE(NavCategory::IsValidViewMode(static_cast<ViewMode>(18)));
     VERIFY_IS_FALSE(NavCategory::IsValidViewMode(static_cast<ViewMode>(19)));
}
else
{
     // There are 18 total options when graphing calculator is not present, so int 18 should be the first invalid
     VERIFY_IS_TRUE(NavCategory::IsValidViewMode(static_cast<ViewMode>(17)));
     VERIFY_IS_FALSE(NavCategory::IsValidViewMode(static_cast<ViewMode>(18)));
}
```

Apart from these, there are some other changes also that you need to make. You can take help from [here](https://github.com/microsoft/calculator/commit/091732aa944d4d7f0085513f40cf91c1007e85e6). This is the initial commit for the Graphing Calculator. From here, you can refer to the changes made to this file at the time of addition of a new mode.

Chips1234:
Thanks @Rahul-Dixit

Chips1234:
@Rahul-Dixit unit tests seem to be still failing...

Rahul-Dixit:
@Chips1234 Sorry for the delay! The test Deserialize_AllInvalid is now passing. But the tests IsValidViewMode_AllInvalid and IsConverterViewMode are failing.

I'm having a hard time finding the problem. @sanderl Could you please help?

src/CalculatorUnitTests/NavCategoryUnitTests.cpp
```diff
@@ -225,6 +228,7 @@ namespace CalculatorUnitTests
         VERIFY_IS_TRUE(NavCategory::IsConverterViewMode(ViewMode::Data));
         VERIFY_IS_TRUE(NavCategory::IsConverterViewMode(ViewMode::Pressure));
         VERIFY_IS_TRUE(NavCategory::IsConverterViewMode(ViewMode::Angle));
+        VERIFY_IS_TRUE(NavCategory::IsConverterViewMode(ViewMode::Settings));
```
Rahul-Dixit, line 231: I'm not sure about this one.

Chips1234: But the graphing calculator is there

Chips1234:
@grochocki @sanderl anymore suggestions? Apart from th unit tests I believe that this is ready to be merged. Senderl can you please help with the tests?

Rahul-Dixit:
@Chips1234 Could you please update your screenshots?

Chips1234:
Sure.

Rahul-Dixit:
@Chips1234 Could you make 'App theme' and 'About this app' smaller by one heading level? See [this](https://github.com/microsoft/calculator/pull/1343#issuecomment-684172053)

Chips1234:
I tried making the heady style “BaseTextStyle” but it’d be too small. I’m trying to ask grochocki what he use but he isn’t replying

grochocki:
> Why is the string 'You will see your changes the next time you start the app' below the heading 'App theme'?

> Look at the one for Movies and TV’s. They’re on top of all the radio buttons, and I think they look better that way

This string should appear below the radio buttons to avoid content shifting from underneath the mouse, which can be jarring.

> What heading level and style did you use for your spec?

I am reviewing the headings in Figma, and we are using a non-standard type ramp. I will sync with the designer to confirm we want to do this and get back to you with the specific TextBlock styles they recommend using. The current weight is too heavy, so we'll want to make some adjustments.

Chips1234:
Ok thanks

________________________________

Chips1234:
@grochocki any estimate on when the heading level will be done?

Chips1234:
Probably the first screenshot had the window not focused and the second one has the window focused; thus the colour difference

________________________________

Rahul-Dixit:
OK

Chips1234:
@grochocki any updates?

Chips1234:
Alright, thanks!

________________________________

Chips1234:
@grochocki I'm sorry, but what do you mean by "(20,24)"? And am I supposed to turn the "Open Navigation" button into the back button? Thanks

Chips1234:
@grochocki

grochocki:
In parenthesis, those are the font sizes/line heights for reference, but they are the values used in the TextBlock styles (e.g., SubtitleTextBlockStyle takes 20/24 into account).

The Settings page is a forward-navigation to a new page that does not have the NavigationView control. The back button should take the user back to the page they were on previously.

Chips1234:
Thank you. Do you know how we can navigate to a page without the nav bar? How do we navigate to the last page? Thanks

________________________________

grochocki:
\cc @sanderl to help out here

mdtauk:
> Thank you. Do you know how we can navigate to a page without the nav bar? How do we navigate to the last page? Thanks

Maybe collapse the navigation pane when on the Settings page?

Chips1234:
@mdtauk Hmm, that could work. But I'll see what sanderl says. Thanks.

sanderl:
> Thank you. Do you know how we can navigate to a page without the nav bar? How do we navigate to the last page? Thanks

Are you trying to navigate to settings? Or away from Settings? For navigating to settings, users can use the navbar or the shortcut keys. For navigating away from settings, I would think you would want to navigate back to the last used mode.

Chips1234:
@sanderl Sorry, I meant, do we collapse the NavBar when setting page is loaded? How do we navigate back to the previous page? Thanks

sanderl:
> @sanderl Sorry, I meant, do we collapse the NavBar when setting page is loaded? How do we navigate back to the previous page? Thanks

@Chips1234, yes the NavBar should collapse when the settings page is loaded. In general, the NavBar should behave the same way as navigating to any other page.

@grochocki, do you know why the design includes a back button for the settings page? None other pages have a back button like this. Where should the back button navigate to when clicked, the last used mode? Standard mode?

Chips1234:
@sanderl I believe you still misunderstood me there. To show the back button, we first need to hide the show/collapse navigation button, how do we do that? And yes, when the back button is pressed, it should navigate back to the previous page. How do we do that? Thank you!

Chips1234:
@sanderl also, could you help with the UnitTests?

sanderl:
> @sanderl I believe you still misunderstood me there. To show the back button, we first need to hide the show/collapse navigation button, how do we do that? And yes, when the back button is pressed, it should navigate back to the previous page. How do we do that? Thank you!

@Chips1234, you don't actually need to show/hide the back button. The NavMenu will float over the top of that part of the screen. It appears as if it is showing/hiding, but in reality the button can just be always there and the NavMenu will just cover that part of the UI. If you look at Standard Mode, you see how the NavMenu just covers the number pad. I don't think you should need to do any extra work here for making this happen. When the user clicks outside the NavMenu, the NavMenu will close and then you can see the button.

As for the navigation itself, in the case that the Settings page is selected you will need to store the previous mode in a place where the back button can access it. Does that make sense?

For the unit tests, I suspect the tests are failing because of the added page. About used to be a flyout and now it is a page, so it looks like the unit tests need to be updated to account for Settings.

Please let me know if you have more questions.

Chips1234:
@sanderl Thank yo

Chips1234:
@sanderl. One more question, do I store the previous mode in AppData? Or is there a better method to do so?

Chips1234:
@sanderl Also, how do we stop it from lunching into settings mode upon initialization?

sanderl:
@Chips1234, the previous mode should be in AppData, you want to make sure not to overwrite it with Settings. When navigating away from settings, you should just be able to read it. 

I've been really swamped today. I'll have to get back to you on the initialization question later.

Chips1234:
Thank you

Chips1234:
@sanderl. How about the converter? I see all converters have the same entry point, how do I store what mode the converter used to be on with AppData? Thx

ad1Dima:
> The map app requires a restart, however, the XAML controls gallery does not (could it be that it is written in C#?):
> map theme change

May be it's too late. But you can archive dynamic theme change by setting [ElementTheme](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.elementtheme?view=winrt-18362) on RootFrame instead of setting `RequestedTheme` for whole app. But if you do that, you also have to set theme manually for every popup in the app.

mdtauk:
> > The map app requires a restart, however, the XAML controls gallery does not (could it be that it is written in C#?):
> > map theme change
> 
> May be it's too late. But you can archive dynamic theme change by setting [ElementTheme](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.elementtheme?view=winrt-18362) on RootFrame instead of setting `RequestedTheme` for whole app. But if you do that, you also have to set theme manually for every popup in the app.

I think the Map App renders the Map control with dark colours, so re-drawing that probably requires the restart

Chips1234:
> @sanderl. How about the converter? I see all converters have the same entry point, how do I store what mode the converter used to be on with AppData? Also, here's to code I have currently, the app crashed upon clicking the button:
> 
> ```c++
> void SettingsPage::BackButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e, ViewMode /*model*/)
> {
>     auto previousModeValue = Windows::Storage::ApplicationData::Current->LocalSettings->Values->Lookup(L"previousMode");
>     String ^ previousMode = safe_cast<String ^>(previousModeValue);
> 
>     if (previousMode == "Standard")
>     {
>         Model->Mode = ViewMode::Standard;
>     }
> }
> ```
> 
> Thank you

@sanderl Could you answer the above question? Also, what do you think about app restart?

Chips1234:
@ad1Dima It looks like `rootFrame` does not have the ElementTheme member.

ad1Dima:
> @ad1Dima It looks like `rootFrame` does not have the ElementTheme member.

Yes, it called [RequestedTheme](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.frameworkelement.requestedtheme?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(Windows.UI.Xaml.FrameworkElement.RequestedTheme);k(TargetFrameworkMoniker-.NETCore,Version%253Dv5.0);k(DevLang-csharp)%26rd%3Dtrue&view=winrt-19041) either. But application one has type `ApplicationTheme` and FrameworkElement one is ElementTheme

Chips1234:
@ad1Dima Thank you. Could you tell me how one can change the RequestedTheme for RootFrame

ad1Dima:
> @ad1Dima Thank you. Could you tell me how one can change the RequestedTheme for RootFrame?

In C# i did this (AppTheme is my custom enum): 

```cs
internal static void UpdateAppTheme(AppTheme value)
{
    var rootFrame = (Frame)Window.Current.Content;
    switch (value)
    {
        case AppTheme.Light:
            rootFrame.RequestedTheme = ElementTheme.Light;
            break;
        case AppTheme.Dark:
            rootFrame.RequestedTheme = ElementTheme.Dark;
            break;
        case AppTheme.System:
        default:
            rootFrame.RequestedTheme = ElementTheme.Default;
            break;
        }
    UpdateAppColors();
}
```

in this case You can do the same after you creating [Frame here](https://github.com/microsoft/calculator/blob/b34756d7943e2cd5c31a49e2255554a177403956/src/Calculator/App.xaml.cpp#L207) and when user change settings. New color should apply automatically

Chips1234:
@ad1Dima it looks like changing the RootFrame's theme does not work very well (the "close", "maximize", and "minimize" button does not change to white).

ad1Dima:
@Chips1234 yes, i forgot about this. If you still going this way you can set that colors [manually](https://docs.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.applicationviewtitlebar)

sanderl:
> > @sanderl. How about the converter? I see all converters have the same entry point, how do I store what mode the converter used to be on with AppData? Also, here's to code I have currently, the app crashed upon clicking the button:
> > ```c++
> > void SettingsPage::BackButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e, ViewMode /*model*/)
> > {
> >     auto previousModeValue = Windows::Storage::ApplicationData::Current->LocalSettings->Values->Lookup(L"previousMode");
> >     String ^ previousMode = safe_cast<String ^>(previousModeValue);
> > 
> >     if (previousMode == "Standard")
> >     {
> >         Model->Mode = ViewMode::Standard;
> >     }
> > }
> > ```
> > 
> > 
> > Thank you
> 
> @sanderl Could you answer the above question? Also, what do you think about app restart?

@Chips1234 for navigating back to the previous mode, I think you need to use the Deserialize method in NavCategory. 

// This function should only be used when restoring the mode from app data.
ViewMode NavCategory::Deserialize(Platform::Object ^ obj)

You'll want to make sure you never save Settings as the mode in AppData. 

I think your method probably crashed because there may not be a key called "previousMode". Without the exception details it's hard for me to know for sure.

I'm not sure what the restart question is about. If it is about changing the theme, then yes app restart is expected because RequestedTheme is not a dynamic api and can only be updated during app start.

Chips1234:
@sanderl. Hello, I've been playing around with it and thought maybe we could have a frame that contains SettingsPage and it would just be collapsed when the back button is clicked. I think that would be easier to implement. For example, check [this branch](https://github.com/Chips1234/calculator/tree/experimental)

Chips1234:
@grochocki? What do you think of collapsing a frame for the back button instead?

sanderl:
> @sanderl. Hello, I've been playing around with it and thought maybe we could have a frame that contains SettingsPage and it would just be collapsed when the back button is clicked. I think that would be easier to implement. For example, check [this branch](https://github.com/Chips1234/calculator/tree/experimental)

@Chips1234, it looks like you are continuing to use the flyout instead of navigating to a new page. Is that what you mean by having a frame contain the Settings page? If so, I don't think that is what is specified in the spec for the Settings page. I think we would like it to navigate to a page like the other pages and clicking the back button should go to the last used mode.

Chips1234:
@sanderl Thank you for your answer. No, I am not continuing to use the fly out. I just did not delete the fly out. If you run the program, you'll see what I mean. I think that collapsing a frame is much more easy to implement then storing the previous mode in app data.

Chips1234:
@sanderl, check the branch again, I've updated it.

Chips1234:
@sanderl See below gif. It is based on the [experimental branch](https://github.com/Chips1234/calculator/tree/experimental). If you think we can go forward with this way of doing it, I'll update this branch.
Calculator Settings Test
_When the settings button is clicked, the frame containing the settings page will appear. When the "back" button is clicked, the frame is collapsed. It's that simple._

sanderl:
> @sanderl See below gif. It is based on the [experimental branch](https://github.com/Chips1234/calculator/tree/experimental). If you think we can go forward with this way of doing it, I'll update this branch. Thank you.
> Calculator Settings Test
> _When the settings button is clicked, the frame containing the settings page will appear. When the "back" button is clicked, the frame is collapsed. It's that simple._

@Chips1234, I looked in the spec and it doesn't specify what should happen when pressing back. So I'm not 100% sure that the behavior you have is what we want or not.

@grochocki, what should happen when the user clicks the back button from the settings page?

Chips1234:
@sanderl He previously said "The back button should take the user back to the page they were on previously."

sanderl:
> @sanderl He previously said "The back button should take the user back to the page they were on previously."

@Chips1234, I'm glad you remembered. That's what I thought I remembered as well. I think the NavView being opened is the issue with the current implementation. The user should be able to use the previous mode right away without needing to dismiss the NavMenu first.

Chips1234:
> The user should be able to use the previous mode right away without needing to dismiss the NavMenu first.

Yes, I added the line `NavView->IsPaneOpen = false;`. Now, once the frame is collapsed, the NavView is also dismissed. If there are no more issues to address, I'll make changes to this branch.

sanderl:
> > The user should be able to use the previous mode right away without needing to dismiss the NavMenu first.
> 
> Yes, I added the line `NavView->IsPaneOpen = false;`. Now, once the frame is collapsed, the NavView is also dismissed. If there are no more issues to address, I'll make changes to this branch.

@Chips1234, that sounds good to me.

Chips1234:
@sanderl The back button is now implemented. Can you please review this? Thank you.

Chips1234:
@grochocki @sanderl review please? Thank you.

sanderl:
@Chips1234, Sorry for the slow reply, I was on vacation. I'll get to this later in the week after I have caught up.

Chips1234:
@sanderl No worries, take your time. Happy new year.

Chips1234:
@sanderl, I hope you're doing well, would you please review this pull request? Thank you.

Chips1234:
@grochocki Would you mind reviewing this, please? Thank you.

src/Calculator/App.xaml.cpp
```diff
@@ -70,6 +70,33 @@ App::App()
...
+        // Apply theme choice.
+        if (colorSetting == L"Dark")
+        {
+            App::Current->RequestedTheme = ApplicationTheme::Dark;
+        }
+        else if (colorSetting == L"Light")
+        {
+            App::Current->RequestedTheme = ApplicationTheme::Light;
+        }
+        else if (colorSetting == L"System")
```
sanderl, line 88: You don't need to explicitly detect and set the system theme. You just need to remove the saved theme from the app local settings. When the app restarts, then it will see there is no saved theme and the UWP defaults will kick in.

src/Calculator/Resources/en-US/Resources.resw
```diff
@@ -4699,4 +4663,79 @@
     <value>Rotate Through Carry Circular Shift selected</value>
     <comment>Label for a radio button that toggles rotate circular with carry behavior for the shift operations.</comment>
   </data>
-</root>
+  <data name="SettingsButton.Content" xml:space="preserve">
+    <value>Settings</value>
+    <comment>The text that shows in the dropdown navigation control to open the Settings page.</comment>
+  </data>
+  <data name="SettingsButton.AccessKey" xml:space="preserve">
```
sanderl, line 4670: I'm not sure we want to leave the access key as AB. @grochocki, can you confirm if this is the correct key combination to launch settings?

src/Calculator/SettingsPage.xaml
```diff
@@ -0,0 +1,164 @@
...
+            <Setter Property="FontWeight" Value="SemiBold"/>
+            <Setter Property="OpticalMarginAlignment" Value="TrimSideBearings"/>
+        </Style>
+    </Page.Resources>
+
+    <ScrollViewer>
+        <Grid x:Name="PageGrid" AutomationProperties.LandmarkType="Main">
+            <Grid.RowDefinitions>
+                <RowDefinition Height="1*"
+                           MinHeight="{StaticResource HamburgerHeight}"
```
sanderl, line 25: This seems strange to me. If I'm understanding correctly, then you are trying to explicitly give space to the NavView hamburger button. Are the other pages needing to do this?

src/Calculator/SettingsPage.xaml
```diff
@@ -0,0 +1,164 @@
+<Page x:Class="CalculatorApp.SettingsPage"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:Windows10version1809="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 7)"
```
sanderl, line 4: This should be renamed to "xmlns:contract7Present" to match the other instances of the Contract 7 API check.

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+    if (mainPage != nullptr)
+    {
+        MainPageProperty = mainPage;
+    }
+}
+
+void SettingsPage::InitializeContributeTextBlock()
+{
+    std::wstring contributeHyperlinkText  = resourceLoader->GetResourceString(L"SettingsContribute")->Data();
+
```
sanderl, line 79: Is there a reason you are not creating the link in Xaml using the hyperlink class? https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Documents.Hyperlink?view=winrt-19041

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+    ContributeRunBeforeLink->Text = contributeTextBeforeHyperlink;
+    ContributeRunLink->Text = contributeTextLink;
+    ContributeRunAfterLink->Text = contributeTextAfterHyperlink;
+
+}
+
+void SettingsPage::BackButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    MainPageProperty->CollapseSettings();
```
sanderl, line 105: You should be navigating back to the previous mode, not just collapsing settings.

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+void SettingsPage::SettingsFeedbackButtonClick(_In_ Object ^ sender, _In_ RoutedEventArgs ^ e)
+{
+    PackageVersion version = Package::Current->Id->Version;
+    String ^ versionNumber = ref new String(L"Version ");
+    versionNumber = versionNumber + version.Major + version.Minor + version.Build + version.Revision;
+
+    Launcher::LaunchUriAsync(ref new Uri("windows-feedback:?contextid=130&metadata=%7B%22Metadata%22:[%7B%22AppBuild%22:%22" + versionNumber + "%22%7D]%7D"));
+}
+
+void SettingsPage::LightChecked(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
```
sanderl, line 137: Can you collapse these into a ThemeChanged method that could handle of these cases in one method? This would reduce the duplicative code.

sanderl:
A couple of general comments, I don't think the settings page is going to work with Narrator since I don't see any automation names being set. 

It would also be good to check other accessibility modes like high contrast, large text, etc. 

The unit tests are also still failing, so those will need to be fixed before this PR can be merged.

Chips1234:
@sanderl Sure, here it is in high-contrast mode:
image

And here it is at 150% text size:
image

src/Calculator/App.xaml.cpp
```diff
@@ -70,6 +70,33 @@ App::App()
     this->HighContrastAdjustment = ApplicationHighContrastAdjustment::None;
 
     this->Suspending += ref new SuspendingEventHandler(this, &App::OnSuspending);
+    auto value = Windows::Storage::ApplicationData::Current->LocalSettings->Values->Lookup(L"themeSetting");
+    auto DefaultTheme = ref new Windows::UI::ViewManagement::UISettings();
+    auto uiTheme = DefaultTheme->GetColorValue(Windows::UI::ViewManagement::UIColorType::Background).ToString();
+    if (value != nullptr)
+    {
+        String ^ colorSetting = safe_cast<String ^>(value);
```
joseartrivera, line 78: Could we use an enum here instead of a string?

src/Calculator/SettingsPage.xaml
```diff
@@ -0,0 +1,164 @@
...
+            <Grid Height="{StaticResource HamburgerHeight}"
+              HorizontalAlignment="Left"
+              VerticalAlignment="Top">
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="{StaticResource HamburgerHeightGridLength}"/>
+                    <ColumnDefinition Width="Auto"/>
+                </Grid.ColumnDefinitions>
+
+                <Button x:Name="BackButton"
+                    x:Uid="BackButton"
```
joseartrivera, line 39: This needs an automation name, currently narrator just reads "button"

src/Calculator/Views/MainPage.xaml
```diff
@@ -151,6 +151,10 @@
             </Button>
         </Grid>
 
+        <Grid Grid.Row="1" Canvas.ZIndex="0">
+            <Frame x:Name="SettingsHolder" Visibility="Collapsed"/>
```
joseartrivera, line 155: This has an odd bug, the page behind the settings page is still loaded so when I use keyboard navigation I can still access UI elements behind the settings page. For example I can pop up the 'scientific' menu from scientific calculator with the keyboard. Can we handle this page like the other calculator modes?

image

src/Calculator/SettingsPage.xaml
```diff
@@ -0,0 +1,164 @@
...
+
+                    <RadioButton x:Name="SettingsSystemTheme"
+                             x:Uid="SettingsSystemTheme"
+                             Grid.Row="4"
+                             Checked="SystemChecked"/>
+
+
+                    <TextBlock x:Name="SettingsRestartApp"
+                           Style="{StaticResource CaptionTextBlockStyle}"
+                           Foreground="{StaticResource SystemColorGrayTextColor}"/>
```
joseartrivera, line 86: In dark mode the color of this text is hard to read
image

src/Calculator/SettingsPage.xaml
```diff
@@ -0,0 +1,164 @@
...
+                             x:Uid="SettingsLightTheme"
+                             Grid.Row="3"
+                             Checked="LightChecked"/>
+
+                    <RadioButton x:Name="SettingsDarkTheme"
+                             x:Uid="SettingsDarkTheme"
+                             Grid.Row="2"
+                             Checked="DarkChecked"/>
+
+                    <RadioButton x:Name="SettingsSystemTheme"
```
joseartrivera, line 78: If I set this radio button, hit the back button, and then go back to settings, my selection is reset. I think you may be missing code to load from settings on page navigation.

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Xaml::Controls::Primitives;
+using namespace Windows::UI::Xaml::Data;
+using namespace winrt::Windows::Storage;
+
+#ifndef BUILD_YEAR
+#define BUILD_YEAR 2021
+#endif
+
+auto resourceLoader = AppResourceProvider::GetInstance();
```
joseartrivera, line 28: member variables are typically declared in the header file and prefixed with m_ in this project. Also these should be initialized in the constructor instead.

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+auto themevalue = Windows::Storage::ApplicationData::Current -> LocalSettings -> Values -> Lookup(L"themeSetting");
+String ^ colorSetting = safe_cast<String ^>(themevalue);
+
+SettingsPage::SettingsPage()
+{
+    auto locService = LocalizationService::GetInstance();
+
+	InitializeComponent();
```
joseartrivera, line 36: nit: Odd spacing

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    Language = locService->GetLanguage();
```
joseartrivera, line 38: What exactly is this Language property doing? Is it part of the page?

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,177 @@
...
+    {
+        SettingsSystemTheme->IsChecked = true;
+    }
+
+    InitializeContributeTextBlock();
+}
+
+void SettingsPage::OnNavigatedTo(Windows::UI::Xaml::Navigation::NavigationEventArgs ^ e)
+{
+    MainPage ^ mainPage = (MainPage ^) e->Parameter;
```
joseartrivera, line 69: Use static_cast or dynamic_cast instead

joseartrivera (changes requested):
Good job with this PR! The settings page is looking great. I found a few bugs that should be addressed before we merge this in, let me know what you think. Thanks for tackling this!

Chips1234:
@joseartrivera Hello, and thank you for your comments. I'll review them soon.

src/Calculator/Views/MainPage.xaml
Chips1234: @joseartrivera I tried to handle it that way, but I can't figure out how to navigate back to the previous page. This is easier since pressing the "Back" button just collapses the frame.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: Because of localization. See #1449 for more.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: @sanderl If we do that, we'll have to store the previous mode in AppData, then navigate back to it, which is complicated. This just collapses the grid that contains the settings page.

src/Calculator/App.xaml.cpp
Chips1234: @sanderl How can I do that? Thank you.

src/Calculator/Resources/en-US/Resources.resw
Chips1234: @grochocki Should it be "SE" instead?

src/Calculator/Views/MainPage.xaml
Chips1234: @joseartrivera If we don't use the collapse frame method for the back button, what should we do?

src/Calculator/App.xaml.cpp
Chips1234: @joseartrivera Why do we need to do that? Thanks for your suggestions, by the way.

src/Calculator/SettingsPage.xaml
Chips1234: Hi @sanderl, I'm doing that because this is not the NavView button, this is the back button.

Chips1234:
@joseartriveria @sanderl Can you two please review this again? I understand that both of you want to navigate to the settings page just like the other pages, but if we do that, it'd be hard to navigate back to the previous page, this just simply collapses the settings page. Thanks, have a great weekend.

Chips1234:
@joseartriveria Hello?

Chips1234:
@joseartrivera  Hello! I was wondering if you can be so kind as to review this PR? Thanks.

src/Calculator/Views/MainPage.xaml
joseartrivera: @Chips1234 we should utilize [page ](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.page?view=winrt-19041) navigation or [frame](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.Frame?view=winrt-19041 )navigation instead

Chips1234: @joseartrivera Hello, since we are now handling the settings page like a normal calculator page, do you have a idea how we can store the previous page on file, then navigate back to it? (My previous solution, to just collapse the frame containing the page was much easier to program, in my opinion.) Thanks, have a great Easter.

joseartrivera: Would you be able to use the GoBack() method on the frame? https://docs.microsoft.com/en-us/windows/uwp/design/basics/navigate-between-two-pages

Chips1234: @joseartrivera I don't think we can do that, since the other pages are not contained in a frame.

Chips1234: @joseartrivera I'll try it though, thank you for your suggestion.

joseartrivera: I see, how do we handle switching between different modes today? Maybe sticking to that pattern would work.

Chips1234: @joseartrivera Hello, I tried using `rootFrame->Navigate(SettingsPage::typeid);` and to go back, `rootFrame->GoBack();`. It works fine, but when I close the application during debugging, it throws an exception saying that "Source not available." If you would like to debug, here's the branch I used to experiment: [https://github.com/Chips1234/calculator/tree/experimental](https://github.com/Chips1234/calculator/tree/experimental). Do you know how to fix that? Thanks.

Chips1234: OK, @joseartrivera. The SettingsPage is now navigated via rootFrame, and when the back button is clicked, it navigates back to MainPage. This method should resolve the <key>tab</key> problem. Could you please review this PR again? Thanks, have a great rest of your weekend!

joseartrivera: Awesome! Thanks for the updates, I will re-review this soon.

Chips1234: > 
> 
> Awesome! Thanks for the updates, I will re-review this soon.

Thanks.

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    if (SettingsRestartApp->Visibility == ::Visibility::Visible)
+    {
+        Windows::Storage::ApplicationData::Current->LocalSettings->Values->Insert(L"restartApp", L"True");
+    }
+    else
+    {
+        Windows::Storage::ApplicationData::Current->LocalSettings->Values->Insert(L"restartApp", L"False");
+    }
+    rootFrame->Navigate((MainPage::typeid), this);
```
joseartrivera, line 143: There is kind of a weird glitch where the hamburger menu flys open for a second and then closes after navigating.

src/Calculator/SettingsPage.xaml
```diff
@@ -0,0 +1,164 @@
...
+                                 Checked="ThemeChecked"
+                                 Tag="Dark"/>
+
+                    <RadioButton x:Name="SettingsSystemTheme"
+                                 x:Uid="SettingsSystemTheme"
+                                 Checked="ThemeChecked"
+                                 Tag="System"/>
+
+
+                    <TextBlock x:Name="SettingsRestartApp"
```
joseartrivera, line 84: In light mode this is hard to read, might be reproed by switching to high contrast and back?
image

src/Calculator/SettingsPage.xaml.cpp
joseartrivera: Actually, this may only repro in debugging mode

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,208 @@
...
+using namespace winrt::Windows::Storage;
+
+#ifndef BUILD_YEAR
+#define BUILD_YEAR 2021
+#endif
+
+SettingsPage::SettingsPage()
+{
+    auto resourceLoader = AppResourceProvider::GetInstance();
+    auto themevalue = Windows::Storage::ApplicationData::Current->LocalSettings->Values->Lookup(L"themeSetting");
```
joseartrivera, line 31: What happens if these settings don't exist?

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,208 @@
...
+    InitializeComponent();
+
+    this->SetVersionString();
+    this->SetCopyrightString();
+    this->InitializeContributeTextBlock();
+    SetDefaultFocus(); 
+    
+    if (themevalue != nullptr)
+    {
+        if (colorSetting == L"Light")
```
joseartrivera, line 47: Consider making these strings a const instead

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,208 @@
...
+    ContributeRunLink->Text = contributeTextLink;
+    ContributeRunAfterLink->Text = contributeTextAfterHyperlink;
+
+}
+
+void SettingsPage::BackButtonClick(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    auto rootFrame = dynamic_cast<::Frame ^>(Window::Current->Content);
+
+    if (SettingsRestartApp->Visibility == ::Visibility::Visible)
```
joseartrivera, line 135: Might be better to do this at the time the user switches their selection instead to avoid writing this to settings every time they use the back button.

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    // Only display the restart app notice if the theme selected is not the current theme.
+    if (currentTheme == tag)
+    {
+        SettingsRestartApp->Visibility = ::Visibility::Collapsed;
+    }
+    else
+    {
+        SettingsRestartApp->Visibility = ::Visibility::Visible;
+        SettingsRestartApp->Text = resourceLoader->GetResourceString("SettingsRestartNotice");
```
joseartrivera, line 193: Can you set the Text in XAML and then just control visibility in code behind instead so that it only gets set once?

src/Calculator/SettingsPage.xaml.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    if (tag == "Light")
```
joseartrivera, line 196: Would it be possible to make these an enum instead of a string?

joseartrivera (changes requested):
The changes are looking really good! Thanks for the fixes :). I left just a few more comments that should be minor changes / verifications. @sanderl can you do a review of this too to make sure we closed out any open issues you may have had?

src/Calculator/App.xaml.cpp
joseartrivera: I left this comment again, didn't realize I had left it previously. An enum is safter to use than a string, it avoids people from entering a typo into the string if they make changes to this in the future.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: @joseartrivera Yeah, I'm only getting this when I'm debugging.

Chips1234:
Thanks, I'll take a look.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: @joseartrivera Hello, just to be clear, do you want me to `std::map` the strings to enum? Thanks.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: @joseartrivera Sorry, but what string are you referring to? colorSetting?

src/Calculator/SettingsPage.xaml.cpp
joseartrivera: Would it be possible to just use an enum and store the enum in storage directly?

src/Calculator/SettingsPage.xaml.cpp
joseartrivera: Yes, but if you can make them an enum that would be better. (I posted this before thinking of the enum approach)

src/Calculator/SettingsPage.xaml
Chips1234: @joseartrivera Hmm, I'm not quite sure how to fix this, actually.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: @joseartrivera I think I've resolved the issue you're referring to, but I'm not quite sure. Could you please check? Thanks.

Chips1234: @joseartrivera?

src/Calculator/SettingsPage.xaml
joseartrivera: Try making the foreground a ThemeResource instead of a StaticResource

src/Calculator/SettingsPage.xaml.cpp
joseartrivera: Yes, that works also!

src/Calculator/App.xaml.cpp
joseartrivera: If the colorSetting == L"System" you can just not set the App::Current->RequestedTheme and the app will automatically default to the desired system theme.

src/Calculator/SettingsPage.xaml
Chips1234: @joseartrivera Sorry for the late reply, I was busy lately. I'll try that.

Chips1234: OK, this should be resolved now.

src/Calculator/SettingsPage.xaml.cpp
Chips1234: I think I solved the issue you were talking about; if not, please open another conversation.

sanderl:
Based on a few prototypes our team did, we have recently made the decision to shift over to C# as a language for parts of the core code for calculator [(#893).](https://github.com/microsoft/calculator/issues/893)  However, sadly, this PR will be directly affected by this design change. Due to that, we will have to close this PR out as this will not be compatible.  The PR for the shift will be coming in the next few days.

We thank you so much for all of your work.  We know this is a disappointing bit of news that the PR will have to be closed out.

Chips1234:
OK, @sanderl. Thanks for letting me know. I'm much more experienced in C#, therefore, I look forward to contributing to this project in the future.

hanzhang54:
@Chips1234 Thank you so much for the great work on the settings page! We have created a feature branch feature/winui2.6 to update styles and controls to WinUI 2.6 in the C# based project and the new settings page has been added to the feature branch. We are stabilizing the feature branch, and looking forward to your contribution in the future! 🎉

Chips1234:
@hanzhang54 All right, thank you.

PR #1337, opened 2020-08-06
Updated the Calculator projects to include the /Z7 compiler flag

sanderl:
### Description of the changes:
- Added the /Z7 compiler flag to the AdditionalOptions compiler configuration

PR #1356, opened 2020-08-21
Revert "Updated the Calculator projects to include the /Z7 compiler flag"

sanderl:
Reverts microsoft/calculator#1337


sanderl:
@Rahul-Dixit, we determined that the default /Zi flag was sufficient and we did not actually need to override it with the /Z7 flag. Since there was no significant gain from the /Z7 flag, I thought it was best to revert the change.

PR #1365, opened 2020-09-04
Edited KeyboardShortcutManager.cpp Issue #744

### LINKED ISSUE #744: Clicking on the History Label causes the [Shift] + [Ctrl] + [D] hotkeys to break
**Describe the bug**
Clicking on the History Label causes the [Shift] + [Ctrl] + [D] hotkeys to break

**Steps To Reproduce**
1. Launch Windows Calculator
2. Ensure the window is sized large enough so that the History panel is viable
3. Perform a calculation
    - E.g. 4*5/6=
    - A history item should populate the history panel
3. Click on the "history" pivot button.
4. Try to use [Shift] + [Ctrl] + [D] to clear history.
5. See error

**Observed behavior**
Using the [Shift] + [Ctrl] + [D] hotkeys has no impact; nothing appears to happen when the keys are pressed

**Expected behavior**
Using the [Shift] + [Ctrl] + [D] hotkeys clears the history panel

**Screenshots**
image

 - OS Build: 10.0.19505.0
 - Architecture: X64
 - Application Version: 10.1909.1.0
 - Region: en-US
 - Dev Version Installed: True

**Additional context**
- This caused my automated test to fail, but it took manual testing to figure out why this hotkey function broke
- Once this function breaks, it stays broken until the application closed and relaunched
- The history label is naturally clicked when navigating between Memory and History, so this issue could impact users who switch between Memory and History and then want to use the hotkeys to clear the history panel  

**Requested Assignment**

razan45:
- Moved the check for Shift and Control out of the "!args->Handled" as it was fails for just ctrl and just shift leading to the issue
-
-

- Manually tested by following steps on issue
- Ran included tests to make sure no other functionality broke
-

### DIFF
--- a/src/Calculator/Common/KeyboardShortcutManager.cpp
+++ b/src/Calculator/Common/KeyboardShortcutManager.cpp
@@ -495,12 +495,38 @@ const std::multimap<MyVirtualKey, WeakReference>& GetCurrentKeyDictionary(MyVirt
 
 void KeyboardShortcutManager::OnKeyDownHandler(CoreWindow ^ sender, KeyEventArgs ^ args)
 {
+    auto key = args->VirtualKey;
+    int viewId = Utils::GetWindowId();
+    //Handle Control and Shift first 
+    if (key == VirtualKey::Control)
+    {
+        // Writer lock for the static maps
+        reader_writer_lock::scoped_lock lock(s_keyboardShortcutMapLock);
+
+        auto currControlKeyPressed = s_ControlKeyPressed.find(viewId);
+
+        if (currControlKeyPressed != s_ControlKeyPressed.end())
+        {
+            s_ControlKeyPressed[viewId] = true;
+        }
+        return;
+    }
+    else if (key == VirtualKey::Shift)
+    {
+        // Writer lock for the static maps
+        reader_writer_lock::scoped_lock lock(s_keyboardShortcutMapLock);
+
+        auto currShiftKeyPressed = s_ShiftKeyPressed.find(viewId);
+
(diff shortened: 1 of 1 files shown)


greedyAI:
Since there aren't any other event handlers for `Pivot` which handles the OnKeyDown event when `KeyEventArgs`'s  `handled ` is true, this fix appears sufficient to fix the indicated issue without causing side effects.

However, the suggested fix may not be fully extensible in the sense that future keyboard shortcuts that don't involve the Ctrl or Shift keys could also run into a similar issue, and it's unclear whether this fix would work if a four-key keyboard shortcut was ever created (eg. Ctrl-Shift-Alt-A). But this issue may be moot depending on the trajectory of calculator's product management.

Deferring to @grochocki for PM sign-off.

grochocki:
It is true that there might be other exceptional cases in the future, but I would say we can tackle those if they come up. Probably good to get this in to help in this (relatively) common case.

razan45:
@grochocki Let me know if you need me to change anything! :)

mcooley:
It looks like #1223 already fixed this issue, so this change is no longer necessary. That PR stopped attempting to track up/down events for the ctrl and shift keys, and just [checks with CoreWindow directly](https://github.com/microsoft/calculator/blob/15ae66626ad8d23c525c17ff11a2146eb060102f/src/Calculator/Common/KeyboardShortcutManager.cpp#L437). That fixes lots of issues, including this one.

I'm sorry it took so long to review this. Even though it didn't turn out to be necessary, I'm thankful for your efforts here.

PR #1373, opened 2020-09-15
Fix Race condition in tests involving context menu

EriWong:
- Added extension method to wait for an element to be created in the UI tree.
- Waits for the context menu items to be visible before trying to interact with them.

- ran tests failing race condition locally

### DIFF
--- a/src/CalculatorUITestFramework/WindowsDriverExtensions.cs
+++ b/src/CalculatorUITestFramework/WindowsDriverExtensions.cs
@@ -102,8 +102,6 @@ namespace CalculatorUITestFramework
                         throw;
                     }
                 }
-
-                Logger.LogMessage("Waiting for 10ms in WaitForElementByAccessibilityId");
                 Thread.Sleep(10);
             }
             timer.Stop();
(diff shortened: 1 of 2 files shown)


src/CalculatorUITestFramework/WindowsDriverExtensions.cs
```diff
@@ -68,5 +71,45 @@ public static void SwitchToCurrentWindowHandle(this WindowsDriver<WindowsElement
...
+                    {
+                        Logger.LogMessage("Element not found. Waiting for 10ms in WaitForElementByAccessibilityId");
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
+
+                Logger.LogMessage("Waiting for 10ms in WaitForElementByAccessibilityId");
```
EriWong, line 106: extraneous logging.  Will remove.

sanderl (approved):
Thanks for adding this waiter for the context menu.

mcooley:
@EriWong @sanderl I think we should revert this PR, since the test run reliability has not improved since it was merged. I don't think this is how "waiters" should be implemented with WinAppDriver--WinAppDriver has an implicit wait for most UI elements and we should probably start by adjusting those settings.

PR #1375, opened 2020-09-16
fix UI tests in lab runs

EriWong:
We are seeing failures in our official runs when trying to access the context menu.  Changing some tests to avoid using the context menu and instead using keyboard commands

- local automation runs pass.


mcooley:
Are we sure this is the right approach?

I think we should revert #1373, since it seems like that didn't actually improve the reliability of this test (and I don't think the "waiter" implementation in that PR is the right approach, since we should be using WinAppDriver's implicit wait).

mcooley (changes requested):
Still want to discuss the approach here.

EriWong:
@mcooley Locally, these tests pass (even before #1373), but they don't work in the pipeline runs. These tests are more focused around copy paste management, so this change is mostly using keyboard to bypass the context menu (and reverts #1373, as well as removes the code that doesn't pass in pipeline runs).  We can do this separately by first reverting that previous attempt at fixing it. I'm not sure what the benefit of doing this as 2 PRs is (though I'm happy to do it that way if there's a good reason to).

The failure that prompted this and #1373 was automation failing to find the context menu items.

mcooley (approved):

PR #1394, opened 2020-09-25
Update2010 release

volkcor:
-
-
-

-
-
-


mdtauk:
Is Calculator sticking with 2010 as a version convention, even though 20H2 seems to be the official designation publically and internally?

image

grochocki:
Yes, in part because we ship retail versions of Calculator more frequently than there are Windows OS releases, but also because our releases are tied directly to the app build number, which only support numerical values.

For example, our 2008 release is 10.2008.*:
image

PR #1401, opened 2020-10-02
Edit fields are scalable for large fonts

### LINKED ISSUE #1381: Numbers present at "Step" edit field do not get properly visible at text scaling in 220%.
**Pre-Requisites:**
Set the slider value to maximum i.e. 220% (Settings -> Ease of access -> text scaling).

**Repro Steps:** 
1. Launch Calculator app.
2. Navigate Open navigation button and activate it.
3. Navigate to graphing calculator list item and activate it.
4. Navigate to Enter a expression edit box.
5. Writ e the expression as "A+B".
6. Now navigate to variable option button of any variable say "a" and activate it.
7. Now the variable options will get visible.
8. Navigate to step edit field.
9. Observe the issue.

**Actual Result:**
Numbers present at "Step" edit field do not get properly visible at text scaling in 220%.
For e.g. number in step edit field is "0.1" so "1" is not getting properly visible.

**Expected Result:**
Numbers present at "Step" edit field should get properly visible at text scaling in 220%.

image

JackRainy:
- The edit boxes (min value, step value, max value) max width is not limited now, so they can be scaled for any large fonts.

- The changes are tested manually with different font sizes from 100 to 225%
Screenshot 2020-10-02 10-26-13


grochocki:
Thanks, @JackRainy! A quick note -- I think the trick here is to support all scale sizes, while still maintaining this visual:

image
(notice the visual footprint of the underline for each textblock)

JackRainy:
I have updated the PR in order to retain the initial edit fields size (and scale if needed) according to the @grochocki suggestion.
Screenshot 2020-10-05 21-55-44

greedyAI:
This fix should be appropriate in fixing the indicated issue without causing side effects.

Deferring to @grochocki for PM sign-off.

grochocki:
Thanks for that change! Visuals look good

quentin987:
Great changes, looks pretty close! Looks like scaling is working, but the values get clipped for longer numeric values. I did some testing with some values like 5.6789 The clipping is a little more obvious with the bigger font sizes. The textbox should probably expand to use all available horizontal space before clipping.

Here's an example at 225% font. 
image

Here's clipping at 100% font.
image

quentin987 (changes requested):
Left a comment here about some tweaks to make for clipping. https://github.com/microsoft/calculator/pull/1401#issuecomment-713200024

JackRainy:
@quentin987 , thank you for taking a look. Could you, please, tell how did you reproduce the issue? My screenshots are also made with 100% and 225% font, and there is no such problem. Your screenshot looks like it was *before* the fix. Are you sure, you got it with the latest changes from the PR included?

quentin987:
> @quentin987 , thank you for taking a look. Could you, please, tell how did you reproduce the issue? My screenshots are also made with 100% and 225% font, and there is no such problem. Your screenshot looks like it was _before_ the fix. Are you sure, you got it with the latest changes from the PR included?

Thanks for mentioning this, my machine wasn't picking up the changes. Looks good! Approved!

quentin987 (approved):

PR #1411, opened 2020-10-22
fix: change floppy disk size

leuri397:
## Fixes #.
Fixed floppy disk size to real. It is 1474560 bytes for standard 3,5' floppy (https://en.wikipedia.org/wiki/Floppy_disk)

-
-
-

-
-
-


mcooley:
Thanks! A standard (HD) 3.5" floppy is indeed 1440 _Kibibytes_.

mcooley (approved):

PR #1412, opened 2020-10-24
Fixes issue #1409

### LINKED ISSUE #1409: Copy pasting into Currency locks the editor
**Describe the bug**

Copy pasting from any source into the currency locks the editor and you cannot clear the editor.

**Steps To Reproduce**

Open Notepad
Type text "20.83" and copy the text 
Open the calculator app and paste the text with Ctrl+V
See that you removing or CE button doesn't clear anything.

**Expected behavior**

I would expect the editor still continue to edit. Nothing happens after copy paste.

**Screenshots**

Currency

 - OS Build: 10.0.19041.0
 - Architecture: X64
 - Application Version: 10.2009.4.0
 - Region: en-US
 - Dev Version Installed: false

**Additional context**

N/A

**Requested Assignment**

BuraChuhadar:
#Fixes #1409.

- Skipping the validation of full editor lock when backspace or clear button is clicked.

- Manual testing.
- Added UI tests
- Added Unit tests

### DIFF
--- a/src/CalcViewModel/UnitConverterViewModel.cpp
+++ b/src/CalcViewModel/UnitConverterViewModel.cpp
@@ -474,7 +474,9 @@ void UnitConverterViewModel::OnButtonPressed(Platform::Object ^ parameter)
 
     static constexpr UCM::Command OPERANDS[] = { UCM::Command::Zero, UCM::Command::One, UCM::Command::Two,   UCM::Command::Three, UCM::Command::Four,
                                                    UCM::Command::Five, UCM::Command::Six, UCM::Command::Seven, UCM::Command::Eight, UCM::Command::Nine };
-    if (m_isInputBlocked && (command != UCM::Command::Clear && command != UCM::Command::Backspace))
+    if (m_isInputBlocked &&
+        command != UCM::Command::Clear &&
+        command != UCM::Command::Backspace)
 	{
 		return;
 	}
--- a/src/CalcViewModel/UnitConverterViewModel.cpp
+++ b/src/CalcViewModel/UnitConverterViewModel.cpp
@@ -474,7 +474,7 @@ void UnitConverterViewModel::OnButtonPressed(Platform::Object ^ parameter)
 
     static constexpr UCM::Command OPERANDS[] = { UCM::Command::Zero, UCM::Command::One, UCM::Command::Two,   UCM::Command::Three, UCM::Command::Four,
                                                    UCM::Command::Five, UCM::Command::Six, UCM::Command::Seven, UCM::Command::Eight, UCM::Command::Nine };
-	if (m_isInputBlocked)
+    if (m_isInputBlocked && (command != UCM::Command::Clear && command != UCM::Command::Backspace))
 	{
 		return;
 	}
--- a/src/CalculatorUITestFramework/UnitConverterPage.cs
+++ b/src/CalculatorUITestFramework/UnitConverterPage.cs
@@ -1,0 +1,68 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using OpenQA.Selenium.Appium.Windows;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace CalculatorUITestFramework
+{
+    public class UnitConverterPage
+    {
+        public UnitConverterOperatorsPanel UnitConverterOperators = new UnitConverterOperatorsPanel();
+        public NavigationMenu NavigationMenu = new NavigationMenu();
+        public UnitConverterResults UnitConverterResults = new UnitConverterResults();
+
+        private WindowsDriver<WindowsElement> session => WinAppDriver.Instance.CalculatorSession;
+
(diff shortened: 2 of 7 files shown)


src/CalculatorUITestFramework/UnitConverterPage.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void EnsureCalculatorResultTextIsZero()
+        {
+            if ("0" != this.UnitConverterResults.GetCalculationResult1Text())
+            {
+                this.ClearAll();
+            }
+        }
+
+        /// <summary>
+        /// Navigates the caclulator to Standard mode and ensures that it is in standard mode
```
ItzLevvie, line 38:
```suggestion
        /// Navigates the calculator to Standard mode and ensures that it is in standard mode
```

ItzLevvie (changes requested):

src/CalculatorUITestFramework/UnitConverterPage.cs
BuraChuhadar: This is fixed. Thank you.

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -474,7 +474,7 @@ void UnitConverterViewModel::OnButtonPressed(Platform::Object ^ parameter)
 
     static constexpr UCM::Command OPERANDS[] = { UCM::Command::Zero, UCM::Command::One, UCM::Command::Two,   UCM::Command::Three, UCM::Command::Four,
                                                    UCM::Command::Five, UCM::Command::Six, UCM::Command::Seven, UCM::Command::Eight, UCM::Command::Nine };
-	if (m_isInputBlocked)
+    if (m_isInputBlocked && (command != UCM::Command::Clear && command != UCM::Command::Backspace))
```
EriWong, line 477: Minor nit:
You've got an extra set of parens here.  I understand that you are adding them for clarity, but I think in this case, it's straightforward enough that we don't need them.  Also, for readability, I like having each condition on a separate line if there are more than 2 conditions.

src/CalculatorUITests/UnitConverterFunctionalTests.cs
```diff
@@ -0,0 +1,121 @@
...
+// Licensed under the MIT License.
+using CalculatorUITestFramework;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OpenQA.Selenium;
+using System;
+
+namespace CalculatorUITests
+{
+    [TestClass]
+    public class UnitConverterFunctionalTests
```
EriWong, line 11: Nit: For organization purposes, can we call these CurrencyConverterFunctionalTests?  If we end up adding tests for each mode, it will help keep things organized.

src/CalculatorUITests/UnitConverterFunctionalTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        #region Basic UI Functionality via Mouse Input Tests
+
+        /// <summary>
+        /// These automated tests verify clicking each of the buttons in the Calculator UI and getting an expected result
+        /// Via mouse input, all basic UI functionality is checked 
+        /// </summary>
+        [TestMethod]
+        [Priority(0)]
+        public void MouseInput_EnterInputWithFullDecimalAndClear()
```
EriWong, line 61: I love that you've added UI tests to cover both basic functionality and your fix.  Also while we normally prefer unit tests over UI tests to cover functionality (as they tend to be both faster and more stable), this particular fix is tricky to unit test.

That said, because UI tests are harder to maintain, I'd like to try to keep these to as simple as possible.  In this case, I think it means having 3 tests:
1. a priority 0 test that covers the most basic of input.  ie: hits button 2 and then verifies both calculation results
2. a priority 1 test that covers backspace.  ie: inputs 2.13, verifies the initial input, and then backspaces all 3, checking the expected output each time.
3. a priority 1 test that covers clear.  ie: inputs 1.23, verifies the initial input, and then clicks clear, checking for the input to be cleared afterwards.

Thoughts?

EriWong (changes requested):

src/CalcViewModel/UnitConverterViewModel.cpp
BuraChuhadar: This is done.

src/CalculatorUITests/UnitConverterFunctionalTests.cs
BuraChuhadar: This is done.

src/CalculatorUITests/UnitConverterFunctionalTests.cs
BuraChuhadar: Sure I changed the UI tests no problem. I also added some unit test cases in addition to UI tests. The testing value however for the testing should be two digit value not one digit. So I kept "20.43" value in every test case.

EriWong (approved):
Thanks for fixing this!  Also, thanks for adding UI and Unit tests!

PR #1416, opened 2020-10-28
Changed output of GetCurrentRadix to uint32_t match the variable type of m_radix

PokeCodec:
## Fixes #.
The m_radix variable in the CCalcEngine class is a uint32_t variable. However, it is returned as an int, despite the only cases of this variable being used expect the variable to be a uint32_t. Because of this, I changed the return type of GetCurrentRadix() to be uint32_t.

- Changed GetCurrentRadix() to be uint32_t
- Changed variable being assigned the returned value of GetCurrentRadix to auto
- Changed the function prototype to match the above as well

- Manual testing
- Passed all tests


mcooley (approved):
Thanks for this cleanup!

PR #1427, opened 2020-11-08
Set Full Link Time Optimization

PokeCodec:
For build machines on release builds, there is no reason to do incremental link time optimization, especially since this program is relatively small.

- Set Full Time Optimization to improve Release builds


PokeCodec:
Interesting to note, the UITests passed in less time than usual.

mcooley:
We already set `<WholeProgramOptimization>true</WholeProgramOptimization>`, which turns on /GL. And [the documentation says](https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation) that "/LTCG is implied with /GL". So I don't think that this change to add /LTCG actually changes any settings.

PokeCodec:
Not true: for x64, the setting is incremental Link time, which is different from the link time optimization setting that every other architecture for this project has.

mcooley:
Continuing discussion in #1427

PR #1428, opened 2020-11-08
Switch to unsigned indexes for Date Calculations

PokeCodec:
- Made checks unsigned, since most of the 32 bits in an int are irrelevant to Date Processing

- Unit Tests, as well as ensuring all values are unsigned in all cases.


PokeCodec:
The values changed to unsigned will NEVER go below 0

mcooley:
I don't think this change will cause any problems, but I also don't see any benefits. One could also argue that `int` is better because it's shorter to type, or `size_t` is better because it's more semantically correct. Without a stronger rationale for `unsigned int`, I think we should just keep the code the way it is today.

PR #1430, opened 2020-11-11
Lock m_renderMain in Grapher::TryInitializeGraph to avoid race condition

### LINKED ISSUE #1293: Graphing - remove equation crashes app
**Describe the bug**
Calculator crashes when removing an equation in graphing mode

**Steps To Reproduce**
1. Calculator graphing mode
2. Graph three expressions, for example: x^2 , x^3 , x^4
3. Edit expression x^2 by clicking on it
4. Click the 'x' to the right of the expression to clear it
5. On expression x^3 choose 'Remove equation'
6. Calculator crashes

**Expected behavior**
The x^3 expression should be removed leaving an empty expression and x^4

 - OS Build: 10.0.19041.0
 - Architecture: X64
 - Application Version: 10.2005.23.0
 - Region: en-GB

**Requested Assignment**

calc-crash

xuhongxu96:
## Fixes #1293 .

The Calculator crashes when removing an expression with another expression just cleared.  
This behavior will trigger the graph updating **twice in a short time**, which leads to a potential race condition to `m_graph`.
The app crashes because one thread is rendering the `m_graph` and another tries to initialize the `m_graph`.

- Add lock to `m_renderMain->GetCriticalSection()` in `Grapher::TryInitializeGraph`
- Minor improvements to code style

- After this fix, the bug cannot repro per steps in #1293

> As `Graph` is not included in the community version, tests cannot be added.

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -118,7 +118,7 @@ namespace GraphControl
     {
         if (m_graph != nullptr && m_renderMain != nullptr)
         {
-            if(auto renderer = m_graph->GetRenderer())
+            if (auto renderer = m_graph->GetRenderer())
             {
                 HRESULT hr;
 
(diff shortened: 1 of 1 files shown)


EriWong (approved):
Code looks good, and buddy testing does show the issue is resolved.  Thanks for fixing this!

Issue #1204, opened 2020-04-30
Cannot input decimals in currency converter


hez2010:
**Describe the bug**
Cannot input decimals in currency converter

**Steps To Reproduce**
1. Go to Currency converter
2. Input decimals
3. Cannot input any number after floating point

**Expected behavior**
Should be able to input numbers after floating point

**Device and Application Information**
 - OS Build: 10.0.19041.0
 - Architecture: X64
 - Application Version: 10.2003.0.0
 - Region: ja-JP
 - Dev Version Installed: False

**Requested Assignment**
I'm just reporting this problem.  I don't want to fix it.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

DhirajPAmrutkar:
I want to fix the issue.

rudyhuyn:
Hi @hez2010, thank you for reporting this issue!

This bug is a duplicate of #424. Here is more information about the reason:

From @Geveffe 
> The reason that this happens is that when you change your Format to Japanese (Japan) the Currency Format changes by default to have 0 "No. of digits after decimal". This may or may not be correct, but Calculator is correctly using the format. I will follow up with the intl.cpl team form them to decide if the current intl.cpl behavior is correct or not.

I will close this issue but let's continue this conversation in the other one!

PR #1432, opened 2020-11-12
Use different formatter for different currency

### LINKED ISSUE #424: Application fails to take float value for Currency converter when Japanese Calendar is enabled

xuhongxu96:
## Fixes #424 , #1204 , #1438  .

Before, the Calculator will use `CurrencyFormatter` per system locale for all currencies.
So, users cannot input `USD` with fraction digits in Japanese locale, because `JPY` doesn't allow fraction digits.

- Allow to use different formatter for different currency
   - By adding `m_currencyFormatter1`, `m_currencyFormatter2`

- <del>Functional UI tests</del> (it does not work because there is no full currency list in CI)
- Add unit test case `TestCurrencyFormattingLogic` for `UnitConverterViewModel`

### DIFF
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -24,7 +24,7 @@ static constexpr uint32_t OPTIMALDIGITSALLOWED = 7U;
 static constexpr wchar_t LEFTESCAPECHAR = L'{';
 static constexpr wchar_t RIGHTESCAPECHAR = L'}';
 
-static const double OPTIMALDECIMALALLOWED = 1e-6; // pow(10, -1 * (OPTIMALDIGITSALLOWED - 1));
+static const double OPTIMALDECIMALALLOWED = 1e-6;  // pow(10, -1 * (OPTIMALDIGITSALLOWED - 1));
 static const double MINIMUMDECIMALALLOWED = 1e-14; // pow(10, -1 * (MAXIMUMDIGITSALLOWED - 1));
 
 unordered_map<wchar_t, wstring> quoteConversions;
@@ -147,6 +147,11 @@ void UnitConverter::SetCurrentUnitTypes(const Unit& fromType, const Unit& toType
     if (!CheckLoad())
     {
         return;
+    }
+
+    if (m_fromType != fromType)
+    {
+        m_switchedActive = true;
     }
 
     m_fromType = fromType;
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -196,13 +196,6 @@ bool UnitConversionManager::UnitConverter::IsSwitchedActive() const
     return m_switchedActive;
 }
 
-void UnitConversionManager::UnitConverter::SetValue(const std::wstring& newValue)
-{
-    m_returnDisplay = m_currentDisplay;
-    m_currentDisplay = newValue;
-    m_currentHasDecimal = (m_currentDisplay.find(L'.') != wstring::npos);
-}
-
 wstring UnitConverter::CategoryToString(const Category& c, wstring_view delimiter)
 {
     return Quote(std::to_wstring(c.id))
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -191,6 +191,18 @@ void UnitConverter::SwitchActive(const wstring& newValue)
     }
 }
 
+bool UnitConversionManager::UnitConverter::IsSwitchedActive() const
+{
+    return m_switchedActive;
+}
+
+void UnitConversionManager::UnitConverter::SetValue(const std::wstring& newValue)
+{
+    m_returnDisplay = m_currentDisplay;
+    m_currentDisplay = newValue;
+    m_currentHasDecimal = (m_currentDisplay.find(L'.') != wstring::npos);
+}
+
 wstring UnitConverter::CategoryToString(const Category& c, wstring_view delimiter)
 {
     return Quote(std::to_wstring(c.id))
--- a/src/CalcViewModel/UnitConverterViewModel.cpp
+++ b/src/CalcViewModel/UnitConverterViewModel.cpp
@@ -134,7 +134,6 @@ UnitConverterViewModel::UnitConverterViewModel(const shared_ptr<UCM::IUnitConver
     m_currencyFormatter->IsGrouped = true;
     m_currencyFormatter->Mode = CurrencyFormatterMode::UseCurrencyCode;
     m_currencyFormatter->ApplyRoundingForCurrency(RoundingAlgorithm::RoundHalfDown);
-    m_currencyMaxFractionDigits = m_currencyFormatter->FractionDigits;
 
     auto resourceLoader = AppResourceProvider::GetInstance();
     m_localizedValueFromFormat = resourceLoader->GetResourceString(UnitConverterResourceKeys::ValueFromFormat);
@@ -228,6 +227,8 @@ void UnitConverterViewModel::OnUnitChanged(Object ^ parameter)
         return;
     }
 
+    UpdateCurrencyFormatter();
+
     m_model->SetCurrentUnitTypes(UnitFrom->GetModelUnit(), UnitTo->GetModelUnit());
     if (m_supplementaryResultsTimer != nullptr)
     {
@@ -246,7 +247,7 @@ void UnitConverterViewModel::OnSwitchActive(Platform::Object ^ unused)
     if (m_relocalizeStringOnSwitch)
     {
         // clean up any ill-formed strings that were in progress before the switch
-        ValueFrom = ConvertToLocalizedString(m_valueFromUnlocalized, false);
+        ValueFrom = ConvertToLocalizedString(m_valueFromUnlocalized, false, CurrencyFormatterParameterFrom);
     }
 
     SwitchConversionParameters();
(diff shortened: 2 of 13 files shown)


xuhongxu96:
Seems I cannot add currency functional test because it only has "Jupiter - Gas Money" in ComboBox in CI

xuhongxu96:
Seems I cannot add currency functional test because it only has a fake currency in CI.
Any suggestions?

xuhongxu96:
Seems in CI the Calculator doesn't have full currency list?

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -312,18 +338,18 @@ String ^ UnitConverterViewModel::ConvertToLocalizedString(const std::wstring& st
...
             m_decimalFormatter->FractionDigits = static_cast<int>(stringToLocalize.length() - (posOfDecimal + 1));
-            m_currencyFormatter->FractionDigits = m_currencyMaxFractionDigits;
+            currencyFormatter->FractionDigits = lastCurrencyFractionDigits;
         }
 
         if (IsCurrencyCurrentCategory)
         {
-            wstring currencyResult = m_currencyFormatter->Format(stod(stringToLocalize))->Data();
-            wstring currencyCode = m_currencyFormatter->Currency->Data();
+            wstring currencyResult = currencyFormatter->Format(stod(stringToLocalize))->Data();
```
joseartrivera, line 351: This results in the conversions to Yen from other currencies always displaying a decimal, is this existing behavior?

image

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -280,10 +282,33 @@ String ^ UnitConverterViewModel::ConvertToLocalizedString(const std::wstring& st
         return result;
     }
 
+    CurrencyFormatter ^ currencyFormatter;
+
+    switch (cfp)
+    {
+    case CurrencyFormatterParameter::Default:
```
joseartrivera, line 289: Might be better to just make this the 'default:' case

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -803,10 +833,26 @@ void UnitConverterViewModel::UpdateInputBlocked(_In_ const wstring& currencyInpu
     m_isInputBlocked = false;
     if (posOfDecimal != wstring::npos && IsCurrencyCurrentCategory)
     {
-        m_isInputBlocked = (posOfDecimal + static_cast<size_t>(m_currencyMaxFractionDigits) + 1 == currencyInput.length());
+        m_isInputBlocked = (posOfDecimal + static_cast<size_t>(CurrencyFormatterFrom->FractionDigits) + 1 == currencyInput.length());
     }
 }
 
+void UnitConverterViewModel::UpdateCurrencyFormatter()
```
joseartrivera, line 840: Should we disable the '.' keyboard key in the case of yen too?

joseartrivera (changes requested):
Thanks for the fix! This seems to be working great, just a few questions to answer, let me know what you think.

src/CalcViewModel/UnitConverterViewModel.cpp
xuhongxu96: It's an existing behavior. If you change locale to Japanese, all currency converted to will have a decimal:
image

But I think we can discuss about this -- is it really needed?

src/CalcViewModel/UnitConverterViewModel.cpp
xuhongxu96: Sure.

src/CalcViewModel/UnitConverterViewModel.cpp
xuhongxu96: I think this question is related to the first one -- Shall we keep the trailing decimal?
If we decided to show the trailing decimal, it would be weird to disable the '.' key.

src/CalcViewModel/UnitConverterViewModel.cpp
joseartrivera: I don't think it's needed, would be great if it is easy to remove!

src/CalcViewModel/UnitConverterViewModel.cpp
joseartrivera: I am fine with keeping the existing behavior

src/CalcViewModel/UnitConverterViewModel.cpp
xuhongxu96: Should be easy. Let me do it together with disabling decimal input.

src/CalcManager/UnitConverter.cpp
```diff
@@ -191,6 +191,11 @@ void UnitConverter::SwitchActive(const wstring& newValue)
     }
 }
 
+bool UnitConversionManager::UnitConverter::IsSwitchedActive() const
+{
+    return m_switchedActive;
+}
+
```
xuhongxu96, line 198: Before, after user switches active and then changes the currency unit, the input may be blocked.
This is because the unit changed to has less fraction digits than the previous unit.

However, as the new value will always replace the old value after switched active, user should be able to input regardless of the old value. 

So, I have to add this method so that `ViewModel` can decide whether to block the input per `switchedActive`.

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -473,14 +504,12 @@ void UnitConverterViewModel::OnButtonPressed(Platform::Object ^ parameter)
...
-	{
-		return;
-	}
-	m_model->SendCommand(command);
+                                                 UCM::Command::Five, UCM::Command::Six, UCM::Command::Seven, UCM::Command::Eight, UCM::Command::Nine };
+    if (m_isInputBlocked && !m_model->IsSwitchedActive() && command != UCM::Command::Clear && command != UCM::Command::Backspace)
+    {
+        return;
+    }
+    m_model->SendCommand(command);
```
xuhongxu96, line 512: `!m_model->IsSwitchedActive()` means input is allowed as long as user just switches active.

src/CalcManager/UnitConverter.cpp
xuhongxu96: I think we'd better add some real currency data (only abbr. matters, the rates can be fake) in UI Tests.
Otherwise, we cannot easily test the case mentioned here.

src/CalcViewModel/UnitConverterViewModel.cpp
xuhongxu96: I've added the same comment in code.

src/CalcManager/UnitConverter.cpp
xuhongxu96: @joseartrivera As the test data is discussed in another issue thread, maybe we can do it in another PR. Let's skip it for now?

src/CalculatorUITestFramework/UnitConverterPage.cs
```diff
@@ -64,5 +64,44 @@ public void EnsureCalculatorIsCurrencyMode()
...
+            CalculatorApp.ClickOnWindow();
+            UnitConverterOperators.Units1.SendKeys(OpenQA.Selenium.Keys.Home);
+            UnitConverterOperators.Units1.SendKeys(OpenQA.Selenium.Keys.Enter);
+
+            CalculatorApp.ClickOnWindow();
+            UnitConverterOperators.Units2.SendKeys(OpenQA.Selenium.Keys.Home);
+            UnitConverterOperators.Units2.SendKeys(OpenQA.Selenium.Keys.Enter);
+
+            CalculatorApp.ClickOnWindow();
+        }
```
xuhongxu96, line 81: Should fix #1438

src/CalcViewModel/UnitConverterViewModel.cpp
```diff
@@ -293,7 +319,8 @@ String ^ UnitConverterViewModel::ConvertToLocalizedString(const std::wstring& st
         std::wstring significandStr(stringToLocalize.substr(0, posOfE));
         std::wstring exponentStr(stringToLocalize.substr(posOfSign + 1, stringToLocalize.length() - posOfSign));
 
-        result += ConvertToLocalizedString(significandStr, allowPartialStrings) + "e" + signOfE + ConvertToLocalizedString(exponentStr, allowPartialStrings);
+        result += ConvertToLocalizedString(significandStr, allowPartialStrings, cfp) + "e" + signOfE
```
joseartrivera, line 322: There is a bug when testing with a currency without fractional digits, I get two different answers in this circumstance:

1. Switch first currency to "Test Fractional Digits"
2. Switch second currency to "Test Fractional Digits"
3. In the first currency input, type 1.999
4. Switch the first currency to "Test No Fractional Digits" 
5. Notice the conversion says 2 -> 399.800, even though the conversion rate is 1 = 200.00
6. Hit the CE buton, type 2 in manually, notice the answer is now 2 -> 400.

I think the issue is that we display how the currency is formatted and displayed but we don't use that value in the actual calculation.

joseartrivera (changes requested):

src/CalcViewModel/UnitConverterViewModel.cpp
xuhongxu96: Wow, thanks for your finding! 
Seems that I have to truncate the digits after switching the currency, which means the digits will be gone forever even if I switch back.

xuhongxu96: Fixed and added a UI test case.

src/CalculatorUITests/CurrencyConverterFunctionalTests.cs
```diff
@@ -256,6 +257,45 @@ public void MouseInput_SwitchCurrencyWithLessFractionalDigitsAndEnterInput()
...
+            page.UnitConverterOperators.ClearButton.Click();
+
+            page.UnitConverterOperators.NumberPad.Num2Button.Click();
+            page.UnitConverterOperators.NumberPad.Num0Button.Click();
+            page.UnitConverterOperators.NumberPad.Num0Button.Click();
+
+            Assert.AreEqual("200", page.UnitConverterResults.GetCalculationResult1Text());
+            Assert.AreEqual(result, page.UnitConverterResults.GetCalculationResult2Text());
+        }
+
```
xuhongxu96, line 298: Added a UI test case for the bug you found

joseartrivera (approved):
Works great, thanks again for the fix!

PR #1444, opened 2020-11-29
Fix copy-pasting result computed by the calculator gives "Invalid input"

### LINKED ISSUE #1407: Copy-pasting result computed by the calculator gives "Invalid input"
Copy-pasting result computed by the calculator (5.562684646268003457725581793331e-309) gives "Invalid input":
image

See also: https://github.com/microsoft/calculator/issues/1301.

chaitanyamehta:
## Fixes #1407 
This pull request fixes copy paste issue with exponential numbers. 

When a user tries to enter a big exponential number (eg. 1.1111111111111111111111111111111e+1142). The length of entire number is compared with maximum operand length. Only the length of mantissa should be compared with max operand length.

- While calculating operand length for standard and scientific modes, if the operand contains exponent then subtract exponent string length from operand length.

- Application was manually tested with inputs having exponential component
- Additional positive and negative unit test inputs have been added to CopyPasteManagerTest.cpp

### DIFF
--- a/src/CalculatorUnitTests/CopyPasteManagerTest.cpp
+++ b/src/CalculatorUnitTests/CopyPasteManagerTest.cpp
@@ -825,9 +825,13 @@ namespace CalculatorUnitTests
                                      L"12^2",
                                      L"-12.12^-2",
                                      L"61%99"
-                                     L"-6.1%99" };
+                                     L"-6.1%99",
+                                     L"1.1111111111111111111111111111111e+1142" };
         String
             ^ negativeInput[] = { L"abcdef", L"xyz",   L"ABab", L"e+234", L"123456789123456781234567890123456" /*boundary condition: greater than 32 digits*/,
+                                  L"11.1111111111111111111111111111111e+1142",
+                                  L"1.1e+10001", /*boundary condition: exponent greater than 5 digits*/
+                                  L"0.11111111111111111111111111111111111e+111111" /*boundary condition: both exponent and non exponent exceed limits*/
                                   L"SIN(2)", L"2+2==", L"2=+2" };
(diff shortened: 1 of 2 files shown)


src/CalculatorUnitTests/CopyPasteManagerTest.cpp
```diff
@@ -825,9 +825,12 @@ namespace CalculatorUnitTests
                                      L"12^2",
                                      L"-12.12^-2",
                                      L"61%99"
-                                     L"-6.1%99" };
+                                     L"-6.1%99",
+                                     L"1.1111111111111111111111111111111e+1142" };
         String
             ^ negativeInput[] = { L"abcdef", L"xyz",   L"ABab", L"e+234", L"123456789123456781234567890123456" /*boundary condition: greater than 32 digits*/,
+                                  L"11.1111111111111111111111111111111e+1142",
+                                  L"1.1e+10001", /*boundary condition:exponent greater than 5 digits*/
```
greedyAI, line 833: Can you add another unit test for the 0.xxxe+yyy case (eg. 0.11111111111111111111111111111111111+111111)? This case is handled separately in CopyPasteManager.cpp, as shown in the diff.

greedyAI (changes requested):

src/CalculatorUnitTests/CopyPasteManagerTest.cpp
chaitanyamehta: Will add this boundary case. Here both exponent and non exponent components exceed limits.

greedyAI (approved):
Looks good. Thanks for your help!

PR #1448, opened 2020-12-04
Rename LICENSE to LICENSE.txt

jeremygautama:
## Just a minor update #.
Changed LICENSE to LICENSE.txt


mcooley:
Thanks for the interest in this repo, but we'd prefer to keep the filename as-is.

PR #1452, opened 2020-12-17
Fix compiler warnings

PokeCodec:
Fixed compiler warnings that appear when compiling this project

-
-
-


joseartrivera:
Hi @PokeCodec, for some reason I am not seeing these compiler warnings in the logs when building calculator without these changes. Are you using another tool to detect them?

joseartrivera (changes requested):
Hi @PokeCodec, for some reason I am not seeing these compiler warnings in the logs when building calculator without these changes. Are you using another tool to detect them?

PokeCodec:
JetBrains Resharper C++ gave warnings about casting

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -1032,7 +1031,7 @@ bool stripzeroesnum(_Inout_ PNUMBER pnum, int32_t starting)
     // point pmant to the LSD
     if (cdigits > starting)
     {
-        pmant += cdigits - starting;
+        pmant = pmant + cdigits - starting;
```
joseartrivera, line 1034: Why not leave this as +=?

src/CalcManager/Ratpack/num.cpp
```diff
@@ -418,12 +418,11 @@ void _divnum(PNUMBER* pa, PNUMBER b, uint32_t radix, int32_t precision)
...
     destroynum(tmp);
 
-    int32_t digit;
     int32_t cdigits = 0;
     while (cdigits++ < thismax && !zernum(rem))
     {
-        digit = radix - 1;
-        PNUMBER multiple = nullptr;
+        uint32_t digit = radix - 1;
+        PNUMBER multiple = rem;
```
joseartrivera, line 425: Why set this to rem in this case?

joseartrivera (changes requested):
Hi @PokeCodec, sorry for my delay in response here. Most of the changes look good, just a few questions regarding the reasoning behind a few. Thanks again!

src/CalcManager/Ratpack/num.cpp
SleepyDweams: This is because rem is the starting place before we change the multiple anyway

PR #1455, opened 2020-12-28
Fixed an issue where keyboard focus navigates twice at Domain value

### LINKED ISSUE #1389: Keyboard focus navigates twice at Domain value "⁅𝑥∈ℝ⁆".

yaichenbaum:
- This change sets the `UseSystemFocusVisuals` property to false in the KeyGraphFeaturesControls template.

Unfortunately, I was not able to get the KeyGraphFeaturesControl to show up in debug mode, but after looking at the XAML, I am fairly sure this change will fix the issue.

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
+++ b/src/Calculator/Views/GraphingCalculator/KeyGraphFeaturesPanel.xaml
@@ -177,7 +177,9 @@
                                         <ColumnDefinition Width="Auto" MinWidth="64"/>
                                         <ColumnDefinition Width="Auto"/>
                                     </Grid.ColumnDefinitions>
-                                    <controls:MathRichEditBox Style="{StaticResource KGF_RichEditBoxStyle}" MathText="{x:Bind Expression}"/>
+                                    <controls:MathRichEditBox Style="{StaticResource KGF_RichEditBoxStyle}"
+                                                              IsTabStop="False"
+                                                              MathText="{x:Bind Expression}"/>
                                     <TextBlock Grid.Column="1"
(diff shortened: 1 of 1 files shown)


joseartrivera:
This change doesn't seem to fix the issue. Also I believe MathRichEdit box does need to receive focus for narrator to correctly read out the contents from the textbox. You should be able to get to the KeyGraphFeaturesControl by typing anything into the equation textbox while in debug mode, the mock should treat it as a valid equation iirc.

joseartrivera (changes requested):
This change doesn't seem to fix the issue. Also I believe MathRichEdit box does need to receive focus for narrator to correctly read out the contents from the textbox. You should be able to get to the KeyGraphFeaturesControl by typing anything into the equation textbox while in debug mode, the mock should treat it as a valid equation iirc.

yaichenbaum:
> This change doesn't seem to fix the issue. Also I believe MathRichEdit box does need to receive focus for narrator to correctly read out the contents from the textbox. You should be able to get to the KeyGraphFeaturesControl by typing anything into the equation textbox while in debug mode, the mock should treat it as a valid equation iirc.

For some reason, the graph control doesn't load in debug, perhaps because of licensing? I am pushing another commit that might fix this issue.

joseartrivera (changes requested):
UseSystemFocusVisuals would only hide the blue rectangle, but the control would still get focus, except it would be hidden. The GraphingControl isn't available in debug, but you should still be able to add equations and access the UI with the MockGraphControl for testing.

yaichenbaum:
@joanassantos Closing this as I haven't been able to solve it correctly.

Issue #1386, opened 2020-09-23
No confirmation is announced by the narrator after activating 'Remove equation' button


MicrosoftIssueBot:
**Pre-Requisites:**
Expressions should be added.

**Repro Steps:** 
1. Launch Calculator app.
2. Turn on the narrator.
3. Navigate Open navigation button and activate it.
4. Navigate to graphing calculator list item and activate it.
5. Navigate to Enter a expression edit box.
6. Enter any expression say "7+8".
7. Now open context menu and activate "Remove equation" button.
8. Observe the issue.

**Actual Result:**
No confirmation is announced by the narrator after activating 'Remove equation' button.

**Expected Result:**
Confirmation should be announced by the narrator after activating 'Remove equation' button.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

PR #1465, opened 2021-01-08
fix bug: No confirmation is announced by the narrator after activating 'Remove equation' button #1386

MSFT-Tilia:
## Fixes #1386 .
Changing AutomationNotificationProcessing from `MostRecent` to `ImportantMostRecent` to prevent this confirmation messages being superseded by the incoming announcement requests.

Target issue: #1386

### Description of the changes:
- Uses `ImportantMostRecent` instead of `MostRecent` for function removing announcement.

### How changes were validated:

1. Launch Calculator app.
1. Turn on the narrator.
1. Navigate Open navigation button and activate it.
1. Navigate to graphing calculator list item and activate it.
1. Navigate to Enter a expression edit box.
1. Enter any expression say "7+8".
1. Now open context menu and activate "Remove equation" button.
1. Check that the confirmation - 'Function Removed' - is announced by Narrator.

PR #1469, opened 2021-01-14
Adds unit-test cases for NarratorAnnouncement after fixing issue #1386

MSFT-Tilia:
## Fixes #.
Adds unit test cases for NarratorAnnouncement
PR #1465 which fixed issue #1386 has changed a part of NarratorAnnouncement's behaviors. To strengthen the quality of the code, add unit test cases for NarratorAnnouncement.

- add Test-Class NarratorAnnouncementUnitTests with 18 test methods.

- Run CalculatorUnitTests project
- All newly added test methods succeed.

### DIFF
--- a/build/pipelines/templates/build-app-internal.yaml
+++ b/build/pipelines/templates/build-app-internal.yaml
@@ -27,7 +27,7 @@ jobs:
       downloadDirectory: $(Build.SourcesDirectory)
       vstsFeed: WindowsInboxApps
       vstsFeedPackage: calculator-internals
-      vstsPackageVersion: 0.0.53
+      vstsPackageVersion: 0.0.54
 
   - template: ./build-single-architecture.yaml
     parameters:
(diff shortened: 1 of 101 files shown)


joseartrivera (changes requested):
Thanks for the PR! I think testing Narrator functionality is a great idea. Since the NarratorAnnouncement classes doesn't actually contain any narrator functionality these tests would only be testing that the right values are set and returned. What do you think about holding off on these tests for now until we can see if there is a good way to automate narrator tests?

MSFT-Tilia:
Thanks for reviewing this PR, Pepe :)

Yes, these unit tests are going to ensure right values are returned by class `CalculatorAnnouncement`. They are very naive and straight forward but the purpose I made them is to put a barrier that asks the person who is changing the announcements to make sure he/she knows what he/she is doing.

I'm fine to holding off this PR since so far I didn't come up with a way to manage narrator tests, which I think should be integration/functional tests instead of unit tests.

Thinking from unit test perspective, I'd like to have the tests if it doesn't cost too much resources.

joseartrivera (approved):
@MSFT-Tilia thanks for the clarification! I think it makes sense to have these tests if we think it will help from causing the bug again.

PR #1426, opened 2020-11-05
switch to use policy manager apis for detecting graphing policy

sanderl:
### Description of the changes:
- In NavCategory.cpp, update the use the Policy Manager APIs instead of the registry to look up the AllowGraphingCalculator policy.
-Since the AllowGraphingCalculator policy is a user policy, we get the users and look up the policy using the Education area and AllowGraphingCalculator with the GetPolicyFromPathForUser Policy Manager API: https://docs.microsoft.com/en-us/uwp/api/windows.management.policies.namedpolicy.getpolicyfrompathforuser?view=winrt-19041
### How changes were validated:

- Manually using Group Policy to validate the new method still picks up the policy value.

PR #1471, opened 2021-01-20
Fixed a graphing calculator "permissions" bug caused by PR #1426

greedyAI:
- The PR #1426 can cause a crash when no users are returned via `User::FindAllAsync(UserType::LocalUser)` when subsequently trying to access the first user. The existing code also does not guarantee that the returned user is the currently active user.
- This fix retrieves the user that opened the app and passes this user into a function to check if this user has the proper permissions to access the graphing mode. This makes sense since the active user is indistinguishable (at least from the app's perspective) to the user who opened the app. This user's permissions are then propagated downwards to properly set up the navigation menu of the app.
- Implementation detail worth pointing out: `s_categoryManifest` is what is used to populate the navigation menu of the app, but this variable is static by design, so a separate function was written to override the appropriate `isEnabled` value in `s_categoryManifest`. This function is called by `onLaunched`.

- Manual testing

### DIFF
--- a/src/CalcViewModel/Common/NavCategory.cpp
+++ b/src/CalcViewModel/Common/NavCategory.cpp
@@ -62,34 +62,29 @@ bool IsGraphingModeAvailable()
 Box<bool> ^ _isGraphingModeEnabledCached = nullptr;
 bool IsGraphingModeEnabled(User ^ firstUser = nullptr)
 {
-    if (firstUser)
+    if (!IsGraphingModeAvailable())
     {
-        if (!IsGraphingModeAvailable())
-        {
-            return false;
-        }
-
-        if (_isGraphingModeEnabledCached != nullptr)
-        {
-            return _isGraphingModeEnabledCached->Value;
-        }
-
-        auto namedPolicyData = NamedPolicy::GetPolicyFromPathForUser(firstUser, L"Education", L"AllowGraphingCalculator");
-        _isGraphingModeEnabledCached = namedPolicyData->GetBoolean() == true;
-
-        return _isGraphingModeEnabledCached->Value;
+        return false;
     }
-    else
+
+    if (_isGraphingModeEnabledCached != nullptr)
     {
         return _isGraphingModeEnabledCached->Value;
     }
+
+    if (!firstUser)
+    {
+        return false;
+    }
+
+    auto namedPolicyData = NamedPolicy::GetPolicyFromPathForUser(firstUser, L"Education", L"AllowGraphingCalculator");
+    _isGraphingModeEnabledCached = namedPolicyData->GetBoolean() == true;
+
+    return _isGraphingModeEnabledCached->Value;
 }
 
 // The order of items in this list determines the order of items in the menu.
-static list<NavCategoryInitializer> s_categoryManifest;
-
-void NavCategory::CreateCategoryManifest(User ^ user)
-{
+static list<NavCategoryInitializer> s_categoryManifest = [] {
     auto res = list<NavCategoryInitializer>{ NavCategoryInitializer{ ViewMode::Standard,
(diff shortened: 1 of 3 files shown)


src/Calculator/App.xaml.cpp
```diff
@@ -198,12 +198,12 @@ void App::OnLaunched(LaunchActivatedEventArgs ^ args)
         // If the app got pre-launch activated, then save that state in a flag
         m_preLaunched = true;
     }
+    NavCategory::InitializeCategoryManifest(args->User);
```
joseartrivera, line 201: Any scenarios where this would be null? And if so, does it make sense to fallback to the User::FindAllAsync logic?

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -72,17 +72,19 @@ bool IsGraphingModeEnabled()
...
     }
 
-    User ^ firstUser;
-    create_task(User::FindAllAsync(UserType::LocalUser)).then([&firstUser](IVectorView<User ^> ^ users) {
-        firstUser = users->GetAt(0); }).wait();
-        auto namedPolicyData = NamedPolicy::GetPolicyFromPathForUser(firstUser, L"Education", L"AllowGraphingCalculator");
-        _isGraphingModeEnabledCached = namedPolicyData->GetBoolean() == true;
+    if (!firstUser)
+    {
+        return false;
```
joseartrivera, line 77: Since Graphing is available by default and is only disabled when specifically configured, does it make sense to set this to true if User is null?

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -276,6 +278,13 @@ static const list<NavCategoryInitializer> s_categoryManifest = [] {
     return res;
 }();
 
+void NavCategory::InitializeCategoryManifest(User ^ user)
+{
+    auto navCatInit = s_categoryManifest.begin();
+    std::advance(navCatInit, 2);
```
joseartrivera, line 284: Is it possible to not hardcode the index and detect which entry is for Graphing Mode? I can see this easily breaking in the future if someone updates the s_categoryManifest but doesn't update this code.

joseartrivera (changes requested):
Thanks for the fix! Left a few comments to consider.

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -60,7 +60,7 @@ bool IsGraphingModeAvailable()
 }
 
 Box<bool> ^ _isGraphingModeEnabledCached = nullptr;
-bool IsGraphingModeEnabled()
+bool IsGraphingModeEnabled(User ^ firstUser = nullptr)
```
mcooley, line 63: nit: `currentUser` would be a better name for this variable

src/Calculator/App.xaml.cpp
greedyAI: Since `User::FindAllAsync` does not guarantee that the returned user is the currently active user, fundamentally its result is not super meaningful. So, if `args->User` is null, it'd make more sense to go along with the comment below, which would the permission to true.

src/CalcViewModel/Common/NavCategory.cpp
greedyAI: Resolved

src/CalcViewModel/Common/NavCategory.cpp
greedyAI: Resolved

src/CalcViewModel/Common/NavCategory.cpp
greedyAI: Resolved

joseartrivera (approved):

Issue #1380, opened 2020-09-23
Narrator is not reading the already entered equation when focus lands on Enter a equation edit box.


MicrosoftIssueBot:
**Repro Steps:** 
1. Launch 'Calculator' and turn on narrator.
2. Home screen will be displayed.​
3. Navigate Open navigation button and activate it.
4. Navigate to graphing calculator list item and activate it.
5. Navigate to Enter a equation edit box.
6. Observe the issue. 

**Actual Result:**
Narrator is not reading the already entered equation when focus lands on Enter a equation edit box.

**Expected Result:**
Narrator should read the already entered equation when focus lands on Enter a equation edit box.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

grochocki:
After further investigation, this issue appears to be a platform bug. Closing this and routing to one of our internal partners.

PR #1482, opened 2021-02-01
fixes up bug: Narrator is not reading the already entered equation when focus lands on Enter an equation edit box.

tian-lt:
Target Issue: #1380 Narrator is not reading the already entered equation when focus lands on Enter a equation edit box.

## Fixes #
Once there's focus landed on equation text box, engage Narrator to announce the equation by jumping the insertion position back and forth ending with a left arrow key action.

- Added an event handler listening to [GotFocus event](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.gotfocus?view=winrt-19041)
- When equation edit box got focus:
  1. Set insertion position to the end of line.
  2. Set insertion position to the start of line.
  3. execute a left-arrow command.

1. Launched 'Calculator' and turn on narrator.
1. Home screen displayed.​
1. Navigated Open navigation button and activate it.
1. Navigated to graphing calculator list item and activate it.
1. Navigated to Enter a equation edit box.
1. Checked that the equation content has been announced.

### DIFF
--- a/src/Calculator/Controls/MathRichEditBox.cpp
+++ b/src/Calculator/Controls/MathRichEditBox.cpp
@@ -115,7 +115,7 @@ void MathRichEditBox::SetMathTextProperty(String ^ newValue)
 
 void MathRichEditBox::OnGotFocus(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
 {
-    // bug: https://dev.azure.com/microsoft/OS/_workitems/edit/28498627
+    // [BUG 28498627][GithubIssue #1380]
     // below directives on Selection are going to engage Narrator to announce the content of this richedit
     // this is a workaround since richedit doesn't announce its content automatically.
     this->Document->Selection->EndKey(TextRangeUnit::Line, false);
--- a/src/Calculator/Controls/MathRichEditBox.cpp
+++ b/src/Calculator/Controls/MathRichEditBox.cpp
@@ -133,7 +133,6 @@ void MathRichEditBox::OnLosingFocus(UIElement ^ sender, LosingFocusEventArgs ^ a
     SubmitEquation(EquationSubmissionSource::FOCUS_LOST);
 }
 
-
 void MathRichEditBox::OnKeyUp(Object ^ sender, KeyRoutedEventArgs ^ e)
 {
     if (!this->IsReadOnly && e->Key == VirtualKey::Enter)
--- a/src/Calculator/Controls/MathRichEditBox.cpp
+++ b/src/Calculator/Controls/MathRichEditBox.cpp
@@ -12,6 +12,7 @@ using namespace std;
 using namespace Windows::ApplicationModel;
 using namespace Windows::UI::Core;
 using namespace Windows::UI::Xaml;
+using namespace Windows::UI::Xaml::Automation::Peers;
 using namespace Windows::UI::Xaml::Input;
 using namespace Windows::UI::Xaml::Controls;
 using namespace Windows::UI::Text;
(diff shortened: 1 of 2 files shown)


src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -122,6 +133,7 @@ void MathRichEditBox::OnLosingFocus(UIElement ^ sender, LosingFocusEventArgs ^ a
     SubmitEquation(EquationSubmissionSource::FOCUS_LOST);
 }
 
+
```
nandin-borjigin, line 136: redundant line

tian-lt: Great finding!
Resolved.

src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -72,6 +72,7 @@ MathRichEditBox::MathRichEditBox()
     {
         throw Exception::CreateException(hr);
     }
+    this->GotFocus += ref new RoutedEventHandler(this, &MathRichEditBox::OnGotFocus);
```
nandin-borjigin, line 75: Should we remove the delegate combination somewhere else?

mcooley (changes requested):
This approach (moving the cursor around) is a hack that might cause other problems. Instead, we should consider whether we can override the automation peer to make the RichEditBox behave more like a normal text box, or suggest a change to Narrator's behavior.

src/Calculator/Controls/MathRichEditBox.cpp
```diff
@@ -112,6 +113,16 @@ void MathRichEditBox::SetMathTextProperty(String ^ newValue)
...
 }
 
+void MathRichEditBox::OnGotFocus(Platform::Object ^ sender, Windows::UI::Xaml::RoutedEventArgs ^ e)
+{
+    // [BUG 28498627][GithubIssue #1380]
+    // below directives on Selection are going to engage Narrator to announce the content of this richedit
+    // this is a workaround since richedit doesn't announce its content automatically.
+    this->Document->Selection->EndKey(TextRangeUnit::Line, false);
+    this->Document->Selection->HomeKey(TextRangeUnit::Line, false);
+    this->Document->Selection->MoveLeft(TextRangeUnit::Character, 1, false);
```
joseartrivera, line 123: This seems like it would be annoying for the end user if they click at a specific character they want to edit but the cursor is suddenly moved to a different location.

joseartrivera (changes requested):

src/Calculator/Controls/MathRichEditBox.cpp
joseartrivera, line 123: This seems like it would be annoying for the end user if they click at a specific character they want to edit but the cursor is suddenly moved to a different location.

tian-lt: Agree with this annoying point and the suggest that @mcooley mentioned above.
Instead of hacking upon this hack-fix, I'll try to find a better way to resolve this bug from either RichEditBox side or Narrator side.

tian-lt:
> This approach (moving the cursor around) is a hack that might cause other problems. Instead, we should consider whether we can override the automation peer to make the RichEditBox behave more like a normal text box, or suggest a change to Narrator's behavior.

Hi @mcooley ,
Thanks for your comment!
It's true that this fix is a hack.
About AutomationPeer, I've given it a try. However, regarding below [clarification](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.richeditboxautomationpeer?view=winrt-19041#textpattern-support) about RichEditBoxAutomationPeer, it doesn't seem to be able to override the peer to make RichEditBox behave more like a normal text box.
>TextPattern support
RichEditBoxAutomationPeer supports the TextPattern Microsoft UI Automation pattern. The support for this pattern is implemented as native code by the peer, so you won't see public API for the RichEditBoxAutomationPeer peer class that exposes the pattern implementation for extension. And you won't see ITextRangeProvider and the similar managed interfaces in the RichEditBoxAutomationPeer inheritance. But Microsoft UI Automation clients can use the TextPattern pattern implemented by this peer and call its native API surface.

But it indeed makes sense that we suggest changes for Narrator or RichEditBox.
I'll dig into RichEdit first to see if there is a chance to have a simple correct fix.

src/Calculator/Controls/MathRichEditBox.cpp
tian-lt: Sorry, what do you mean by '*delegate combination*'?

mcooley:
We shouldn't use different behavior when "Narrator is turned on". There could be a UI automation listener attached for many reasons besides Narrator--the same concerns about the hack still apply.

tian-lt:
yes, commit 029e781 is a workaround above another workaround.
I pushed it because it's a cheap but tricky fix, meanwhile I was investigating at RichEdit control.
Now, it turns out that seems RichEdit returns good `textpattern` back to the requester which should be Narrator in this case, and somehow, Narrator doesn't read out the text.

tian-lt:
This bug has been confirmed that is a Narrator bug.
Redirected to their team and they have logged some tasks to fix it in future.

PR #1518, opened 2021-03-25
A few fixes/improvements to the Portuguese translation.

ric2b:
## Fixes

A few incorrect translations to pt-PT that I've noticed.

They're just string changes and they don't significantly change the string lengths so it's very unlikely that they'll cause text rendering issues.


alli03106 (approved):

worldtiki (approved):

grochocki:
Unfortunately, we are not able to make localization changes directly into the codebase. I have filed a bug with an internal team to help address this issue. Thanks for flagging these! :)

PR #1523, opened 2021-04-11
Disabled focus for 2nd button in Scientific Mode

### LINKED ISSUE #1503: Pressing enter does not calculate after I click on "2nd"
**Describe the bug**
When I click on the button to use secondary functions in scientific mode and then press enter it activates the button I clicked rather than calculate the result

**Steps To Reproduce**
1. Enter the Scientific mode
2. Click **2nd** button
3. You may click secondary functions that change when you clicked the previous button (I did not test all of them, but **yroot** seems to work)
4. Enter some calculations from the keyboard (don't click any other buttons) **2 + 2**, for example
5. Press Enter
6. Get functions change as if you pressed **2nd** button again

**Expected behavior**
Seeing the result of the calculation (**4**)

 - OS Build: 10.0.19042.0
 - Architecture: X64
 - Application Version: 10.2012.21.0
 - Region: en-US

**Additional context**
Similar issue: #860 

**Requested Assignment**

> A team member has edited this comment to update repro steps for clarity.

achmurali:
- Added a `AllowFocusOnInteraction` property for `ShiftButton`. 

- Manually Tested

### DIFF
--- a/src/Calculator/Views/CalculatorScientificOperators.xaml
+++ b/src/Calculator/Views/CalculatorScientificOperators.xaml
@@ -866,7 +866,9 @@
                       Checked="ShiftButton_Check"
                       Content="&#xF897;"
                       IsEnabledChanged="ShiftButton_IsEnabledChanged"
-                      Unchecked="ShiftButton_Check"/>
+                      Unchecked="ShiftButton_Check"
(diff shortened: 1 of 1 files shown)


joseartrivera (approved):
Good find! Thank you for fixing this :)

PR #1529, opened 2021-04-16
Added KeyDown Event Handler for Navigation Pane.

### LINKED ISSUE #1463: Keyboard user have to press 'Enter/Space' key twice to activate the list items

achmurali:
- Handle KeyDown Event for NavPane.
- Modify the Selected Item Binding on KeyDown, using FocusManager

- Manually Tested


joseartrivera (commented):
This seems to be an issue with WinUI. It seems like it is fixed if I update to a newer version. We plan on updating it soon so I think we could probably wait on that update to fix the bug.

PR #1542, opened 2021-05-14
Context Flyout fix for Calculation results

### LINKED ISSUE #1534: Backspace on "Select All" doesn't clear the whole expression
**Describe the bug**

Pressing `Backspace` after selecting the whole content of `CalculationResult` only removes the previous entered digit.

**Steps To Reproduce**

1.Enter some number
2.Click `Select All` on the flyout that appears after right clicking in the result area or Press `Ctlr + A` (after focusing in the result area)
3.Press backspace key or click backspace button on keyboard.

**Expected behavior**

Should clear the whole content.

 - OS Build: 10.0.19041
 - Architecture: x64
 - Application Version: 
 - Region: en-IN
 - Dev Version Installed: True

**Additional context**

Similar to "CE" behavior.

**Requested Assignment**

If possible and if applicable, I would like to fix this.

achmurali:
- Removed to right click flyout for the internal `TextBlock` as the default behavior of the `Copy` in rest of the `CalculationResult` right click flyout is the same. 
- The other reason to remove the `TextBlock`'s right-click flyout was the style was inconsistent with the rest of the app.

### The style of the right-click flyout for the internal `TextBlock`.
image

### The style for rest of the flyouts in the app.
image

- Manually Tested.

This solution tackles the issue by removing the `Select All` flyout functionality.

### DIFF
--- a/src/Calculator/Controls/CalculationResult.cpp
+++ b/src/Calculator/Controls/CalculationResult.cpp
@@ -127,6 +127,7 @@ void CalculationResult::OnApplyTemplate()
         m_textBlock = dynamic_cast<TextBlock ^>(GetTemplateChild("NormalOutput"));
         if (m_textBlock)
         {
+            m_textBlock->ContextMenuOpening += ref new ContextMenuOpeningEventHandler(this, &CalculationResult::OnContextMenuOpening);
             m_textBlock->Visibility = ::Visibility::Visible;
             m_textBlockSizeChangedToken = m_textBlock->SizeChanged += ref new SizeChangedEventHandler(this, &CalculationResult::OnTextBlockSizeChanged);
         }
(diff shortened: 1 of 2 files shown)


joseartrivera:
Thanks for the fix! I agree with the change. @grochocki double checking with you that the correct fix should be to remove Select All as an option and not to actually allow users to delete all input after selecting all.

grochocki:
If the user selects the text via click & highlight, the same issue occurs, so I don't think removing the context menu addresses the underlying issue. Is it possible to detect when input is selected (partial or full) and delete the selected portion?

sanderl:
We're converting Calculator to C# with a PR for the conversion coming in the next few days, so we're going to close this PR. If you want to revisit the issue, please open a new PR after the conversion has completed.

PR #1553, opened 2021-05-28
Add internal CI pipeline

mcooley:
Add a build which can be used to verify features like graphing calculator without signing and releasing the build.


tian-lt (approved):
LGTM. Thanks for helping complete our pipelines.

PR #1567, opened 2021-06-06
[C# Calc]Removes WinMeta.cs

tian-lt:
- Removes WinMeta.cs
- Add a method into Utilities to provide a constant defined in native codebase

- Manually Tested.


tian-lt:
Hi folks, could you help review this PR? thanks!

wmx16835:
Why removing the definition of the constants on the C# side?

tian-lt:
*WinMeta.cs* is created as a temporary solution to unblock C# migration works and it is mapping to *winmeta.h*.
Now it seems there is only **ONE** consuming point that is depending on the value of the macro `WINEVENT_KEYWORD_RESPONSE_TIME`.
Hence, it's not a good idea to maintain such strange file for the C# codebase.

wmx16835:
In this case, I agree to remove other redundant definitions (and the class), but would also suggest to keep this specific constant on the C# side as a static field. It does not seem to be necessary to have another function on the C++ side only to produce this constant.

guominrui (approved):
Looks good to me.

tian-lt:
Yes, putting a constant in C# side is good for performance and is in clean code. However, it means we need to understand and make sure the new constant is exactly correct with the native value for now and for the future as well. (I know the value might hard to be changed and we can see the actual value in winmeta.h, but it is quite brute to do that).
That's why I'd prefer to put a utility method here to let the native layer to decide the value as what it originally was. This workaround might not be performance friendly but it should be fine here because `LogAppLifecycleEvent` isn't a too much performance sensitive method.

guominrui:
The constant is defined in winmeta.h which is from Windows SDK, in my opinion, it's reasonable to not copy/duplicate the definition manually in C# side.

hanzhang54 (approved):
LGTM

wmx16835 (approved):

PR #1576, opened 2021-06-10
Fixes the missed NULLs

tian-lt:
## Fixes the missed NULLs in UnitTests
Missing `CommandNULL` is causing violation of access, which will fail the tests randomly.

- Adds the missed `CommandNULL`

- Tested manually.

### DIFF
--- a/src/CalculatorUnitTests/HistoryTests.cpp
+++ b/src/CalculatorUnitTests/HistoryTests.cpp
@@ -413,139 +413,139 @@ namespace CalculatorFunctionalTests
 
         void HistoryStandardOrderOfOperations_1()
         {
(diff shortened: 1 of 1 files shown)


hanzhang54 (approved):
LGTM

wmx16835 (approved):
LGTM. Do you think we should create a task to clean up the code (i.e., replacing the MFC-style array to C++ `std::vector`)?

guominrui (approved):
LGTM

tian-lt:
@wmx16835 
> Do you think we should create a task to clean up the code (i.e., replacing the MFC-style array to C++ std::vector)?

Sounds good.
I created an issue(#1577) to track.

PR #1579, opened 2021-06-11
Fix the Missing Part in Unit Converter Constructor

hanzhang54:
Problem: 
- The OfflineBlock doesn't show up while network offline. 
- ESC doesn't clear the input as a keyboard shortcut.
- Issue with copy/paste menu.

- Added the code which were missed during migration into UnitConverter constructor.

- Tested manually.


guominrui (approved):
LGTM

PR #1585, opened 2021-06-15
Fixes: GraphingNumber doesn't work correctly

tian-lt:
## Fixes: GraphingNumber doesn't work correctly
Problem: The expressions input from number pad doesn't work.

- Checks out on null or empty for the retval of GetMathTextProperty()

- Tested manually


guominrui (approved):
LGTM

PR #1589, opened 2021-06-23
[C# Calc] Resolve some TODO items for C# conversion

tian-lt:
- renames: `LocalizationSettings::InitializeLocalizationSettings()` ➡️  `LocalizationSettings::Initialize`
- exports `TraceLogger::LogPlatformException` as `TraceLogger::LogPlatformExceptionInfo` since `Platform::Exception` doesn't cross ABI boundary.
- exports `TraceLogger::LogInputPasted`
- leaves some comments for `ViewModel::Common::Utilities`
- fixes an `if` condition in *GraphingSettings.xaml.cs*

- Tested manually

### DIFF
--- a/src/CalcViewModel/Common/TraceLogger.h
+++ b/src/CalcViewModel/Common/TraceLogger.h
@@ -83,12 +83,12 @@ namespace CalculatorApp::ViewModel::Common
         void LogVariableSettingsChanged(Platform::String ^ setting);
         void LogGraphSettingsChanged(GraphSettingsType settingsType, Platform::String ^ settingValue);
         void LogGraphTheme(Platform::String ^ graphTheme);
+        void LogInputPasted(CalculatorApp::ViewModel::Common::ViewMode mode);
+        void LogPlatformExceptionInfo(CalculatorApp::ViewModel::Common::ViewMode mode, Platform::String ^ functionName, Platform::String ^ message, int hresult);
 
         internal:
-        // CSHARP_MIGRATION: TODO:
         void LogPlatformException(CalculatorApp::ViewModel::Common::ViewMode mode, Platform::String ^ functionName, Platform::Exception ^ e);
         void LogStandardException(CalculatorApp::ViewModel::Common::ViewMode mode, std::wstring_view functionName, _In_ const std::exception& e);
-        void LogInputPasted(CalculatorApp::ViewModel::Common::ViewMode mode);
 
     private:
         // Create an instance of TraceLogger
--- a/src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml.cs
+++ b/src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml.cs
@@ -110,11 +110,9 @@ namespace CalculatorApp
         {
             if (e.Key == VirtualKey.Enter)
             {
-                // CSHARP_MIGRATION: TODO:
-                // check if condition
-                if (FocusManager.TryMoveFocusAsync(FocusNavigationDirection.Next).Status != AsyncStatus.Completed)
+                if (FocusManager.TryMoveFocusAsync(FocusNavigationDirection.Next) == null)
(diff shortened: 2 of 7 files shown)


tian-lt:
Hi folks,
@hanzhang54 @guominrui @wmx16835 Could you help review this PR? thanks.

src/CalcViewModel/Common/TraceLogger.h
```diff
@@ -83,12 +83,12 @@ namespace CalculatorApp::ViewModel::Common
         void LogVariableSettingsChanged(Platform::String ^ setting);
         void LogGraphSettingsChanged(GraphSettingsType settingsType, Platform::String ^ settingValue);
         void LogGraphTheme(Platform::String ^ graphTheme);
+        void LogInputPasted(CalculatorApp::ViewModel::Common::ViewMode mode);
+        void LogPlatformExceptionInfo(CalculatorApp::ViewModel::Common::ViewMode mode, Platform::String ^ functionName, Platform::String ^ message, int hresult);
 
         internal:
-        // CSHARP_MIGRATION: TODO:
         void LogPlatformException(CalculatorApp::ViewModel::Common::ViewMode mode, Platform::String ^ functionName, Platform::Exception ^ e);
```
wmx16835, line 90: What's the difference between this one and `LogPlatformExceptionInfo`?

tian-lt: `LogPlatformExceptionInfo` only cares about message and hresult of a exception.
Though `LogPlatformException` seems to be handling more details than `LogPlatformExceptionInfo`, it does the same work.
The way of introducing `LogPlatformExceptionInfo` is because `Platform::Exception` doesn't across the ABI boundary between C# and C++/CX.

wmx16835: > Platform::Exception doesn't across the ABI boundary between C# and C++/CX.

I don't get it, since `Platform::Exception` is not a C++ native type. Is there anything special in `Platform::Exception`?

src/Calculator/Views/GraphingCalculator/GraphingSettings.xaml.cs
```diff
@@ -110,11 +110,9 @@ private void GridSettingsTextBox_PreviewKeyDown(object sender, KeyRoutedEventArg
         {
             if (e.Key == VirtualKey.Enter)
             {
-                // CSHARP_MIGRATION: TODO:
-                // check if condition
-                if (FocusManager.TryMoveFocusAsync(FocusNavigationDirection.Next).Status != AsyncStatus.Completed)
+                if (FocusManager.TryMoveFocusAsync(FocusNavigationDirection.Next) == null)
```
guominrui, line 113: Per the original condition, should we use *asyncResult == null ||  asyncResult.Status != AsyncStatus.Completed*?

tian-lt: here's the original condition in C++/CX
```
if (!FocusManager::TryMoveFocusAsync(FocusNavigationDirection::Next))
{
  FocusManager::TryMoveFocusAsync(FocusNavigationDirection::Previous);
}
```
It's weird but I don't think it was equivalent to `asyncResult.Status != AsyncStatus.Completed`

guominrui (approved):
Looks good to me.

hanzhang54 (approved):
LGTM

src/CalcViewModel/Common/TraceLogger.h
tian-lt: AFAIK, there's no equivalent implementation in C# that is mapped to Platform::Exception.
So, when C# compiler found a declaration of `Platform::Exception` it will ask for the way of linking it to an implementation, which is what we are not able to provide.

wmx16835 (approved):
LGTM

PR #1602, opened 2021-07-09
Fix compiler warnings

gAlfonso-bit:


src/CalcManager/Ratpack/num.cpp
```diff
@@ -602,20 +597,19 @@ bool lessnum(_In_ PNUMBER a, _In_ PNUMBER b)
...
-    int32_t length;
-    MANTTYPE* pcha;
-    length = a->cdigit;
-    pcha = a->mant;
+    MANTTYPE* pcha = a->mant;
 
     // loop over all the digits until you find a nonzero or until you run
     // out of digits
-    while (length-- > 0)
+    for (int32_t length = a->cdigit; length > 0; length--)
```
tian-lt, line 604: nit:
1. ln604: `length--` -> `--length`
2. ln612: `pcha++` -> `++pcha`

tian-lt: same for the other places that have been touched.

src/CalcManager/Ratpack/conv.cpp
```diff
@@ -200,7 +200,7 @@ PNUMBER _createnum(_In_ uint32_t size)
     if (SUCCEEDED(Calc_ULongAdd(size, 1, &cbAlloc)) && SUCCEEDED(Calc_ULongMult(cbAlloc, sizeof(MANTTYPE), &cbAlloc))
         && SUCCEEDED(Calc_ULongAdd(cbAlloc, sizeof(NUMBER), &cbAlloc)))
     {
-        pnumret = (PNUMBER)zmalloc(cbAlloc);
+        pnumret = reinterpret_cast<PNUMBER>(zmalloc(cbAlloc));
```
tian-lt, line 203: nit: instead of doing `reinterpret_cast` everywhere for `zmalloc()`, may be we should consider making `zmalloc()` more friendly to use. Such as 
```
// declaration of zmalloc 
template<typename T_Ptr>
T_Ptr zmalloc(size_t a);

// usage
auto pnumret = zmalloc<PNUMBER>(cbAlloc);
```

Besides, since `calloc()` is a C-Style API, both doing `reinterpret_cast` for `void*` and C-Style type casting look good to me. @mingxwa do you have suggestions about the practices in such case?

mingxwa: 1. Since this function allocates raw memory, I would vote for pure C-Style API to remain simplicity.
2. If a C++ style API is preferred, I would suggest to design the abstraction layer based on *types* instead of raw memory, e.g., the [operator new](https://en.cppreference.com/w/cpp/language/new).
3. [`static_cast`](https://en.cppreference.com/w/cpp/language/static_cast) is sufficient to convert `void*` to another pointer type. `reinterpret_cast` is not needed in this case.

Issue #312, opened 2019-03-15
Focus moves to "Clear all memory" button after activating "Clear memory item" from the context menu of "Memory" list item.


MicrosoftIssueBot:
A team member will update this issue with more details soon.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I created this issue automatically as requested by a team member.

LauraLaureus:
Can I grab this as part of the Hacktoberfest ?

grochocki:
Sure thing! Thanks for taking a look.

PR #1605, opened 2021-07-13
Fix Focus bug by proactively moving focus before the ListView is updated

zherman:
## Fixes #.
#312 

The reason why focus never went to a sensible place after removing an item from the memory list is because focus was on the flyout menu when the item got removed. Focus was then re-evaluated when the flyout menu went away. The XAML Focus Manager in this scenario will try to move focus back to what had it before the flyout menu was invoked, but we removed that item.

The fix implemented in this change proactively moves focus to the correct element right before mutating the memory item list. This is accomplished in 2 steps. First, we hide the Flyout menu so that the XAML Focus Manager moves focus back into the ListView. Second, we move Focus Up or Down depending on which is more reasonable (Up if the last element is being removed otherwise Down).

- Manually went through the repro steps and removed an item in the middle of the memory list (with a list size > 3). Saw that focus was on the expected memory item after the remove.
- Manually went through the repro steps and removed the last memory item (with a list size > 3). Saw that focus moved to the new last item after the removal.
- Manually went through the repro steps and removed the first memory item when the list size is of size 1. Saw that focus moved to the XAML element above where the empty list is (The XAML element that contains the Header and Memory buttons)

### DIFF
--- a/src/Calculator/Views/Memory.xaml.cpp
+++ b/src/Calculator/Views/Memory.xaml.cpp
@@ -57,9 +57,28 @@ void Memory::MemoryListItemClick(_In_ Object ^ sender, _In_ ItemClickEventArgs ^
 
 void Memory::OnClearMenuItemClicked(_In_ Object ^ sender, _In_ RoutedEventArgs ^ e)
 {
+    // We are proactively hiding the context menu so that XAML focus goes back to a ListViewItem.
+    // If we let the context menu hide on its own then it would lose focus and get hidden after the
+    // coresponding memory item is removed. The XAML Focus Manager will then try to set focus to the
+    // item that no longer exists and will end up setting focus to the root of the ListView
+    // because the item that XAML wants to give focus to does not exist.
+    MemoryContextMenu->Hide();
+
     auto memoryItem = GetMemoryItemForCurrentFlyout();
     if (memoryItem != nullptr)
     {
+        // Move focus to a neighboaring item before removing memoryItem. Ideally the same index in
+        // the list will be selected after the remove operation, but if we remove the last item then
+        // we want the new last item to have focus.
+        auto items = MemoryListView->Items->GetView();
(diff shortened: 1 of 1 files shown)


guominrui:
@zherman Please resolve the conflicts, as the UI part has migrated to C#.

src/Calculator/Views/Memory.xaml.cpp
```diff
@@ -57,9 +57,28 @@ void Memory::MemoryListItemClick(_In_ Object ^ sender, _In_ ItemClickEventArgs ^
 
 void Memory::OnClearMenuItemClicked(_In_ Object ^ sender, _In_ RoutedEventArgs ^ e)
 {
+    // We are proactively hiding the context menu so that XAML focus goes back to a ListViewItem.
+    // If we let the context menu hide on its own then it would lose focus and get hidden after the
+    // coresponding memory item is removed. The XAML Focus Manager will then try to set focus to the
```
jogo-, line 62: Typo: coresponding -> cor**r**esponding

src/Calculator/Views/Memory.xaml.cpp
```diff
@@ -57,9 +57,28 @@ void Memory::MemoryListItemClick(_In_ Object ^ sender, _In_ ItemClickEventArgs ^
...
+    // item that no longer exists and will end up setting focus to the root of the ListView
+    // because the item that XAML wants to give focus to does not exist.
+    MemoryContextMenu->Hide();
+
     auto memoryItem = GetMemoryItemForCurrentFlyout();
     if (memoryItem != nullptr)
     {
+        // Move focus to a neighboaring item before removing memoryItem. Ideally the same index in
```
jogo-, line 70: Typo: neighbo**a**ring -> neighboring

mcooley:
Closing this PR since it's no longer up to date with the latest changes to migrate the project to C#.

PR #1615, opened 2021-07-22
Fixed the display issue in AOT mode

hanzhang54:
### Description of the issue:
After Calculator enters AOT mode, it looks good to change the window size with the UI controls’ size changed coordinately when the window size is large. However, if we keep decreasing the window size, we will see the UI control’s height is always automatically adapted to the window size while the width stops adapting to smaller size at a certain point. 

This issue is caused by the WinUI 2.6 NavigationView. A workaround is to set the OpenPaneLength to be smaller than the minimum window width in AOT mode.

Passed the build and manually tested different window size in AOT mode.

### DIFF
--- a/src/Calculator/Views/MainPage.xaml.cs
+++ b/src/Calculator/Views/MainPage.xaml.cs
@@ -604,6 +604,11 @@ namespace CalculatorApp
             }
         }
 
+        private double NavigationViewOpenPaneLength(bool isAlwaysOnTop)
+        {
+            return isAlwaysOnTop ? 0 : 256;
+        }
+
         private CalculatorApp.Calculator m_calculator;
         private GraphingCalculator m_graphingCalculator;
         private CalculatorApp.UnitConverter m_converter;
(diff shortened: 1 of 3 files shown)


src/Calculator/Views/MainPage.xaml.cs
```diff
@@ -604,6 +604,11 @@ private static void SetSettingsButtonAccessKey(MUXC.NavigationView navView)
             }
         }
 
+        private double NavigationViewOpenPaneLength(bool isAlwaysOnTop)
+        {
+            return isAlwaysOnTop ? 0 : 256;
```
guominrui, line 609: I suggest we still have the panel length defined in App.xaml, and here read it by  `Application.Current.Resources["SplitViewOpenPaneLength"])` .

hanzhang54: Done.

guominrui (approved):
LGTM

PR #1619, opened 2021-07-27
Disable Mica before close the window to resolve a crash

guominrui:
Disable Mica before closing window to resolve the crash in multi-view case.

Manually tested.


hanzhang54 (approved):
LGTM

PR #1633, opened 2021-08-04
Fix the inconsistent behavior between narrator and keyboard focus

hanzhang54:
In the function analysis of graphing mode, some controls use system focus visuals for keyboard focus while the UIA is not set up. In this case, the narrator will focus on the last UIA control when the keyboard focus on a control with UIA not available.

Update the keyboard focus and UIA on the ListView to align their behavior.

Passed the unit test as well as UI test, and manually tested.


guominrui (approved):
LGTM

PR #1638, opened 2021-08-09
Fixes up some names of dependency property

tian-lt:
- Fixes some wrong names of dependency property while registering them.

- Tested manually


guominrui (approved):
LGTM

Issue #114, opened 2019-03-06
Don't clear input when switching modes


Diggsey:
**Problem Statement**
Often I need to use functions from both programmer and scientific modes in the same calculation, but if I switch modes it loses my input. When the input is representable in both modes there is no reason to clear it.

**Evidence or User Insights**
No other calculator does this, and there's no need for the input to be cleared. There's a button to clear the input.

**Proposal**
Preserve the value when it is within the domain of the target mode, otherwise show an error in the display area.

**Goals**
Remove the annoyance of losing the value when you switch modes.

**Non-Goals**
Settle on a unified representation of values such that separate "modes" are not even necessary.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

dgendreau:
Agreed. I believe the Windows 7 calculator kept the current value when switching modes and this has always been an annoyance for me with the w10 calculator.

grochocki:
Thanks for submitting your idea! Let's keep this idea open for discussion so the community has the chance to provide feedback. Check out our [New Feedback Process](https://github.com/Microsoft/calculator/blob/master/docs/NewFeatureProcess.md) for more info on the user-centered process we follow for new feature development.

rba100:
Might be good to think of the use case for switching modes, just in case the problem can be solved another way. For me, the use case is to see the hex or binary representation of a number having got there from scientific, which is generally my default mode because it supports the power and log keyboard shortcuts. Personally, I think persisting the values (and truncating if necessary) between modes is a sensible thing to do in any case.

ygoe:
@dgendreau AFAIR the Windows 7 calculator didn't have the separate scientific and programmer modes in the first place. Dec/hex view was possible in the same view as 1/x. So there was no need to switch modes. When I first saw the Windows 10 calculator, I immediately felt like I wanted back the Windows 7 calculator. By now, I somehow came to an arrangement with it, but I'm not really happy with that experience.

To Microsoft: My telemetry data (I only learned through this GitHub project that the calculator has that!) should have told you that already. Not sure you heard it because nothing has changed.

seanofw:
I can't second this feature request enough.  My use case for this feature is a bit different from @rba100's; so here's my "Agile user story," if you like developing software around such things:

[...]

PR #1654, opened 2021-08-18
Fix the result in wrong number base while leaving and switching back …

hanzhang54:
…to Programmer

Issue: When switching from programmer to graphing/date/converters and switching back, 
1) the hex buttons are disabled even HEX in selection;
2) the number base selection is the same as when leaving but the result value is always in DEC;
3) the memory values are always displayed in DEC even the number base is not DEC when leaving programmer.

Rename ResetDisplay() to ResetRadixAndUpdateMemory(). Reset radix to DEC for standard and scientific modes, and only update memory for programmer mode.

Passed the unit tests and manually tested.

### DIFF
--- a/src/CalcViewModel/StandardCalculatorViewModel.cpp
+++ b/src/CalcViewModel/StandardCalculatorViewModel.cpp
@@ -1188,14 +1188,14 @@ void StandardCalculatorViewModel::SetCalculatorType(ViewMode targetState)
     {
     case ViewMode::Standard:
         IsStandard = true;
-        ResetDisplay();
+        ResetRadix();
         SetPrecision(StandardModePrecision);
         UpdateMaxIntDigits();
         break;
 
     case ViewMode::Scientific:
         IsScientific = true;
-        ResetDisplay();
+        ResetRadix();
         SetPrecision(ScientificModePrecision);
         break;
 
@@ -1226,7 +1226,7 @@ String ^ StandardCalculatorViewModel::GetLocalizedStringFormat(String ^ format, 
     return LocalizationStringUtil::GetLocalizedString(format, displayValue);
 }
 
-void StandardCalculatorViewModel::ResetDisplay()
+void StandardCalculatorViewModel::ResetRadix()
 {
     AreHEXButtonsEnabled = false;
     CurrentRadixType = NumberBase::DecBase;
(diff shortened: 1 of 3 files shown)


guominrui (dismissed):
This fix looks good to me. But I suggest updating the ResetDisplay method name to be more specific, per the implementation of this method, it just resets radix related states.

guominrui (dismissed):

tian-lt (approved):
Though I suppose there might be some different logics that don't re-initialize programmer mode, which caused this bug, it looks good to me to have such minimum changes to fix this issue. But if there's any chance, it would be great if we can spend some efforts to take a closer look to below cases that reset programmer mode differently(comparing with switching from standard mode to programmer mode)
 - From graphing mode to programmer mode.
 - From Date Calculation mode to programmer mode.

hanzhang54:
> Though I suppose there might be some different logics that don't re-initialize programmer mode, which caused this bug, it looks good to me to have such minimum changes to fix this issue. But if there's any chance, it would be great if we can spend some efforts to take a closer look to below cases that reset programmer mode differently(comparing with switching from standard mode to programmer mode)
> 
> * From graphing mode to programmer mode.
> * From Date Calculation mode to programmer mode.



> Though I suppose there might be some different logics that don't re-initialize programmer mode, which caused this bug, it looks good to me to have such minimum changes to fix this issue. But if there's any chance, it would be great if we can spend some efforts to take a closer look to below cases that reset programmer mode differently(comparing with switching from standard mode to programmer mode)
> 
> * From graphing mode to programmer mode.
> * From Date Calculation mode to programmer mode.

The standard, scientific and programmer share the same view mode. There is one issue https://github.com/microsoft/calculator/issues/114 which describes the cleared input when switching among those three modes. So when switching from standard to programmer as you said, the result will get cleared. 

However, this is not the same issue addressed here. The problem here is when switching from graphing/date/converters to programmer, the result is retained but is not in the same number base as the selected number base if the selected number base is not DEC when leaving programmer.

The reason we do ResetDisplay() to standard and scientific is that standard and scientific modes only support DEC number base. In ResetDisplay(), it will disable the HEX buttons, set current radix type to DEC number base, set radix to DEC number base for CalculatorManager, and update numbers in memory to current radix. These operations are not needed when switching to programmer mode. So it would be a better way to correct the logic here by removing it.

hanzhang54:
The memory is shared among standard, scientific, and programmer modes. The memory in programmer should not support the decimal number with point while the memory in standard and scientific modes should.

Removing the ResetDisplay() when switching to programmer would cause another issue. In this case, the decimal number with point saved in standard/scientific mode will keep the point in programmer. To avoid this, rename ResetDisplay() to ResetRadixAndUpdateMemory(), discard the radix resetting but keep the memory updating part in programmer mode.

guominrui (approved):
LGTM

PR #1659, opened 2021-08-20
Restructure pipeline to move release steps out of Package ES

quentin987:
Restructuring Pipeline release steps

Queued Release build through ADO. 
Note: Aero step is currently failing, potentially because of an issue on the Package ES side. This is expected behavior at this point and we'll address when we remove that step from the build.


build/pipelines/templates/build-single-architecture.yaml
```diff
@@ -1,10 +1,40 @@
...
+
+  - ${{ if eq(parameters.isReleaseBuild, true) }}:
+    - task: UniversalPackages@0
+      displayName: Download internals package
+      inputs:
+        command: download
+        downloadDirectory: $(Build.SourcesDirectory)
+        vstsFeed: WindowsInboxApps
+        vstsFeedPackage: calculator-internals
+        ${{ if eq(parameters.isPublicRelease, true) }}:
```
mcooley, line 33: I think `isPublicRelease` can be removed everywhere, since it means the same thing as `isReleaseBuild`.

PR #1679, opened 2021-08-31
Fix for #1666- Back button is not accessible via keyboard when navigating to the about page.

### LINKED ISSUE #1666: Back button is not accessible via keyboard when navigating to the about page.

tian-lt:
## Fixes #1666
Back button is not accessible via keyboard when navigating to the about page.

- Moves the `BackButton` from titlebar to Settings page.
- Changes correlated code-behind

- Build passed and tested manually

### Remarks:
The static UI won't have any difference, but there is a little difference with the animation of BackButton.
Before the fix: 
before

After the fix:
after


src/Calculator/Views/MainPage.xaml.cs
```diff
@@ -15,6 +15,8 @@
 using Windows.UI.ViewManagement;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Automation;
+using Windows.UI.Xaml.Automation.Peers;
```
hanzhang54, line 18: Could you remove these in TitleBar.xaml.cs if they are no longer used?

src/Calculator/Views/MainPage.xaml
```diff
@@ -163,16 +162,31 @@
...
             <Popup.ChildTransitions>
                 <TransitionCollection>
                     <EntranceThemeTransition/>
                 </TransitionCollection>
             </Popup.ChildTransitions>
 
             <Grid x:Name="PopupContent" x:Load="false">
-                <local:Settings/>
+                <Grid.RowDefinitions>
+                    <RowDefinition Height="32"/>
```
hanzhang54, line 174: The height of the title bar row here is 32 while the Auto height of the title bar row in TitleBar.xaml would be 20 with your change.

It would be better to define the title bar height in the application resource to make the grid row height here align with the row height in TitleBar.xaml.

guominrui:
@tian-lt , suggest moving the xaml ui codes about Back button into local:Settings control, local:Settings control will be responsible for managing/displaying all the UI parts in settings page, Main page only controls the popup show/hide.

src/Calculator/Views/MainPage.xaml
guominrui: Suggest binding the row height with title bar height.

guominrui: And please search "PopupContent.Height" in Mainpage.xaml.cs, you will see some other lines about setting the popup height, you need to update accordingly.

tian-lt: Changes in the latest commit:
 - Move the back-button to settings control.
 - Bind the height of back-button with title bar
 - Remove unused items.

guominrui:
@tian-lt please take a look at https://github.com/microsoft/calculator/blob/f46f63f340e959090cc5d9ee79e6b7484ba7d6f3/src/Calculator/Views/MainPage.xaml.cs#L157 and https://github.com/microsoft/calculator/blob/f46f63f340e959090cc5d9ee79e6b7484ba7d6f3/src/Calculator/Views/MainPage.xaml.cs#L303 , the PopupContent.Height assignment should be changed to " = windowBounds.Height". Otherwise, the popup height will not be enough to cover the window.

Except this one, other changes look good to me.

src/Calculator/Views/MainPage.xaml.cs
tian-lt: done

tian-lt:
Nice catch, @guominrui.
Fixed the popup content height in the last commit.

hanzhang54 (approved):
LGTM

guominrui (approved):
Nice.

PR #1680, opened 2021-08-31
Specify which class's function to call during constructor

AtariDreams:
Especially since the two functions in question are virtual

- Specify to call the superclass's version of the functions during construction, since it is not defined in the subclasses

- Manual testing
- Passed all tests

### DIFF
--- a/src/CalcManager/UnitConverter.cpp
+++ b/src/CalcManager/UnitConverter.cpp
@@ -65,8 +65,8 @@ UnitConverter::UnitConverter(_In_ const shared_ptr<IConverterDataLoader>& dataLo
     unquoteConversions[L"{sc}"] = L';';
     unquoteConversions[L"{lb}"] = LEFTESCAPECHAR;
     unquoteConversions[L"{rb}"] = RIGHTESCAPECHAR;
-    ClearValues();
-    ResetCategoriesAndRatios();
+    UnitConverter::ClearValues();
+    UnitConverter::ResetCategoriesAndRatios();
 }
 
 void UnitConverter::Initialize()


AtariDreams:
An updates? @tian-lt

src/CalcManager/UnitConverter.cpp
```diff
@@ -65,8 +65,8 @@ UnitConverter::UnitConverter(_In_ const shared_ptr<IConverterDataLoader>& dataLo
     unquoteConversions[L"{sc}"] = L';';
     unquoteConversions[L"{lb}"] = LEFTESCAPECHAR;
     unquoteConversions[L"{rb}"] = RIGHTESCAPECHAR;
-    ClearValues();
-    ResetCategoriesAndRatios();
+    UnitConverter::ClearValues();
```
tian-lt, line 68: I agree with Mingxin's comment. Couldn't see the motivation to get rid of polymorphism here.
Could you share more context?

tian-lt (changes requested):

src/CalcManager/UnitConverter.cpp
AtariDreams: There is only one ClearValues function anyway, despite it being virtual. Polymorphism is not used for this class at all, so either we specify class or these functions should not be virtual

mingxwa: Checking the code, `ClearValues()` and `ResetCategoriesAndRatios()` are not declared `virtual` at all. `ClearValues()` has nothing to do with polymorphism. `ResetCategoriesAndRatios` overrides the interface function but does not allow subclasses to override, therefore there's no polymorphism when calling `ResetCategoriesAndRatios()` from `UnitConverter` at runtime.

mingxwa (changes requested):

PR #1683, opened 2021-09-07
Make Calculator the default startup project

mcooley:
Fix two issues with the solution file:
- List the "Calculator" project first, which causes Visual Studio to choose it as the startup project when the repo is first checked out. This makes it easier for people to get started coding in this repo.
- Don't nest the "CalcManager" project under the "Solution Items" folder.

### DIFF
--- a/src/Calculator.sln
+++ b/src/Calculator.sln
@@ -9,6 +9,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		nuget.config = nuget.config
 	EndProjectSection
 EndProject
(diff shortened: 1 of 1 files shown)


tian-lt (approved):
LGTM.
It's indeed easy to forget switching the launch project to Calculator project every first time.
BTW, ARM64 as the default configuration is also disturbing while coding on Windows.

mcooley:
> LGTM.
> It's indeed easy to forget switching the launch project to Calculator project every first time.
> BTW, ARM as the default configuration is also disturbing while coding on Windows.

Yep, unfortunately I'm not aware of a way to set the default platform to x64.

PR #1686, opened 2021-09-08
UI tests should use CI runsettings in internal CI builds

mcooley:
After the change in #1682, the internal CI build uses the .Dev app identity. Use the runsettings for UI tests that has the .Dev app identity.


hanzhang54 (approved):
LGTM

PR #1729, opened 2021-10-19
Drop non-negative tests on unsigned values

janisozaur:
As some of the variables are of unsigned types, it makes no sense to
test them for non-negativeness, as this is enforced by type already.


mcooley (approved):
Sorry it's taken so long to review. Thanks for this cleanup!

PR #1741, opened 2021-11-06
Added description when mouseover on ce and c.

Flre-fly:
-
-
-

-
-
-


mcooley:
It looks like you're suggesting adding a tooltip to the "C" and "CE" buttons. This would be a significant change compared to the current design, and we would prefer to follow the [new feature process](https://github.com/microsoft/calculator/blob/master/docs/NewFeatureProcess.md) for this.

Personally, I think tooltips on these buttons could be distracting more often than they would be helpful. If you think there are reasons we should add tooltips that we haven't considered, feel free to open an issue in this repo.

Issue #379, opened 2019-03-25
Proposal to modernize the architecture of Unit Conversions


rudyhuyn:
# Problem Statement

There are two flaws with the way the calculator currently manages unit conversions: 
- The way unit conversion is managed doesn’t allow us to support new scenarios (future but also already identified by the community), restricting the possibilities and innovations.
- It's obvious, the current code has a long history and some technical choices from the past don’t make sense anymore, making the code inefficient and challenging to understand/modify.

Here is a proposal for improvements to calculator management of unit conversions, and for clearing the path for new features going forward.

This proposal is a work-in-progress, some of the changes will lead to a big refactoring of a part of the application. Before even starting to modify how unit conversion works, let's list all the future needs we will have to be sure to suit all of them and let's debate about the different actions in comments. Don't hesitate to share your ideas and feedback, this document will be regularly updated to take into account your suggestions.

# Goal

This proposal is a work-in-progress, some of the changes will lead to a big refactoring of a part of the application. Before even starting to modify how unit conversion works, let's debate and list all the future needs we will have to be sure to suit all of them and let's debate about the different actions in comments. Don't hesitate to share your ideas and feedback, this document will be regularly updated to take into account your suggestions.

# Proposal

## Simplify Data Loader + merge Unit and UnitData.

When the user opens one of the converters, `UnitConverterViewModel` creates a `UnitConverter` object, hosting 2 Data loaders: `UnitConverterDataLoader` and `CurrencyConverterDataLoader`. To simplify our analysis, we can ignore the second data loader, but there are still many issues from the first data loader.

The `UnitConverterDataLoader` hosts 2 main objects: 

- A `map<CategoryId, vector<Unit>>`:
  - `Unit` objects containing:
    - Unit Id
    - Name (string)
    - Abbreviation (string)
    - (+ an order via `OrderedUnit`)
- A list of `UnitData`:
  - `UnitData` objects containing:
    - Category Id
    - Unit Id
    - A Factor (double) representing the value of the unit measured in an unspecified base unit used by all UnitData from the same category.

Even if these 2 collections are static and hosted by the same class, they have data redundancy, with the association Category↔Unit being defined by the `map<CategoryId, Unit>` but also by the `UnitData`.

The only valuable information hosted by `UnitData` is `Factor` and we have no reason to prevent us from merging `Unit` and `UnitData` and creating a single object containing all the necessary information to fully describe a unit:

| Unit                  |
|-----------------------|
| Unit Id               |
[...]

PR #1743, opened 2021-11-15
Code cleanup: Use number formatting utilities from the standard library

mcooley:
We should avoid using the `NumberFormattingUtils` class in CalcManager. The standard libraries in both C# and C++ offer number-formatting functionality, and we should use that where possible.

This PR removes all use of NumberFormattingUtils from Graphing Calculator code, so the only remaining code that uses it is in unit converter. Unit converter has many other problems (see #379), so I didn't attempt to refactor it in this PR.

There are three areas that changed:
- Remove "fixed point with trimmed trailing zeros" formatting from the Graphing Calculator share output. Instead, use C#'s `double.ToString()`, which defaults to the general ("G") format. The notable change in behavior vs. the previous behavior is that scientific notation might be used if it enables a shorter output, although this should be rare since numbers of this magnitude are hard to input in the UI.
- Remove "fixed point with trimmed trailing zeros" formatting from the Graph Settings code. Instead, use the default format of wostringstream, which is also a "general"-type format. Again, this changes behavior so scientific notation could be used, so we also need to change the code where these numbers are parsed back to doubles to use wistringstream, which can handle the scientific notation format.
- Finally, move NumberFormattingUtils to the UnitConversionManager namespace. This indicates that it's intended for use only in the existing UnitConversionManager code and can be deleted if that code is refactored or removed.


rameeznaufal (approved):

mcooley:
This is a low-priority cleanup PR which is not urgent, but I'd also like to get it merged before I forget all the details about how it works :) @tian-lt @hanzhang54 @guominrui can one of you review this sometime soon?

hanzhang54 (approved):
Thanks Matt! It looks good to me. 😊

Issue #1748, opened 2021-11-28
Doesn't record on history list if I press '=' button before I press the operator button


yhkkkkxx:
**Describe the bug**

If I press '=' button before I press the operator button, the same number before press the '=' button continuously record on history list. I think it is so unnecessary, then I want to modify it so that doesn't record on history list. 

**Expected behavior**

doesn't record
 

**Additional context**

Can I try this?

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

mcooley:
It sounds like you don't want this behavior:

https://user-images.githubusercontent.com/296848/145929377-1db4512e-5c15-4f05-a781-c66046c438fa.mp4

It seems reasonable to me that we should avoid creating duplicate history entries in this case, but I'm curious what others think.

grochocki:
Thanks for the feedback! This is by design as <kbd>=</kbd> is always committing the last operation. Try pressing <kbd>5</kbd><kbd>+</kbd><kbd>2</kbd><kbd>=</kbd><kbd>=</kbd>.

PR #1750, opened 2021-12-04
don't allow to record on history list if I press '=' button before I …

yhkkkkxx:
…press the operator button #1748

- If you press equal(=) button without pressing any operator buttons, it would be not update on history item list.
- Parenthesis, 'e', 'π' that are numbers themselves also can't be recorded. 
- For example, if you input any numbers and press equal(=) button right away, then nothing is recorded.

- ad-hoc testing


mcooley:
Thanks for this PR! I'm not sure this is the best way to fix #1748, because I think there might be cases where we _do_ want to add to history even if the first token is not an operator. Also, ot looks like this might be breaking some unit tests, and we should add new unit tests which verify the new behavior.

I'm going to close this PR for now. Let's continue the discussion about the desired behavior in #1748. Once that issue is approved, we can open a PR again. Thanks!

PR #1761, opened 2021-12-15
Run loc pipeline on windows-2022 image

mcooley:
Run the localization pipeline on the windows-2022 image. We will be switching to windows-2022 and Visual Studio 2022 for the main app builds sometime soon. It doesn't really matter which image the loc pipeline runs on, but switching it to windows-2022 will help keep all pipelines consistent.

### DIFF
--- a/build/pipelines/azure-pipelines.loc.yaml
+++ b/build/pipelines/azure-pipelines.loc.yaml
@@ -23,7 +23,7 @@ name: $(BuildDefinitionName)_$(date:yyMM).$(date:dd)$(rev:rrr)
 jobs:
 - job: Localize
   pool:
-    vmImage: windows-2019
+    vmImage: windows-2022
   variables:
     skipComponentGovernanceDetection: true
   steps:


build/pipelines/azure-pipelines.loc.yaml
```diff
@@ -51,4 +51,4 @@ jobs:
     condition: eq(variables['hasChanges'], '1')
     inputs:
       artifactName: Patch
-      targetPath: $(Build.ArtifactStagingDirectory)
\ No newline at end of file
+      targetPath: $(Build.ArtifactStagingDirectory)
```
hanzhang54, line 54: The LocalizedStrings.patch is stored at $(Build.ArtifactStagingDirectory). Would removing the targetPath here affect this task's goal to publish patch file as artifact?

hanzhang54: Oops, please ignore my comment above. 😂 It looks good to me .

hanzhang54 (approved):

PR #1770, opened 2022-01-03
Update build year in settings page to 2022

mcooley:
Updates the build year in the settings page to 2022.

Also removes a comment which is unclear--the build year is not updated from the pipeline. Maybe that's something we could do in the future, but I don't think we need to mention it in the code since it's not really a regression from the C++ version of the code.


tian-lt (approved):
Looks good to me :)

hanzhang54 (approved):

PR #1778, opened 2022-01-20
Updated function input list and list item names for accessibility purposes

### LINKED ISSUE #1676: The Name property of 'Function input' list must not include the element's control type.
Repro steps:
Open the Application Calculator
Navigate to ‘Graphing’ button and activate it.
Now add some equation and Run accessibility Insight for Windows. 
Observe the issue.

Actual Result:
The Name property of the equation include the element's control type. Name is defined as 'Function input list'.

Expected Result:
The Name property of the equation must not include the element's control type. Name should be defined correctly as 'Function input' and should not include role 'list' in the name.

mateuszplaza:
- Updated name property of 'Function Input List' and 'Function Input List Item' to not include element's control type, as laid out in issue 1676.
- Update was reflected in resources for en-US and en-GB regions

-Tested on current test cases
-Used Windows Accessibility Insights to confirm change is corrected

### DIFF
--- a/src/Calculator/Resources/en-GB/Resources.resw
+++ b/src/Calculator/Resources/en-GB/Resources.resw
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <root>
   <!-- 
     Microsoft ResX Schema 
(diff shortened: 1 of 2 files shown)


mcooley (approved):
Thanks for this fix!

FYI--the only .resw files which need to be changed manually are the en-US ones. All other languages will be overwritten by an automated process.

hanzhang54 (approved):
Thanks for fixing it. 😊

Issue #1736, opened 2021-11-01
Compute result ahead of time instead when user input


moracabanas:
**Problem Statement**
When I use my physical calculator I can type `0,9` `*` `log` `(` `0.8` `)`
And it displays
0.9 * log(0.8)
so I can fix any typo and run `=` for get results

But, doing so it runs `log` at the time you type it so it ruins your formula understanding and you must rethinking its typing hierarchy 

i.e running this example ->  `0,9` `*` `log` `(` `0.8` `)` `=`
Expected behaviour about: 

![image](https://user-images.githubusercontent.com/4382333/139692348-f582ed05-26ca-4458-a0b2-7dd70875b67c.png)

Actual behabiour:

[![Image from Gyazo](https://i.gyazo.com/ecdd75f5b3ae2d743d0960667f350d87.gif)](https://gyazo.com/ecdd75f5b3ae2d743d0960667f350d87)

**Proposal**

As you can se If you want the expected behabiour you must compute `log(0,8) while typing as:
`0.9` `*` `(` `0.8` `log` `)` 
so, this changes the way formula is represented and its becabiour is so counter intuitive it turns out to be unusable.

In my honest opinion you could disable this behabiour or enable ahead of time computation so it lets you write a formula and compute result only when you hit `=`

If possible, I would like to implement this.
I'm just suggesting this idea.  I don't want to implement it.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

seancruise-1:
I would like to try to implement this.

PR #1786, opened 2022-01-31
Possible Patch for Issue #1736

seancruise-1:
##For School Project: Open Source Contribution
## Fixes #.
#1736 

- Changes it so that the log function can be typed first and then when you close the parentheses it solves the logarithm.  
- Made a global variable to keep track of so that it could toggle the log function. 
-

- Issues with change is that there are problems with the equal sign in some specific use cases. 
-


mcooley:
Thanks for your interest in contributing. It looks like the issue #1736 has not been approved yet. I am skeptical of the change suggested in that issue since it could be disruptive to existing workflows. We'd need to think about all the possible scenarios that would be affected besides just the `log` function. Let's allow further discussion to happen in that issue.

PR #1791, opened 2022-02-06
Don't maximize Calculator's window on some small screen devices

cheweilin:
Added an AppExtension to Calculator so it no longer maximize by default on small screen device.

- Added xmlns to `package.appxmanifest` file
- Updated TargetDeviceFamily's MinVersionNumber to `10.0.18308.0` to accommodate long AppExtension name
- Added `uap3:Extension` with AppExtesnion name of `com.microsoft.windows.dontmaximizeonsmallscreen`

- Calculator built and deployed to VM, tested manually


mcooley (changes requested):
Hi, thanks for contributing to Calculator!

- Please follow [the contributing guidelines for this repo](https://github.com/microsoft/calculator/blob/master/CONTRIBUTING.md). We generally expect PRs to be linked to an approved GitHub issue.
- The "com.microsoft.windows.dontmaximizeonsmallscreen" app extension is not documented anywhere. Since Calculator's source code is open to the public, we need some documentation posted publicly that helps other developers understand whether this approach is something they can copy this approach and what the consequences will be.
- If your team has any influence over this API--why an AppExtension? No other windowing behavior in UWP is controlled with AppExtension. Can a WinRT API be provided instead?
- You'll need to make changes in both Package.appxmanifest and Package.Release.appxmanifest. Package.Release.appxmanifest is the one that's used in Calculator's production build.

cheweilin:
- This is a task item created in ADO, should I create an issue for it?
- This will be a case for **Known App Extension** and documented on the docs.microsoft.com. I will follow up on this.
- Based off conversation with Chris Swan and Bret Anderson, Calculator uses system default size for its very first launch and then immediately resize it. The future launch will respect the last modified size as what other UWP apps will do. This behavior makes it difficult to detect preferred launching window size for small screen devices.  Therefore, we opted to go with using AppExtension.
- Updated `Package.Release.appxmanifest` file

src/Calculator/Package.appxmanifest
```diff
@@ -8,7 +8,7 @@
         <Logo>Assets\CalculatorStoreLogo.png</Logo>
     </Properties>
     <Dependencies>
-        <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17133.0" MaxVersionTested="10.0.22000.0" />
+        <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.18308.0" MaxVersionTested="10.0.22000.0" />
```
mcooley, line 11: nit: Let's just make this "18362" since that was a released version of Windows. I don't think anyone is going to test with insider builds earlier than 18362.

src/Calculator/Package.appxmanifest
```diff
@@ -36,6 +36,17 @@
                         <uap:Logo>Assets\CalculatorAppList.png</uap:Logo>
                     </uap:Protocol>
                 </uap:Extension>
+                <uap3:Extension Category="windows.appExtension">
+                    <uap3:AppExtension Name="com.microsoft.windows.dontmaximizeonsmallscreen"
+                                       Id="calculator"
+                                       DisplayName="calculator"
+                                       Description="This app extension makes calculator no longer maximizing by default on small screen device."
```
mcooley, line 43: nit: Let's say "This app extension prevents calculator from being maximized by default on some small screen devices."

In both versions of the manifest.

mcooley (approved):
We had an offline discussion with the team in Windows implementing this feature. We learned that unfortunately, while this app extension is not a very good API for app developers to express this behavior, it's the best method available right now. There is ongoing work to publicly document this app extension so other apps can use it with confidence.

mcooley (approved):

mcooley:
@cheweilin the CI is failing because currently our UI tests run on Windows Server 2019. We are planning a move to Windows Server 2022 very shortly, which should resolve the issue. I'll merge this PR after we've completed that migration.

cheweilin:
Resolved merge conflicts

PR #1793, opened 2022-02-09
Build with Visual Studio 2022

mcooley:
This change updates the Calculator project to build with Visual Studio 2022.

- Use Visual Studio 2022
- Use an Azure Pipelines image which has Visual Studio 2022 installed
- Use the v143 C++ build tools
- Use the Windows SDK version 22000
- Raise the minimum platform version to 17763
  - As a consequence of this change, we will use the "msix" file extension instead of "appx" for our packages
- Update UI tests from .NET Core 2.1 to .NET Core 3.1


build/pipelines/azure-pipelines.release.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
   parameters:
     platform: x86
```
tian-lt, line 46: I suppose we should bring back UI tests to this release pipeline.
The UI tests were removed because window-2019 is not compatible with the min version(*10.0.22000.0*) specified in the [release manifest](https://github.com/microsoft/calculator/blob/aba642351360e980bfe5fa6ee731c82055378036/src/Calculator/Package.Release.appxmanifest#L11).

tian-lt (changes requested):
Looks good to me.
It'll be great if we can bring back UI tests to release pipeline.

build/pipelines/azure-pipelines.release.yaml
mcooley: windows-2022 = Windows Server 2022 = 10.0.20348.0, which is still not compatible with 10.0.22000.0.

We'll need to stand up a pool of real Windows 11 desktop images to get the UI tests back.

tian-lt (approved):

PR #1799, opened 2022-02-16
Fixes a bug: currency converter never shows error message for service failures

tian-lt:
- Before the fix,  CurrencyDataLoader always discard and register a new *NetworkBehavior* even if it doesn't really take action to load the data. This results in many redundant callback invocation which force the state gets reset so that the VM reset the error message back to normal state immediately.
- After the fix, the *NetworkBehavior* will only be registered when the loader decides to take real actions to load data.

### Screenshots
Before the fix:
image

After the fix:
image

- Tested manually

### DIFF
--- a/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
+++ b/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
@@ -188,10 +188,9 @@ void CurrencyDataLoader::ResetLoadStatus()
 #pragma optimize("", off) // Turn off optimizations to work around DevDiv 393321
 void CurrencyDataLoader::LoadData()
 {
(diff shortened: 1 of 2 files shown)


hanzhang54:
The "CurrencySecondaryStatus" TextBlock is not rejuvenated. Could you please remove the "SemiBold" FontWeight and update the Margin to "3,6,0,0" to make it look better with others?

hanzhang54 (approved):

hanzhang54:
The CI pipeline failed on unit test. Could you please take a look?

tian-lt:
> The CI pipeline failed on unit test. Could you please take a look?

No worries, some flaky tests occurred during UI tests.
Rerun the checks and all passed.

PR #1812, opened 2022-03-24
Fixed Graph layout direction in Right To Left languages

### LINKED ISSUE #1352: Some UI elements flipped in graphing mode in RTL
**Describe the bug**
Some UI elements flipped in graphing mode in RTL

**Expected behavior**
All text and UI elements respect RTL languages.

**Screenshots**
image

**Additional context**
Reported on [Reddit](https://www.reddit.com/r/Windows10/comments/ib0i1i/windows_flips_some_elements_graphics_and_texts/)

**Requested Assignment**

Avid29:
## Fixes #1352.

This change is created because the graph renders backwards when the app FlowDirection is RightToLeft.

This change should override the FlowDirection for grapher, fixing that bug

- Added FlowDirection override to the grapher control.

- I could not properly test the result because the grapher control is closed source

### DIFF
--- a/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
+++ b/src/Calculator/Views/GraphingCalculator/GraphingCalculator.xaml
@@ -592,6 +592,7 @@
                                   LosingFocus="GraphingControl_LosingFocus"
                                   LostFocus="GraphingControl_LostFocus"
                                   UseSystemFocusVisuals="True"
+                                  FlowDirection="LeftToRight"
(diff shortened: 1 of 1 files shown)


hanzhang54:
Thanks for coming up with this fix! There were some discussions about the RTL design in Graphing mode and it is unclear about the expected design from the native RTL language speakers. Thanks again for bringing it up and we'll make further investigation on this. 😊 I'm going to close this as of now and reopen it once we confirm the expected design.

PR #1819, opened 2022-04-11
Update CalculatorHistory.cpp

Venerablevivek:
I have add some comments in the code. So, that it looks clearly understandable to new viewers.
Many new users who are new to Git needs an understandable code so that they can learn something from it and contribute something. If they understand the code.


mcooley:
Thanks for your interest, but we typically don't accept contributions which just add code comments.

PR #1820, opened 2022-04-12
NavCategory | Partial refactor and perf optimization

tian-lt:
### Why
- [`NavCategory.InitializeCategoryManifest(args.User);`](https://github.com/microsoft/calculator/blob/6a5b93cab675e3b826c6720a314679db0a0894c7/src/Calculator/App.xaml.cs#L88) takes around 100ms to check whether or not the Graphing mode should be enabled. However, this work can be done asynchronously from the UI thread so that we can save the launch time.
- `class NavCategory` was taking care of some states that should not be out of its responsibility, e.g. [`s_categoryManifest`](https://github.com/microsoft/calculator/blob/6a5b93cab675e3b826c6720a314679db0a0894c7/src/CalcViewModel/Common/NavCategory.cpp#L88).
- `using namespace std;` is never recommended.

### Changes
- Re-implement [`bool IsGraphingModeEnabled(User ^ currentUser = nullptr)`](https://github.com/microsoft/calculator/blob/6a5b93cab675e3b826c6720a314679db0a0894c7/src/CalcViewModel/Common/NavCategory.cpp#L64)
- Check the Graphing mode asynchronously.
- `iterator` to `const_iterator`
- Introduce `NavViewCategoriesSource` in `MainPage.xaml.cs` to track the state change that may happen for the Graphing mode.


src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -48,45 +47,57 @@ static constexpr int CURRENCY_ID = 16;
...
-        return false;
+        auto wstr = std::to_wstring(number);
+        return _wcsdup(wstr.c_str());
     }
 
-    if (_isGraphingModeEnabledCached != nullptr)
+    bool IsGraphingModeAvailable()
     {
-        return _isGraphingModeEnabledCached->Value;
+        static bool supportGraph = Windows::Foundation::Metadata::ApiInformation::IsMethodPresent("Windows.UI.Text.RichEditTextDocument", "GetMath");
```
mcooley, line 63: GetMath is available in Windows 10 version 19041, and these days Calculator does not support OSes older than 19041. So I think we can just remove this check altogether. Graphing mode is always "available".

You might want to submit a separate change to remove this check, since I think it might have a measurable performance improvement even without the rest of this refactoring.

src/CalcViewModel/Common/NavCategory.cpp
```diff
@@ -48,45 +47,57 @@ static constexpr int CURRENCY_ID = 16;
...
-    return supportGraph;
-}
+    Platform::Agile<Windows::System::User^> CurrentUser;
+    std::mutex GraphingModeCheckMutex;
 
-Box<bool> ^ _isGraphingModeEnabledCached = nullptr;
-bool IsGraphingModeEnabled(User ^ currentUser = nullptr)
-{
-    if (!IsGraphingModeAvailable())
+    wchar_t* towchar_t(int number)
```
mcooley, line 55: We should remove this function, since it's allocating memory which is never cleaned up. One way to fix this would be to  call `std::to_wstring(number).c_str()` inline at this function's call sites.

tian-lt: Good point. Yes, `towchar_t` can be remove.
I'd prefer changing the type of `NavCategoryInitializer::accessKey` from `const wchar_t*` into `std::optional<std::wstring>` so that we can manage the memory and keep its semantics unchanged as well.

src/CalcViewModel/Common/NavCategory.cpp
tian-lt: Nice suggestion.
I removed the check and adjusted the parts that depend on this precondition.

guominrui (dismissed):

src/CalcViewModel/Common/NavCategory.h
```diff
@@ -112,91 +87,98 @@ namespace CalculatorApp::ViewModel
...
 
-        private:
-            NavCategoryGroup(const NavCategoryGroupInitializer& groupInitializer);
+            // Query properties from states
+            static ViewMode GetViewModeForFriendlyName(Platform::String ^ name);
+            static Platform::String ^ GetFriendlyName(ViewMode mode);
+            static Platform::String ^ GetNameResourceKey(ViewMode mode);
+            static CategoryGroupType GetGroupType(ViewMode mode);
+
+                // GetIndex is 0-based, GetPosition is 1-based
```
hanzhang54, line 173: Please check the indentation of this comment.

hanzhang54 (dismissed):

src/CalcViewModel/Common/NavCategory.h
tian-lt: Fixe. Thanks.

hanzhang54 (approved):

PR #1821, opened 2022-04-13
Fix blurry buttons

### LINKED ISSUE #1713: Calculator buttons are blurry
You can see that the buttons in the flyout have sharp borders, while the other ones are blurry. This also happened in the past, when the buttons used reveal highlight.
image

```
 - OS Build: 10.0.22000.0
 - Architecture: X64
 - Application Version: 11.2109.8.0
 - Region: en-US
```

dansmor7:
Fixes #1713.

UseLayoutRounding was disabled in a lot of places, which made a lot of buttons blurry.


hanzhang54 (approved):
Thanks for fixing this issue! It looks good to me. 😊

PR #1845, opened 2022-06-03
Update Nuget Packages

AtariDreams:
Update dependencies to the latest stable versions from Microsoft


hanzhang54:
Hi @AtariDreams , thanks for coming up with this update! Currently we don't have plan to update these packages. We'll update the packages later and close this PR as of now. 😊

hanzhang54:
Hi @AtariDreams , thanks for coming up with this update! Currently we don't have plan to update these packages. We'll update the packages later and close this PR as of now. 😊

Issue #1865, opened 2022-07-21
AoT_ErrorMessage_CannotDivideByZero does not work in pipeline with 1024x768 screen resolution


mcooley:
The AoT_ErrorMessage_CannotDivideByZero test does not work when the screen resolution is 1024x768. This occurs because an always-on-top window's maximum height is half of the screen height, and half of a 768-pixel screen is not tall enough to allow the "invert" button to appear.

In (TODO edit here with PR number), I've disabled this test because our current Win11 test machines support only a 1024x768 resolution.

To fix this, we should add a step to the UI test pipeline to set the screen resolution to a higher value before running the tests. We used to use [Set-DisplayResolution](https://docs.microsoft.com/en-us/powershell/module/servercore/set-displayresolution?view=windowsserver2022-ps) for this, but that's available only on Server Core. A better option might be to use the Screen Resolution Utility task, and I have requested that this task be added to the dev.azure.com/ms organization, but that request has not yet been approved.

Once we've re-established a way to increase the screen resolution, we should re-enable the test.

mcooley:
The Screen Resolution Utility task was added in the dev.azure.com/ms organization, and I've added it in the original PR #1866. Closing this issue.

PR #1866, opened 2022-07-21
Move UI tests to Windows 11

mcooley:
Calculator's UI tests currently run on Server 2022 VM images. However, the production app is primarily developed for Windows 11. This change moves the UI test jobs to run on Windows 11, so we can run our tests in an environment that's closer to what real users will experience.

- Switched UI tests to target the EssentialExperiences-Win11 and EssentialExperiencesOpenSource-Win11 pools
- Add a step to disable animation effects on the machines. Windows Server has this setting off by default, but Windows 11 has it on by default. Our tests currently don't pass when this is on.
- Target 22000 as the minimum Windows version in all configurations, instead of using a lower minimum version for CI builds.
- Run UI tests in the release pipeline again.
- Retarget the UI test projects to .NET 6
- Rewrite the AoT_ErrorMessage_MessageRetentionUponExitingAoT test so it no longer relies on the invert button to be visible
- <s>Disable the AoT_ErrorMessage_CannotDivideByZero test until we have a way to increase the screen resolution on the machines. We can't use the existing PowerShell cmdlet since it isn't present on Windows desktop editions, and an alternative solution is not ready yet. Re-enabling this test is tracked with #1865</s> In the latest iteration, the ScreenResolutionUtility task is added to increase the resolution, for parity with the existing behavior.
- Other minor cleanup in the UI test code, including deleting dead code and using strongly-typed booleans instead of strings.

- Pipeline run.


build/scripts/TurnOffAnimationEffects.ps1
```diff
@@ -0,0 +1,22 @@
...
+    $asTask = $asTaskGeneric.MakeGenericMethod($ResultType)
+    $task = $asTask.Invoke($null, @($WinRtTask))
+    $task.GetAwaiter().GetResult()
+}
+
+[Windows.UI.ViewManagement.Core.UISettingsController,Windows.UI.ViewManagement.Core,ContentType=WindowsRuntime] | Out-Null
+
+$controller = WaitForAsyncAction ([Windows.UI.ViewManagement.Core.UISettingsController]::RequestDefaultAsync())  ([Windows.UI.ViewManagement.Core.UISettingsController])
+
+$controller.SetAnimationsEnabled($false)
```
tian-lt, line 22: nit: an empty line here should be better.

shweaver-MSFT:
Changes look good as far as I can tell. But I saw a few errors when I ran the suite of UI tests locally. This is my first time running the project, so I admit I'm not sure if that is expected or not. 

What do you think?

image

mcooley:
@shweaver-MSFT if the currency converter tests are failing, selecting [the right runsettings file](https://github.com/microsoft/calculator/blob/main/src/CalculatorUITests/CalculatorUITests.ci.runsettings) might help. AoT_Tooltip I'm not sure without seeing more of the failure. Note that these tests are passing in the official pipeline.

shweaver-MSFT:
> @shweaver-MSFT if the currency converter tests are failing, selecting [the right runsettings file](https://github.com/microsoft/calculator/blob/main/src/CalculatorUITests/CalculatorUITests.ci.runsettings) might help. AoT_Tooltip I'm not sure without seeing more of the failure. Note that these tests are passing in the official pipeline.

Okay, so any failures I'm seeing are just a side-effect of my personal environment? Gotcha :)

shweaver-MSFT (approved):

tian-lt (approved):

PR #1884, opened 2022-08-28
Update Packages and Solution Files

AtariDreams:
My system has been unable to build with lots of package related errors since the update to Nugent 6.x.

I found that updating the packages seemed to resolve that issue. There is no gain to be had by keeping the packages out of date.


mcooley:
You're right, it's nice to keep these packages up to date. Unfortunately, we can't merge this PR as-is because it includes too many unrelated changes.

I submitted a smaller PR with just the NuGet package updates here: #1891. If you'd like to submit other parts of this change, please break them up into smaller PRs with good descriptions.

PR #1889, opened 2022-08-31
Fix C# warnings

AtariDreams:
A continuation of my last PR. No behavioral changes, just C# modernizations as well as marking some properties to be readonly since they aren't modified.


mcooley:
I like some of these changes (like the dead code removal in App.xaml.cs), but many of the other changes don't clearly improve the code. If you'd like us to consider these changes, please submit them in smaller, focused PRs with clear descriptions.

PR #1880, opened 2022-08-15
Update release package manifest

hanzhang54:
### Description of the changes:
- Update release package manifest.

### How changes were validated:

- Manually tested.

Issue #1878, opened 2022-08-09
I have installed Windows Universal Runtime version 10.0.22000.0, but I am prompted to upgrade my system


xiaodongYao:
DEP3321: To deploy this application, your deployment target should be running Windows Universal Runtime version 10.0.22000.0 or later. You are currently running version 10.0.19043.1826. Update your OS or change the deployment target to a device with the appropriate version.

denyaalt:
I have the same problem.  
Please post the compiled exe file.

xiaodongYao:
> 

OK!,Thank you!!!!!

xiaodongYao:
> I have the same problem. Please post the compiled exe file.

ok,thank you !

clhenaqeeb:
Same issue here:
DEP3321: To deploy this application, your deployment target should be running Windows Universal Runtime version 10.0.22000.0 or higher. You currently are running version 10.0.19044.1889. Please update your OS, or change your deployment target to a device with the appropriate version.	Calculator

clhenaqeeb:
Also the built app does not launch on win 10 and logs these in the event viewer:

Application: CalculatorApp.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: exception code e0434352, exception address 00007FFAA0754FD9
Stack:

**Followed by:**

Faulting application name: CalculatorApp.exe, version: 0.0.0.0, time stamp: 0x630fff6b
Faulting module name: KERNELBASE.dll, version: 10.0.19041.1889, time stamp: 0xe9ede6d6
Exception code: 0xe0434352
Fault offset: 0x0000000000034fd9
Faulting process id: 0x2218
Faulting application start time: 0x01d8bdb164444cad
Faulting application path: C:\Users\clhenaqeeb\source\repos\calculator\src\x64\Debug\Calculator\CalculatorApp.exe
Faulting module path: C:\windows\System32\KERNELBASE.dll
Report Id: 95660539-6ba6-40b5-bdfd-b6841dcf46eb
Faulting package full name: 
Faulting package-relative application ID: 

**Followed by: (Error reporting)**

Fault bucket 1772086560930812384, type 4
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: CalculatorApp.exe
P2: 0.0.0.0
P3: 630fff6b
P4: KERNELBASE.dll
P5: 10.0.19041.1889
P6: e9ede6d6
P7: e0434352
[...]

PR #1893, opened 2022-09-22
README: require Windows 11 to build and run

mcooley:
Update the README to say that Windows 11 is required to build and run the project.

Building on older versions of Windows 10 _might_ work, but the app won't run because Calculator's minimum OS version is currently 22000. Even if we lower the minimum OS version to support some versions of Windows 10 (see #1880), we'd prefer to guide people towards using Windows 11 for development since that's what we use and are able to support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Calculator ships regularly with new features and bug fixes. You can get the late
 
 ## Getting started
 Prerequisites:
-- Your computer must be running Windows 10, version 1809 or newer. Windows 11 is recommended.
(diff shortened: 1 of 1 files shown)


tian-lt (approved):

xahon:
What do you mean under "Windows 10 _might work_"? I use windows 10 which is not out of support yet and want to build the calculator on it (I can't upgrade to 11 due to hardware limitations), what to do then? How does the Microsoft builds newer versions of the project for the upgrades? If you reopen #1878, we can move a discussion there

PR #1902, opened 2022-09-23
Use a markup extension instead of x:Uid to load localized strings

mcooley:
This PR uses XAML markup extensions, `ResourceString` and `ResourceVirtualKey`, to load localized resources instead of using `x:Uid`. I used a script to find all strings corresponding to an x:Uid and substitute them into the XAML, and then manually cleaned up the generated code.

This is intended to improve developer productivity. Now, when reading the XAML file, you don't have to cross-reference the resw file to see if a control has a tooltip or an automation name.

Decoupling the UI from the string names also makes it easier to change the UI without impacting localization: for example, you can now use the same string in multiple places, or use an existing string on a new UI element. For now, I've avoided making any changes to the resw file to avoid impacting the localization process.


mcooley:
@guominrui, @hanzhang54, @tian-lt please review, and let me know whether you'd like to accept this change. I've just resolved merge conflicts with main. Thanks!

hanzhang54 (approved):

src/Calculator/App.xaml
```diff
@@ -803,14 +803,14 @@
                                                  Foreground="{ThemeResource SystemControlForegroundAccentBrush}"
                                                  BorderThickness="0"
                                                  AutomationProperties.AutomationId="CalculationResultScrollLeft"
+                                                 AutomationProperties.Name="{utils:ResourceString Name=CalculationResultScrollLeft/[using:Windows.UI.Xaml.Automation]AutomationProperties/Name}"
```
guominrui, line 806: Nit: "[using:Windows.UI.Xaml.Automation]" is not needed anymore, maybe we can directly replace "[using:Windows.UI.Xaml.Automation]AutomationProperties" with "AutomationProperties", to shorten the resource key.

guominrui (approved):

src/Calculator/App.xaml
mcooley: Yes, I think it would be good to clean up some of these names. But since that will impact the localization process, I think it would be better to make those changes in a separate PR.

PR #1905, opened 2022-09-24
Make the Ping during testing thread-safe

AtariDreams:
So we do not risk blocking the main UI thread

### DIFF
--- a/src/CalculatorUITestFramework/WindowsDriverLocalService.cs
+++ b/src/CalculatorUITestFramework/WindowsDriverLocalService.cs
@@ -20,6 +20,7 @@ using System.IO;
 using System.Net;
 using System.Net.Http;
 using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
 
 namespace CalculatorUITestFramework
 {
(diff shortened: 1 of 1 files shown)


src/CalculatorUITestFramework/WindowsDriverLocalService.cs
```diff
@@ -146,20 +147,21 @@ private bool Ping()
...
+                if (service.IsLoopback)
+                {
+                    status = new Uri("http://localhost:" + Convert.ToString(this.Port) + "/status");
+                }
+                else
+                {
+                    status = new Uri(service + "/status");
+                }
+
+                return Task.Run(() => httpClient.GetAsync(status)).Result.IsSuccessStatusCode;
```
tian-lt, line 163: consider `var httpResponse = Task.Run(() => httpClient.GetAsync(status)).ConfigureAwait(false).GetAwaiter().GetResult();` to avoid self-deadlock?

tian-lt (approved):
LGTM.
Appreciate your contribution.

PR #1907, opened 2022-09-26
Drop unused enum

michaljx-intel:
Drops unused enum


michaljx-intel:
My bad, it is actually in use, but not by type name.

PR #1909, opened 2022-09-26
Remove partial keyword from Globals class

AtariDreams:
Globals does not need to be a partial class at all.

### DIFF
--- a/src/Calculator/App.xaml.cs
+++ b/src/Calculator/App.xaml.cs
@@ -30,7 +30,7 @@ namespace CalculatorApp
 {
     namespace ApplicationResourceKeys
     {
-        public static partial class Globals
+        public static class Globals
         {
             public static readonly string AppMinWindowHeight = "AppMinWindowHeight";
             public static readonly string AppMinWindowWidth = "AppMinWindowWidth";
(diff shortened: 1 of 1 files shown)


src/Calculator/App.xaml.cs
```diff
@@ -423,9 +423,9 @@ private async Task SetupJumpList()
                         continue;
                     }
                     ViewMode mode = option.ViewMode;
-                    var item = JumpListItem.CreateWithArguments(((int)mode).ToString(), "ms-resource:///Resources/" + NavCategoryStates.GetNameResourceKey(mode));
+                    var item = JumpListItem.CreateWithArguments(mode.ToString(), "ms-resource:///Resources/" + NavCategoryStates.GetNameResourceKey(mode));
```
mcooley, line 426: Enum.ToString() has different behavior than int.ToString(). Is the new behavior correct?

mcooley (changes requested):

src/Calculator/App.xaml.cs
AtariDreams: By bad. I fixed this.

mcooley (approved):

PR #1910, opened 2022-09-26
Remove unneeded closing check

AtariDreams:
Turns out the check has not been needed since 2016.

### DIFF
--- a/src/Calculator/Views/Calculator.xaml
+++ b/src/Calculator/Views/Calculator.xaml
@@ -690,6 +690,7 @@
                     <Flyout x:Name="HistoryFlyout"
                             AutomationProperties.AutomationId="HistoryFlyout"
                             Closed="HistoryFlyout_Closed"
+                            Closing="HistoryFlyout_Closing"
                             FlyoutPresenterStyle="{StaticResource HistoryFlyoutStyle}"
                             Opened="HistoryFlyout_Opened"
                             Placement="Full"/>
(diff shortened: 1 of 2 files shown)


mcooley (changes requested):
Thanks for submitting this. This check is no longer necessary because FlyoutBase.Closing is available on all versions of Windows that Calculator supports. Instead of this cleanup change, we should remove this check, and then register for the Closing events directly in the XAML instead of in code-behind.

(Microsoft internal: here's [when this check was added in 2016](https://microsoft.visualstudio.com/Apps/_git/calculator.app.archive2/commit/4b1545fea8d458ac81ddbfb447f7bec287557404))

AtariDreams:
Fixed @mcooley

mcooley:
@AtariDreams can we register for the events in XAML instead of in the .cs file? I think it would be better to register for the `Closing` events in the same place we register for the `Closed` events: [here](https://github.com/microsoft/calculator/blob/edf08514a6773443bc75b8c3add497e1ab454ebe/src/Calculator/Views/Calculator.xaml#L692) and [here](https://github.com/microsoft/calculator/blob/edf08514a6773443bc75b8c3add497e1ab454ebe/src/Calculator/Views/Calculator.xaml#L790).

AtariDreams:
@mcooley Done!

mcooley (approved):
Thanks!

PR #1912, opened 2022-09-26
Simplify ChangeConstants

AtariDreams:
Since BASEX is a constant and radix is a uint32_t radix, there is nothing to be gained from having the limit be 64 bit.

In addition, the g_ratio += !g_ratio only applies if g_ratio is 0, so I simplified that too.

### DIFF
--- a/src/CalcManager/Ratpack/support.cpp
+++ b/src/CalcManager/Ratpack/support.cpp
@@ -25,7 +25,7 @@
 
 using namespace std;
 
-void _readconstants(void);
+void _readconstants();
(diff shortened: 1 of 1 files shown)


tian-lt (dismissed):

mcooley:
Another way to simplify this would be to use a closed-form expression: `g_ratio = static_cast<int32_t>(ceil(log2(BASEX) / log2(radix))) - 1;`. But I think this change is OK as-is.

mcooley (approved):

PR #1914, opened 2022-09-26
Simplify OnAppPropertyChanged

AtariDreams:
Use a switch statement instead of else-if

### DIFF
--- a/src/Calculator/Views/MainPage.xaml.cs
+++ b/src/Calculator/Views/MainPage.xaml.cs
@@ -206,58 +206,67 @@ namespace CalculatorApp
 
                 KeyboardShortcutManager.DisableShortcuts(false);
 
-                if (newValue == ViewMode.Standard)
+                switch (newValue)
                 {
-                    EnsureCalculator();
-                    Model.CalculatorViewModel.HistoryVM.AreHistoryShortcutsEnabled = true;
-                    m_calculator.AnimateCalculator(NavCategory.IsConverterViewMode(previousMode));
-                    Model.CalculatorViewModel.HistoryVM.ReloadHistory(newValue);
-                }
-                else if (newValue == ViewMode.Scientific)
-                {
-                    EnsureCalculator();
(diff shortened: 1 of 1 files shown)


src/Calculator/Views/MainPage.xaml.cs
```diff
@@ -206,58 +206,67 @@ private void OnAppPropertyChanged(object sender, PropertyChangedEventArgs e)
...
-
-                    EnsureConverter();
-                    if (!NavCategory.IsConverterViewMode(previousMode))
-                    {
-                        m_converter.AnimateConverter();
-                    }
+                        Model.CalculatorViewModel.HistoryVM.ReloadHistory(newValue);
+                        break;
+                    case ViewMode.Scientific:
+                        {
```
tian-lt, line 218: Some brackets can probably be omitted.

tian-lt (approved):

src/Calculator/Views/MainPage.xaml.cs
```diff
@@ -206,58 +206,58 @@ private void OnAppPropertyChanged(object sender, PropertyChangedEventArgs e)
...
+                            m_calculator.AnimateCalculator(NavCategory.IsConverterViewMode(previousMode));
+                        }
+                        break;
+                    case ViewMode.Graphing:
+                        EnsureGraphingCalculator();
+                        KeyboardShortcutManager.DisableShortcuts(true);
+                        break;
+                    default:
+                        if (NavCategory.IsDateCalculatorViewMode(newValue))
```
mcooley, line 239: Looks like `IsDateCalculatorViewMode` [just checks for ViewMode.Date](https://github.com/microsoft/calculator/blob/91adfd8e9e14b9719e51f4c5e2e06380b6f429cc/src/CalcViewModel/Common/NavCategory.cpp#L262-L265), and I don't think that is likely to change. In a future change, we could simplify this and put `ViewMode.Date` in a separate `case`.

mcooley (approved):

PR #1919, opened 2022-09-30
Remove redundant breaks

AtariDreams:
They do not need to be there since return works here.


src/CalcViewModel/Common/ExpressionCommandDeserializer.cpp
```diff
@@ -50,7 +42,7 @@ COpndCommand CommandDeserializer::DeserializeOperand()
     std::shared_ptr<std::vector<int>> cmdVector = std::make_shared<std::vector<int>>();
     auto cmdVectorSize = m_dataReader->ReadUInt32();
 
-    for (unsigned int j = 0; j < cmdVectorSize; ++j)
+    for (auto j = cmdVectorSize; j > 0; --j)
```
mcooley, line 45: This is unrelated to the rest of the PR. I don't there's a good reason to change this.

mcooley (changes requested):

mcooley (approved):

PR #1920, opened 2022-09-30
Use array instead of substring

AtariDreams:


mcooley:
Adopting a different syntax here doesn't make a meaningful improvement to the code. I don't think we should merge this one.

(The string.Format to string interpolation change is OK.)

PR #1923, opened 2022-09-30
Digit should be Mantype

AtariDreams:
It just makes more sense that way


AtariDreams:
@mcooley Any updates?

Issue #1894, opened 2022-09-22
Migrate ViewModel layer to C#


mcooley:
In 2021, the view layer of the Calculator codebase was ported to C#. I propose that we continue this work and migrate the ViewModel layer to C#.

General reasons for migrating to C# are discussed in #893. I expect that we would see a few benefits specifically from migrating the ViewModel layer:
- Remove most of the C++/CX code in this repository, making it easier to convert remaining C++ code to C++/WinRT or another supported solution.
- Make it easier to rewrite some components in this layer which have major issues, like unit converters (#379) and pasted expression parsing (#344).
- In some cases, possibly improve performance by 1) making the interface between C# and C++ code less chatty, and 2) exposing more code to .NET Native's whole-program optimization.

This work is best done incrementally. Here is how I would break up the work:
- [ ] Set up a CalculatorApp.ViewModel project to host viewmodel-layer C# code
- [ ] Port ApplicationViewModel and NavCategory, and their tests
- [ ] Port CopyPasteManager, and its tests
- [ ] Port UnitConverterViewModel, data loaders, and CalcManager's UnitConverter, and their tests
- [ ] Port DateCalculatorViewModel, and its tests
- [ ] Port GraphingCalculatorViewModel, and its tests
- [ ] Port miscellaneous utilities
  - [ ] NarratorNotifier
  - [ ] Utilities
  - [ ] NetworkManager

We should avoid modifying the infinite-precision calc engine as part of this work. StandardCalculatorViewModel, and related classes which call into the engine, should remain in C++ for now.

We should port the code as-is, and postpone major refactoring until after the code has been ported.

We should measure the app startup latency after these changes. If these changes cause startup to get slower, we should revert the changes or optimize code to eliminate the regression.

I would like to work on this in my free time, but I'd welcome help from other contributors.

mcooley:
@tian-lt, @hanzhang54, or @guominrui I'd love your opinion on this proposal :smile:

tian-lt:
Hi @mcooley, the plan looks great.
Here are some questions and comments from myself:

[...]

PR #1924, opened 2022-09-30
Convert ApplicationViewModel to C#

mcooley:
Contributes to #1894

- Convert ApplicationViewModel to C#
- Move a handful of methods on the other viewmodels from `internal` to `public` so ApplicationViewModel can call them
- Small simplification: make the "Categories" property a normal vector instead of an observable vector, since in practice it never changes


Neracles (approved):

mcooley:
Closing out this PR. I hope that we get this done someday, but I won't have time to work on it anytime soon. Others should feel free to take this code and move it forward.

PR #1934, opened 2022-10-13
Update Calculator app to run on Windows 10 21H2 and up

AtariDreams:
Also update the SDK to Windows 11 build 22621

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Calculator ships regularly with new features and bug fixes. You can get the late
 
 ## Getting started
 Prerequisites:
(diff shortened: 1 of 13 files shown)


LexonBlackzz (approved):
seems good

mcooley:
I don't think we can accept a PR to change the minimum supported version. Changing which OSes we support is more complicated than just ensuring that the code works; we also would need to change some internal testing and release processes. Not saying we won't do this...but if we do it, it will be in #1880.

I'd be happy to merge a PR which updates the SDK targeting _without_ changing TargetDeviceFamily.

PR #1946, opened 2022-11-04
Do some codebase modernization

AtariDreams:


mcooley:
Some of these changes seem like they'd be nice to have, while others change the code style for minimal benefit. Since it has so many changes mixed together, I don't think we can accept this change.

PR #1961, opened 2022-12-29
Fixed hyperbolioc trignometric functions not working for degrees.

### LINKED ISSUE #1931: Scientific mode - Sinh gives wildly incorrect result.
The Scientific calculator provides the ability to enter angles in degrees, radians, or grads to calculate the sine, cosine, tangent, etc. When using the inverse trig functions the result is converted to the currently selected units. This means that in DEG mode I can enter 30 and press the sin button and get the correct result of 0.5 and when I select the arcsine operation I get 30 back as the result. However, with the hyperbolic sinh function does not do the angle conversion from degrees to radians and provides a nonsensical answer. It requires that the angle be entered in units of radians regardless of the DEG/RAD/GRAD mode selected. This seems to be a bug in the calculator. I have not yet tested the other hyperbolic functions, but I would expect to encounter the same issue.

The calculator is not honoring the DEG angle units for the hyperbolic sine function, and perhaps others.

**Steps To Reproduce**

1. Go to 'Scientific. Calculator mode.
2. Ensure DEG/RAD/GRAD button is showing 'DEG.'
3. Enter '30'
4. Expand the 'Trigonometry' popup window
5. Select 'hyp' to display the hyperbolic functions.
6. Click the 'sinh'.
-->

**Expected behavior**
I expected to see the hyperbolic sine of 30 degrees calculated,  Instead I apparently get the sinh of 30 radians.

It appears that calculator feeds the entered value 30 into the sinh calculation resulting in an incorrect value.
Instead of the correct value of 0.547853... I get somewhere around 5,343,237,290,762.290762...

The problem appears to also affect the inverse hyperbolic sine function.

 - OS Build: Windows 10 Pro 21H2 Build 19044.2006
 - Major  Minor  Build   Revision
 -  -----  -----    -----   --------
    10      0        19044   0
 - Architecture: x64
 - Application Version:  10.2103.8.0
 - Region: US English (en-US)

**Additional context**

I do not know if the same issue affects other hyperbolic trig functions

**Requested Assignment**

avighnac:
## Fixes #1931 
Fixed hyperbolic trignometric functions not working for degrees, thus fixing issue https://github.com/microsoft/calculator/issues/1931.

- Added hyperbolic sine, cosine, and tangent for angles.
- Moved `scalerat()` and `ascalerat()` to `support.cpp`

- Manually tested sinh, cosh, and tanh for both radians and degrees.


avighnac:

avighnac:
For some reason, the test cases that check the changes I've made check against the radian equivalents instead of the degree equivalents of the hyperbolic functions.

avighnac:
Hello?

PR #2028, opened 2023-07-12
Adds Engineering Notation

### LINKED ISSUE #226: Automatic scientific notation can't be disabled
**Describe the bug**
In scientific mode scientific notation can be enabled by F-E button but can't be disabled if number have more than 2 zeros after decimal point. This automatic scientific notation is also enabled even on standard mode. Sometimes it's more useful and more readable if number stay in common decimal format instead of scientific notation. For example while calculating crypto currencies it is more useful to see 0.000005142857 where I can easily see 514 satoshi price than 5.142857e-6.

**Steps To Reproduce**
1. Switch to standard or scientific mode
2. Type 0.000036
3. Press divide symbol
4. Type 7
5. Press Enter
6. Pressing F-E is not changing output format

**Expected behavior**
- Stay in normal decimal format until I press F-E to switch to scientific notation
OR
- Show scientific notation but also enable F-E switch to indicate scientific notation mode with ability to disable F-E to show number in common decimal format
OR
- Add another switch to enable/disable automatic scientific notation if number conains more than defined number of zeros after decimal point
OR
- At disable automatic scientific notation in standard mode calculator at all and add another switch to enable/disable automatic scientific notation in scientific mode if numbers contains more than defined number of zeros after decimal point

 - OS Build: 10.0.17763.0
 - Architecture: X64

### LINKED ISSUE #1304: Option to make exponent divisible by 3 (for metric units reading)
**Problem Statement**
When working with metric units it's usefull to see scientific notation with exponent divisible by 3. Number 3,3898e-4 is not as clear as 338.98e-6 is clearly 460µ units.

**Evidence or User Insights**
You can try "1/2950=" in standard or scientific mode to see result. It was discussed in #226 also.

**Proposal**
For calculations in metric units it's better to see results in format similar to T, G, M, k, m, µ, n, etc. units.

**Goals**
User should be able to set exponent divider to make scientific notation easily readable. Default value should by 1 to preserve actual state with ability to enable predefined value 3 (or even other value) to read metric results more easily.

**Non-Goals**
It's not meant to make any kind of units conversion.

**Low-Fidelity Concept**
If exp>0 then exp=int(exp/div)*div
if exp<0 then exp=int((exp-div+1)/div)*div

**Requested Assignment**

Goofyeet:
## Fixes #1304

https://github.com/microsoft/calculator/assets/76914681/9cd69755-b0ef-4b25-930a-375d05eeb3b6

- Replaces FE button with AUTO, SCI, and ENG buttons similar to DEG, RAD, and GRAD buttons
- A fourth option for OFF can be added in the future to fix #226

- A UI smoketest was added for engineering notation
- Manual testing


Goofyeet:

bflmpsvz:
I am sorry, I am not able to find any basic info about this page. I only know I need calculator with engineering notation. What to do? I am completely lost here. Does it exist? Can I download it or create it myself?

PR #2116, opened 2024-02-01
Fix: Narrator doesn't announce "There's no history yet"

tian-lt:
## Fix: Narrator doesn't announce "There's no history yet"
Repro steps:
1. Launch Calculator app.
1. Navigate Open navigation button and activate it.
1. Navigate to Scientific calculator list item and activate it.
1. Turn on Narrator by 'Ctrl+Win+Enter' keys.
1. Navigate to the "History" tab item and activate it.
1. Observe the issue.

Actual Result:
When no history is present there, narrator does not announce “There’s no history yet”. It announces as 'History tab item 1 of 2, selected'.

Expected Result:
When no history is present there, narrator should announce the information also as 'There’s no history yet'. It should announce as 'History tab item 1 of 2, selected, there's no history yet'.

## How
Due to the technical limitation from XAML UIA infra, there is no canonical way to append Narrator announcement right after the announcement generated for items-container, i.e. after the string _1 of 2, selected_.
To fix this problem, I have to use a work around in the announcement for Pivot Item Headers, which is, instead of saying _History tab item 1 of 2, selected, there's no history yet_, we can make it by saying _There's no history yet. History tab item 1 of 2, selected_.

### DIFF
--- a/src/Calculator/Resources/en-US/Resources.resw
+++ b/src/Calculator/Resources/en-US/Resources.resw
@@ -2778,11 +2778,11 @@
     <comment>{Locked}The Octal button</comment>
   </data>
   <data name="HistoryEmpty.Text" xml:space="preserve">
-    <value>There’s no history yet</value>
+    <value>There’s no history yet.</value>
     <comment>The text that shows as the header for the history list</comment>
   </data>
   <data name="MemoryPaneEmpty.Text" xml:space="preserve">
-    <value>There’s nothing saved in memory</value>
+    <value>There’s nothing saved in memory.</value>
     <comment>The text that shows as the header for the memory list</comment>
   </data>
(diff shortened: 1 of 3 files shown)


hanzhang54:
There is one regression issue about this fix: It is "unable to move narrator cursor" when trying to move the cursor to the item when the History/Memory is not empty. But the narrator could read the History/Memory item before.
image

tian-lt:
> There is one regression issue about this fix: It is "unable to move narrator cursor" when trying to move the cursor to the item when the History/Memory is not empty. But the narrator could read the History/Memory item before. image

Good finding.
Looks like the 11.2311.0.0 version already has such regression. We can create another PR to fix it.

hanzhang54 (approved):

hanzhang54 (approved):

PR #2152, opened 2024-04-03
Fix: heap corruption in the Graphing mode

tian-lt:
## What
We've observed some app crash session happening around the graphing calculator. A high-repro-rate approach is launching Calculator twice to the Graphing mode. Tips: using the gflags tool to inspect pageheap can have almost 100% repro.

## Root cause
Graphing mode may render curves in two different threads. See more at `RenderMain::RunRenderPass()` and `RenderMain::RunRenderPassAsync()`. The former is always executed in the UI thread, while the latter is always going to run on a background thread. Such implementation introduces a complexity of managing inter-thread synchronization, and the current is codebase is not doing good on that.
A minimum fix is do not run render passes on different threads. It helps us get rid of the problem of race condition and data race. It also eliminates the needs of managing the *happens-before* relationship between the store/load operations over the same piece of memory on different threads, which cannot be achieve by simply locking a *critical section*.
It's hard to believe that this minimum could lead to a performance problem on rendering, because:
1. `RunRenderInternal()` is and must be an efficient function for it's the bottleneck of image presenting.
2. Render passes are mostly run synchronously in UI thread, in some cases they may run in a background thread. If we are not seeing any problems of rendering performance, we will not see any of them when doing the same work in UI thread asynchronously.
3. All the graphics are rendered on demand. There is not a stable FPS render running behind the scenes. Hence, if there's any defects in this fix, it won't be amplified.

## Changes
- Always run render passes on the same thread, i.e. the UI thread.
- Use `concurrency::task<bool>` as coroutine to return the result of `RunRenderPassInternal`.
- Remove the *critical_section* - `RenderMain::m_criticalSection`
- Removed some unused fields.
- Fix some wrong use of `IAsyncAction`
- Add `std::` back to *Grapher.cpp*.

### DIFF
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -543,7 +543,7 @@ namespace GraphControl
         if (m_graph != nullptr && m_renderMain != nullptr)
         {
             m_graph->SetArgValue(variableName->Data(), newValue);
-            m_renderMain->RunRenderPass();
+            [](RenderMain ^ renderMain) -> winrt::fire_and_forget { co_await renderMain->RunRenderPassAsync(); }(m_renderMain);
         }
     }
 
--- a/src/GraphControl/Control/Grapher.cpp
+++ b/src/GraphControl/Control/Grapher.cpp
@@ -11,7 +11,6 @@ using namespace GraphControl;
 using namespace GraphControl::DX;
 using namespace Platform;
 using namespace Platform::Collections;
-using namespace std;
 using namespace Concurrency;
 using namespace Windows::Devices::Input;
 using namespace Windows::Foundation;
@@ -54,9 +53,9 @@ namespace
     // posX/posY are the pointer position elements and width,height are the dimensions of the graph surface.
     // The graphing engine interprets x,y position between the range [-1, 1].
     // Translate the pointer position to the [-1, 1] bounds.
-    __inline pair<double, double> PointerPositionToGraphPosition(double posX, double posY, double width, double height)
+    __inline std::pair<double, double> PointerPositionToGraphPosition(double posX, double posY, double width, double height)
     {
-        return make_pair((2 * posX / width - 1), (1 - 2 * posY / height));
+        return std::make_pair((2 * posX / width - 1), (1 - 2 * posY / height));
     }
 }
 
@@ -71,7 +70,7 @@ namespace GraphControl
 
         m_solver->ParsingOptions().SetFormatType(s_defaultFormatType);
         m_solver->FormatOptions().SetFormatType(s_defaultFormatType);
-        m_solver->FormatOptions().SetMathMLPrefix(wstring(L"mml"));
+        m_solver->FormatOptions().SetMathMLPrefix(L"mml");
 
         DefaultStyleKey = StringReference(s_defaultStyleKey);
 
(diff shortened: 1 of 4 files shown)


src/GraphControl/DirectX/RenderMain.cpp
```diff
@@ -210,53 +201,37 @@ namespace GraphControl::DX
...
-
         return RunRenderPassInternal();
     }
 
-    IAsyncAction ^ RenderMain::RunRenderPassAsync(bool allowCancel)
+    concurrency::task<bool> RenderMain::RunRenderPassAsync(bool allowCancel)
     {
-        // Try to cancel the renderPass that is in progress
-        if (m_renderPass != nullptr && m_renderPass->Status == ::AsyncStatus::Started)
+        if (allowCancel)
```
guominrui, line 209: Just according to the original implementation of this method, originally `allowCancel` is for the task being fired. For example, task 1 is fired with `false`, then when firing task 2, no matter `allowCancel` is true or false, the task 1 will not be canceled.
But after this change, task 2 with `true` will cancel the task 1. Not sure if this is as expected.

src/GraphControl/DirectX/RenderMain.cpp
```diff
@@ -47,6 +46,7 @@ namespace GraphControl::DX
 
     RenderMain::~RenderMain()
     {
+        m_renderPassCts.cancel();
```
guominrui, line 49: Maybe it's not enough to protect from using destructed `this` in `RunRenderPassInternal`. For example, 

1. `RunRenderPassAsync();`
2. Then `RunRenderPassAsync(false)`, it will not cancel the first task, but `m_renderPassCts` will be replaced with the one for the new task.
3. Before the two scheduled tasks execute, destroy `this`, the second task will be cancelled, will not call `RunRenderPassInternal`. But the first task can't be cancelled.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -54,9 +53,9 @@ namespace
     // posX/posY are the pointer position elements and width,height are the dimensions of the graph surface.
     // The graphing engine interprets x,y position between the range [-1, 1].
     // Translate the pointer position to the [-1, 1] bounds.
-    __inline pair<double, double> PointerPositionToGraphPosition(double posX, double posY, double width, double height)
+    __inline std::pair<double, double> PointerPositionToGraphPosition(double posX, double posY, double width, double height)
     {
-        return make_pair((2 * posX / width - 1), (1 - 2 * posY / height));
+        return std::make_pair((2 * posX / width - 1), (1 - 2 * posY / height));
```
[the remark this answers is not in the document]
tian-lt: Sounds good. I double checked the c++ version that the project is using, it is targeting C++17, which has CTAD. Updated per the suggestion.

src/GraphControl/Control/Grapher.cpp
```diff
@@ -392,11 +379,11 @@ namespace GraphControl
             {
                 initResult = TryInitializeGraph(keepCurrentView, graphExpression.get());
 
-                if (initResult != nullopt)
+                if (initResult.has_value())
                 {
-                    auto graphedEquations = initResult.value();
+                    auto graphedEquations = *initResult;
```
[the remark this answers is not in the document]
tian-lt: nice catch

src/GraphControl/DirectX/RenderMain.cpp
tian-lt: Good point. Yes, the cancellation point was not doing right. I updated it to the first place where the task gets resumed.
Also, I removed the cancellation token and choose to use an integer to tag a version for each async task. Because it is much cheaper and works well in the same thread.

src/GraphControl/DirectX/RenderMain.cpp
tian-lt: Good point. Updated to use a WeakReference to prevent from use-after-free.

guominrui (approved):

PR #2161, opened 2024-04-23
Multi-instance

tian-lt:
## What
Calculator is a multi-view UWP application, which hosts multiple top-level windows in different threads. This threading model requires shared memory to be carefully protected from unsynchronized concurrent store/load operations.
A multi-instance app does not need to pay the same engineering efforts because the app will be launched in different processes to run independent app sessions.
See more about multi-instance in [doc](https://learn.microsoft.com/en-us/windows/uwp/launch-resume/multi-instance-uwp)
> From Windows 10, version 1803 (10.0; Build 17134) onward, your UWP app can opt in to support multiple instances. If an instance of an multi-instance UWP app is running, and a subsequent activation request comes through, the platform will not activate the existing instance. Instead, it will create a new instance, running in a separate process.

## How
1. Update `Package.appxmanifest` and `Package.Release.appxmanifest` to enable multi-instance mode.
1. Remove `WindowFrameService.cs`.
1. Remove the Frames management works in `App.xaml.cs`.
1. Remove `DisableSystemViewActivationPolicy()`.
1. Load the frame properly when app is activated.
1. Add some diagnostic information.
1. Simplify the graphing mode check.


src/Calculator/App.xaml.cs
```diff
@@ -179,234 +122,45 @@ private void OnAppLaunch(IActivatedEventArgs args, string argument)
...
-                    // We couldn't navigate to the main page, kill the app so we have a good
-                    // stack to debug
-                    throw new SystemException();
-                }
-
-                SetMinWindowSizeAndThemeAndActivate(rootFrame, minWindowSize);
-                m_mainViewId = ApplicationView.GetForCurrentView().Id;
-                AddWindowToMap(WindowFrameService.CreateNewWindowFrameService(rootFrame, false, weak));
+                    FlowDirection = LocalizationService.GetInstance().GetFlowDirection()
+                };
```
hanzhang54, line 131: nit: use `App.CreateFrame()` directly.

src/Calculator/App.xaml.cs
```diff
@@ -179,234 +122,45 @@ private void OnAppLaunch(IActivatedEventArgs args, string argument)
...
+                throw new SystemException("6d430286-eb5d-4f8d-95d2-3d1059552968");
             }
-        }
 
-        private void DismissedEventHandler(SplashScreen sender, object e)
-        {
-            _ = SetupJumpList();
-        }
+            // SetPreferredMinSize should always be called before Window.Activate
+            appView.SetPreferredMinSize(minWindowSize);
```
hanzhang54, line 140: This is also set in line 115. Perhaps we can move it forward.

hanzhang54 (approved):

src/Calculator/App.xaml.cs
tian-lt: `CreateFrame` has been removed by this PR. The work of `CreateFrame` is not complicated and inline code may improve the fluency of reading.

src/Calculator/App.xaml.cs
tian-lt: removed the first one.

hanzhang54 (approved):

guominrui (approved):

PR #2170, opened 2024-05-06
Improve the "about equal to" for Data Converter 

zxxzxxzxhk:

As DVD size is 4700 megabytes. However, current converter's "about equal to" is only showing 0.93 DVDs when 4700 megabytes is entered.
image

Therefore, { ViewMode::Data, UnitConverterUnits::Data_DVD, 5046.586573 } is revised to { ViewMode::Data, UnitConverterUnits::Data_DVD, 4700 }, under UnitConverterDataLoader.cpp
image

Manual testing was done. 1 DVDs is shown when 4700 megabytes was entered.


zxxzxxzxhk:
> Thanks @zxxzxxzxhk for sending this PR. It looks good to me. BTW, the capacity number of CD also seems a bit wired. 700 Megabytes is a common size for CD-R. Could you help take a look and update it as well in this PR? Thanks again.

Done!

tian-lt (approved):

PR #2188, opened 2024-06-11
Fix exception when setting content to clipboard

hanzhang54:
## Fixes #.
35286216

When setting content to clipboard, it may throw exception sometimes. Use SetContentWithOptions() instead of SetContent() to avoid exception. The default options allow the content to be eligible for clipboard history as well as to be synced to other devices, which is the same as SetContent().

Manually tested the copy/paste functionality.

### DIFF
--- a/src/CalcViewModel/Common/CopyPasteManager.cpp
+++ b/src/CalcViewModel/Common/CopyPasteManager.cpp
@@ -67,7 +67,7 @@ void CopyPasteManager::CopyToClipboard(String ^ stringToCopy)
     // Copy the string to the clipboard
     auto dataPackage = ref new DataPackage();
     dataPackage->SetText(stringToCopy);
-    Clipboard::SetContentWithOptions(dataPackage, ref new ClipboardContentOptions());
+    Clipboard::SetContentWithOptions(dataPackage, nullptr);
 }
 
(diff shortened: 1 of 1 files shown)


src/CalcViewModel/Common/CopyPasteManager.cpp
```diff
@@ -67,7 +67,7 @@ void CopyPasteManager::CopyToClipboard(String ^ stringToCopy)
     // Copy the string to the clipboard
     auto dataPackage = ref new DataPackage();
     dataPackage->SetText(stringToCopy);
-    Clipboard::SetContent(dataPackage);
+    Clipboard::SetContentWithOptions(dataPackage, ref new ClipboardContentOptions());
```
tian-lt, line 70: Maybe we can simply give it a `nullptr` to use the built-in default option?

tian-lt (changes requested):

src/CalcViewModel/Common/CopyPasteManager.cpp
hanzhang54: Agree, nullptr will use the default options, which is more straightforward.

tian-lt (approved):

PR #2195, opened 2024-06-20
eng: New CI pipeline with GitHub Action

tian-lt:
## What
The current CI pipeline hosted in Azure DevOps is using a private pool and imaged managed by Microsoft. There are many problems with such situation:
1. The pipeline requires permissions granted by Microsoft employees to run.
1. The pipeline requires internal security tools to run in very jobs, which causes extra time of execution.
1. Some internal tools may need the *write* permission, otherwise it will fail the pipeline run. 

## Fix
Add a [new CI pipeline](https://github.com/microsoft/calculator/actions/workflows/action-ci.yml) built for GitHub action to retire the existing one - [Calculator-CI](https://dev.azure.com/ms/calculator/_build?definitionId=57&_a=summary)
The new pipeline allows us to:
1. run for the open-source community, and
2. get rid of the pool/image hosted in Azure DevOps, and
3. provide a chance for forkers to build their own pipeline, etc.

## Test runs
- [For PR](https://github.com/microsoft/calculator/actions/runs/9601401085)
- [For Push](https://github.com/tian-lt/calculator/actions/runs/9601858480)
- [Original tests](https://github.com/tian-lt/calculator/actions/runs/9601102537)
- [Existing CI](https://dev.azure.com/ms/calculator/_build/results?buildId=585959&view=results)


hanzhang54 (approved):

.github/workflows/action-ci.yml
```diff
@@ -0,0 +1,160 @@
...
+    env:
+      appDir: ${{ github.workspace }}/download/Calculator/AppPackages/Calculator*_Test
+      pubDir: ${{ github.workspace }}/download/publish
+    steps:
+    - uses: actions/download-artifact@v4
+      name: Download build outputs
+      with:
+        name: Build-x64
+        path: ${{ github.workspace }}/download
+    - run: |
```
oneonezhang, line 134: Q: Do we need to turn off animation effects like existing CI?

guominrui (approved):

guominrui (approved):

.github/workflows/action-ci.yml
tian-lt: Good question. The image for ui testing is now Windows Server which is different from the Windows 11 image that ADO is using. I haven't found the evidence to prove that the Animation Effects are off on `windows-latest` but I believe it should be off by default.
Since the UI testing is successful and the time internal has no big difference, I think it's fine to skip this step until we found some problem.

oneonezhang: Yeah, looks good to me😆

PR #2207, opened 2024-07-14
GitHub: Disable blank issue

0x5bfa:
Typically when a author reports a black issue, the managers have to label or change titles. Having them decide and make a proper title could be a good idea for the managers.


0x5bfa:
I personally work on code refactoring and code quality assurance in GitHub therefore, I'd like to propose 'code quality' issue type in New Issue.
What do you think?

PR #2216, opened 2024-08-08
Fix Fluent Design guidelines link to latest version

JuanFKurucz:
I found that the design guidelines link is broken as the fluent design was updated to version two, i changed the link to the new version.


grochocki:
The current link to Windows app-specific design guidelines and principles is correct.

PR #2229, opened 2024-09-09
Add Angstroms Length-Converter option

### LINKED ISSUE #2175: Add angstroms as convertible length unit
**Problem Statement**

The problem is that the Windows calculator does not currently support length conversions to/from angstroms. The target audience is people who want to convert some unit of length to/from angstroms. There is no significant pain in need of remediation, alternate conversion tools are easily accessible.

**Evidence or User Insights**

Lithographies are shrinking, the unit is more relevant in that respect and vertical. It's also reasonable to assume that finer measurement units become more relevant as time passes.

**Proposal**

This feature helps solving the problem stated as it implements the feature the problem states as it's cause.

**Goals**

User can convert to/from angstroms to other relevant length units.

**Low-Fidelity Concept**

image

**Requested Assignment**

DevBoiAgru:
- Adds Angstrom as a unit of length
- Angstrom can be converted to other units and vice versa

- Added unit testing

### DIFF
--- a/src/CalcViewModel/DataLoaders/UnitConverterDataConstants.h
+++ b/src/CalcViewModel/DataLoaders/UnitConverterDataConstants.h
@@ -165,8 +165,9 @@ namespace CalculatorApp
             Data_Zetabytes = UnitStart + 164,
             Area_Pyeong = UnitStart + 165,
             Energy_Kilowatthour = UnitStart + 166,
-			Data_Nibble = UnitStart + 167,
-            UnitEnd = Data_Nibble
+            Data_Nibble = UnitStart + 167,
+            Length_Angstrom = UnitStart + 168,
+            UnitEnd = Length_Angstrom
         };
     }
 }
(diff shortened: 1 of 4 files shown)


DevBoiAgru:

DevBoiAgru:
Hi @tian-lt I hope you're doing well, just wanted to ask if this small MR could be merged? It's a small and straightforward update.

Thanks for your time! 😊

DevBoiAgru:
@tian-lt Thanks for the reply. Is there any way i could run the tests locally on my machine so that I can mention you when it's done?

tian-lt:
Sure. You can run those unit tests by building the unit test project locally, which is located here: https://github.com/microsoft/calculator/tree/main/src/CalculatorUnitTests
And, you may manually add a self-signed certificate to pack the unit test project to install it on your machine and then run for testing. The way to do that is exactly the same as what packing a side-load UWP app needs.
Feel free to let me know if you need helps.

DevBoiAgru:
@tian-lt Apologies for another mention but the most recent commit should fix the test case not passing. I have modified the test case so that it checks for the correct units. Thank you.

tian-lt:
LGTM.
@hanzhang54 do you have any comments on this?

tian-lt (approved):

PR #2245, opened 2024-10-30
Improve the Recall support

tian-lt:


src/CalcViewModel/Snapshots.h
```diff
@@ -0,0 +1,156 @@
...
+    ref struct UnaryCommand sealed : public ICalcManagerIExprCommand
+    {
+        property Windows::Foundation::Collections::IVectorView<int> ^ Commands {
+            Windows::Foundation::Collections::IVectorView<int> ^ get();
+            void set(Windows::Foundation::Collections::IVectorView<int> ^ commands);
+        };
+
+        UnaryCommand();
+
+        internal :;
```
hanzhang54, line 27: What's the meaning of this?

src/CalcViewModel/Snapshots.h
```diff
@@ -0,0 +1,156 @@
...
+        property int Command;
+
+        Parentheses();
+
+        internal :;
+        explicit Parentheses(int cmd);
+    };
+
+public
+    ref struct CalcManagerHistoryToken sealed
```
hanzhang54, line 73: The name is kind of confusing as this token is not only used in history but also in `ExpressionDisplaySnapshot`. Perhaps we can just name it `CalcManagerToken` without history?

src/CalcViewModel/Snapshots.h
tian-lt: `internal` is a special access specifier in C++/CX only. It means its following members are not going to be exported.
Adding an immediate semi-colon `;` here is a workaround to make the code friendly to clang-format since C++/CX is not under its support. Otherwise, after running clang-format on this source file, the lines around specifier `internal` are aweful :(

src/CalcViewModel/Snapshots.h
tian-lt: make sense. updated

src/Calculator/Utils/JsonUtils.cs
```diff
@@ -0,0 +1,330 @@
...
+    internal class ExpressionDisplaySnapshotAlias
+    {
+        [JsonIgnore]
+        public ExpressionDisplaySnapshot Value;
+
+        [JsonPropertyName("t")]
+        public IEnumerable<CalcManagerTokenAlias> Tokens
+        {
+            get => Value.Tokens.Select(x => new CalcManagerTokenAlias(x));
+            set => Value.Tokens = value.Select(Helpers.MapHistoryToken).ToList();
```
hanzhang54, line 210: Remove "History" around "Token" in other places as well?

hanzhang54:
If the snapshot is saved in other modes except for the Standard, relaunch will crash.

tian-lt:
> If the snapshot is saved in other modes except for the Standard, relaunch will crash.

Nice catch. Fixed the unwrapping parts for optional properties.

src/Calculator/Utils/JsonUtils.cs
tian-lt: updated

src/Calculator/Views/MainPage.xaml.cs
```diff
@@ -61,24 +62,37 @@ public MainPage()
...
             {
-                var deferral = args.GetDeferral();
-                if (deferral == null)
+                using (var deferral = args.GetDeferral())
                 {
-                    // Windows Bug in ni_moment won't return the deferral propoerly, see https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/47775705/
-                    return;
+                    if (deferral == null)
+                    {
+                        // Windows Bug in ni_moment won't return the deferral propoerly, see https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/47775705/
```
hanzhang54, line 69: Nit: It seems the bug has been resolved. Maybe we can consider removing this later.

tian-lt: Let's keep it at here, since it was an OS bug. Added a trace to track the actual behavior.

src/Calculator/Views/MainPage.xaml.cs
```diff
@@ -61,24 +62,38 @@ public MainPage()
...
+                        // FIXME: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/47775705/
+                        TraceLogger.GetInstance().LogRecallError("55e29ba5-6097-40ec-8960-458750be3039");
+                        return;
+                    }
+                    var channel = UserActivityChannel.GetDefault();
+                    var activity = await channel.GetOrCreateUserActivityAsync($"{Guid.NewGuid()}");
+                    string embeddedData;
+                    try
+                    {
+                        var json = JsonSerializer.Serialize(new ApplicationSnapshotAlias(Model.Snapshot));
```
hanzhang54, line 78: Perhaps we can use `JsonSerializer.SerializeToUtf8Bytes` to save the transforming work?

hanzhang54 (approved):

src/Calculator/Views/MainPage.xaml.cs
tian-lt: probably next time when we need it. because 1) it's not a bottle neck of performance here. 2)`JsonSerializer` doesn't have a good deserializer accepting utf-8 bytes. 3) it's easy to debug with string rather than an array of bytes.

PR #2249, opened 2024-11-13
Resolve an existing defect when restoring from snapshots

tian-lt:
## Verified test cases:
1. empty state.
1. `number`
1. `number` + `operator`
1. `number` + `operator` + `=`
1. error without expressions
1. error with expressions


guominrui:
Input 0 / 2, press Enter. After restoring, the error message doesn't show.

tian-lt:
> Input 0 / 2, press Enter. After restoring, the error message doesn't show.

Nice catch! Fixed.

guominrui (approved):

PR #2279, opened 2025-01-21
Set warning level to 4 and treat warnings as errors

hanzhang54:
- Set warning level to 4 and turn on Treat Warnings As Errors for the native projects.
- Turn on C4242 explicitly, as it is off by default, but must-fix per Cyber EO.

- Successfully build it without error.


tian-lt (approved):
LGTM. Thanks for paying efforts to remove the warnings which have been a big headache for   many years!

Issue #1993, opened 2023-04-25
Programmer View: Wrong Result: 1.000.000.000 / 64.487 = 0


rensushan:
**Describe the bug**

I am doing some cryptography research and its convinent for me to quickly validate some proofs .

When dividing integers in Programmer view the result of (1<<46) /8380417  is 0 instead of 8396807. 

The results for dividing by 8380416  and 8380418  are correct.

Tested on two different machines running Windows 11.

**Steps To Reproduce**

Steps To Reproduce

Go to Programmer view
Select DEC input
Input  1<<46
Hit  "="
Output  should be  70,368,744,177,664 (Correct)
Hit "/",    Then  input  8380417  (   to divide    70,368,744,177,664 by 8380417  )
Result is 0

**Expected behavior**

Result shoud be 8396807

**Screenshots**

![image](https://user-images.githubusercontent.com/17867516/234164642-25778a62-9ad5-4efd-af18-208ab713e97a.png)

HP ZHAN 66 Pro A, 14" G5 Notebook.

**Device and Application Information**
 - OS Build:  Microsoft Windows 11(Home Edition);
 - Architecture:   AMD Ryzen 7 5825U with Radeon Graphics 2.00 GHz; 
 - Application Version:   Calculator 11.2210.0.0
 - Region:
 - Dev Version Installed:  

**Additional context**

**Requested Assignment**

I'm just reporting this problem.  I don't want to fix it.

rillig:
The wrong value is calculated in `scidisp.cpp:73`, in `result &= GetChopNumber()`.

For some reason, `Rational(8396807.00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000019316) & 0xffff_ffff_ffff_ffff` results in 0.

Thinking further, the many zeroes are already wrong, the rational number should never have `m_mantissa.size() == 16`, it should simply be `Rational(1 << 46, 8380417)`.

According to Perl and Python, `1 << 46 / 8380417` is 8396807.00586427 or 8396807.005864266.

Please don't expect further help or input from me, I just did this as an exercise to try out the Visual Studio debugger.

rillig:
Minimal reproducer:
~~~c++
auto nom = CalcEngine::Rational(1ULL << 46);
auto den = CalcEngine::Rational(8380417);
auto quot = nom / den;

auto mask = CalcEngine::Rational(0ULL - 1);
auto result = quot & mask;
~~~

Issue #1788, opened 2022-01-31
Programmer View: Wrong Result: 1.000.000.000 / 64.487 = 0


bahlea:
**Describe the bug**

When dividing integers in Programmer view the result of 1.000.000.000 / 64.487 is 0 instead of 15.507. It works with the hex equivalent of 0x3B9ACA00 / 0xFBE7 = 0x3C93. Changing the word size (QWORD/DWORD) does not affect this.
The results for dividing by 64.486 and 64.488 are correct.
Tested on two different machines running Windows 10.

**Steps To Reproduce**

1. Go to Programmer view
2. Select DEC input
3. Input 1.000.000.000
4. Hit / 
5. Input 64.487
6. Hit enter
7. Result is 0

**Expected behavior**

Result shoud be 15.507

**Screenshots**

Wrong result for DEC 64.487:
![grafik](https://user-images.githubusercontent.com/12861218/151819477-f0959e5a-2e96-4a10-bc58-c22eaf3f7583.png)

Right result for HEX 0xFBE7:
![grafik](https://user-images.githubusercontent.com/12861218/151820074-b7a83f9c-b417-4fa3-939e-ae404356d49f.png)

Right result for DEC 64.488:
![grafik](https://user-images.githubusercontent.com/12861218/151819626-f0109e9d-c247-4ded-b6ec-01f5cff3bb9a.png)

**Device and Application Information**
 - OS Build: 10.0.19041.0
 - Architecture: X64
 - Application Version: 10.2103.8.0
 - Region: de-DE
 - Dev Version Installed: False

**Additional context**

**Requested Assignment**

I'm just reporting this problem.  I don't want to fix it.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

seanmoran1701:
I have been trying to work on this issue, however, it seems to be a bigger problem than just 1,000,000,000/64487. The calculation always results in zero when any form of this calculation multiplied by 10 is entered. For example, 10,000,000,000/644870 also outputs zero.

IanParks-Coder:
> 1,000,000,000

[...]

Issue #1583, opened 2021-06-12
Division bug in Programmer Calculator DEC mode


korenkonder:
**Describe the bug**
There's weird bug when you divide 4294967296 (0x100000000) by 255 (0xFF) in DEC mode. The output is 0
Also happens with:
4294967297 (0x100000001)
4294967298 (0x100000002)
4294967299 (0x100000003)
4294967300 (0x100000004)
4294967301 (0x100000005)
4294967302 (0x100000006)
4294967303 (0x100000007)

**Steps To Reproduce**
DEC way
1. Go to Programmer Calculator
2. Click on `DEC`
3. Enter `4294967296`
4. Press `/` (Divide)
5. Enter `255`
6. Press `=` (Result)
7. See error

HEX way
1. Go to Programmer Calculator
2. Click on `HEX`
3. Enter `100000000`
4. Press `/` (Divide)
5. Enter `FF`
6. Click on `DEC`
7. Press `=` (Result)
8. See error

**Expected behavior**
Result should be 16843009 (0x1010101)

**Screenshots**
Screenshots may be not enough so I had recorded video
https://www.youtube.com/watch?v=J4If2A0BnQo

**Device and Application Information**
 - OS Build: 10.0.19043.0
 - Architecture: x64
 - Application Version: 10.2103.8.0
 - Region: ru-RU
 - Dev Version Installed: False

**Requested Assignment**
I'm just reporting this problem.  I don't want to fix it.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

randy408:
I can confirm the same issue with Calculator 10.2103.8.0.

bahlea:
#1788
I reported another one of those number combinations before stumbling accross this one.

rogitech:
I noticed the remainder of pret after remrat(&pret, rat_one) is not in the same scale as px, in this case, so subrat(px, pret, precision); will produce a wrong result. 
By doing another flatrat(pret, radix, precision); solves the issue (and #1788 as well). The only problem is this fix breaks one of the unit tests - UnitTest: CalculatorManagerTestScientific2 (commands21). (And I guess it's not ideal to change intrat()). 

void intrat(_Inout_ PRAT* px, uint32_t radix, int32_t precision)
{
    // Only do the intrat operation if number is nonzero.
    // and only if the bottom part is not one.
[...]

PR #2300, opened 2025-02-18
Bugfix: Division Bug in Programmer Dec Mode

ishuah:
## Fixes #.
This PR fixes a division bug in the Programmer Dec mode. The bug is described in #1583, #1788, and #1993.

- Added a check in the [_intrat_ function](https://github.com/ishuah/calculator/blob/224ccdf885b2f4a10722c9cb529a8a782727a7cb/src/CalcManager/Ratpack/support.cpp#L307) that ensures ***px** and **pret** have the same denominator.
- Added a [test case for this particular scenario](https://github.com/ishuah/calculator/blob/224ccdf885b2f4a10722c9cb529a8a782727a7cb/src/CalculatorUnitTests/CalculatorManagerTest.cpp#L756C9-L759C93).

- Added a test case for this scenario
- Ran the test suite locally

### DIFF
--- a/src/CalculatorUnitTests/CalculatorManagerTest.cpp
+++ b/src/CalculatorUnitTests/CalculatorManagerTest.cpp
@@ -752,6 +752,11 @@ namespace CalculatorManagerTest
 
         Command commands10[] = { Command::ModeProgrammer, Command::Command1, Command::CommandRORC, Command::CommandRORC, Command::CommandNULL };
         TestDriver::Test(L"-9,223,372,036,854,775,808", L"RoR(RoR(1))", commands10, true, false);
+
+        Command commands11[] = { Command::ModeProgrammer, Command::CommandDec, Command::Command4, Command::Command2, Command::Command9,   Command::Command4,
+                                Command::Command9,       Command::Command6,   Command::Command7, Command::Command2, Command::Command9,   Command::Command6,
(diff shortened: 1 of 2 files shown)


src/CalculatorUnitTests/CalculatorManagerTest.cpp
```diff
@@ -752,6 +752,11 @@ namespace CalculatorManagerTest
 
         Command commands10[] = { Command::ModeProgrammer, Command::Command1, Command::CommandRORC, Command::CommandRORC, Command::CommandNULL };
         TestDriver::Test(L"-9,223,372,036,854,775,808", L"RoR(RoR(1))", commands10, true, false);
+
```
tian-lt, line 755: Suggest adding a few more test cases mentioned in #1788

ishuah: Sounds good, I'll add more test cases.

ishuah: I've added 3 test cases from #1788 and #1583

tian-lt (approved):
LGTM. Adding @hanzhang54 for any additional comments.

hanzhang54:
> LGTM. Adding @hanzhang54 for any additional comments.

LGTM. Thank you very much for fixing it.

hanzhang54 (approved):

PR #2302, opened 2025-02-21
Work around a currency service defect for localization

tian-lt:
## What
The currency service seems have changed their fallback rules for unrecognized language code about Chinese, which causes users who selected zh-Hant-TW get simplified Chinese content in the currency converter.

## Fix
Regarding the standard of BCP47, manually construct a two-segment language code (e.g. zh-TW) from multi-segment language code (e.g. zh-Hant-TW) to ensure the localization service can recognize the language code and provide the correct content.

### DIFF
--- a/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
+++ b/src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
@@ -75,7 +75,7 @@ namespace
     // TODO: unit testing.
     std::optional<std::wstring> ParseLanguageCode(const wchar_t* bcp47)
     {
-        // the IETF BCP 47 language tag syntax is: language[-script][-region][-variant[*][-extension*][-privateuse*]]
+        // the IETF BCP 47 language tag syntax is: language[-script][-region]...
         std::vector<std::wstring> segments;
         std::wstring cur;
         // TODO: use C++20 - ranges::views::split_view in the future.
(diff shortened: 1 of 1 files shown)


src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -31,41 +33,84 @@ using namespace Windows::System::UserProfile;
...
+
+    // Fallback default values.
+    constexpr auto DEFAULT_FROM_CURRENCY = DefaultCurrencyCode.data();
+    constexpr auto DEFAULT_TO_CURRENCY = L"EUR";
+
+    // ParseLanguageCode returns language code in form of `lang-region`
+    // TODO: unit testing.
+    std::optional<std::wstring> ParseLanguageCode(const wchar_t* bcp47)
+    {
+        // the IETF BCP 47 language tag syntax is: language[-script][-region][-variant[*][-extension*][-privateuse*]]
```
guominrui, line 78: nit: extra `[`?

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
```diff
@@ -105,14 +150,9 @@ CurrencyDataLoader::CurrencyDataLoader(const wchar_t* forcedResponseLanguage)
...
-            std::wstring_view responseLanguage(m_responseLanguage->Data(), m_responseLanguage->Length());
-            std::match_results<std::wstring_view::const_iterator> match;
-            if (std::regex_match(responseLanguage.cbegin(), responseLanguage.cend(), match, std::wregex(L"zh-hans-[a-z]+", std::regex_constants::icase)))
+            if (auto lang = ParseLanguageCode(GlobalizationPreferences::Languages->GetAt(0)->Data()); lang.has_value())
             {
-                m_responseLanguage = L"zh-CN";
+                m_responseLanguage = ref new String{ lang->c_str() };
             }
         }
         else
```
hanzhang54, line 158: Can we remove this?

src/CalcViewModel/DataLoaders/CurrencyDataLoader.cpp
tian-lt: removed the trailing parts that this parser doesn't care.

hanzhang54 (approved):

PR #2310, opened 2025-03-04
Update Resources.resw a few Turkish translation fixed

good2see:
-a few Turkish translation fixed


tian-lt:
Thank you for suggesting the proper translation of Turkish, @good2see.
However, we can't accept the PR directly because Localization needs a special process to go through within Microsoft.
Here's the representative PR(#2351) inspired by your suggestions.  Thank you and we really appreciate your help.

PR #2351, opened 2025-07-04
Fix strings for Turkish

tian-lt:
## What
Thanks to @good2see for proposing PR #2310.
Updating strings in Calculator needs a special process and can't be done by accepting the PR from community. Therefore, this PR is created to represent the work of PR #2310 and fix the wrongly translated Turkish strings.


hanzhang54:
For languages ​​other than en-us, it is not recommended to update the resource file directly, as it could be automatically overwritten by the localization pipeline. I'd suggest reporting it to the localization team to get the localization source updated.

tian-lt:
@hanzhang54 the changes in this are being applied from the loc pipeline

Issue #2253, opened 2024-11-22
[Localization] Calculator UI did not use Windows display language but forced to a Regional settings


ssh4net:
**Describe the bug**
Well, an issue do not specific to a Calculator, but for almost all Microsoft apps.

Microsoft apps use a regional settings and set the language based on region instead of display language.

**Steps To Reproduce**
- Change a region to Chinese
- Struggle to find how change Display language to Englsih
- Optionally remove English keyboard (Asian keyboards have English input and do not required English language for keyboard)
- Restart and open Calculator

**Expected behavior**
App must use Display language. Or have an option to switch a language from a available list

**Screenshots**
![Image](https://github.com/user-attachments/assets/4202b3da-bdde-46de-b211-1c65e62d2b58)
![Image](https://github.com/user-attachments/assets/71d849ec-b31b-4d5c-aa89-5335cbe2cc59)

**Device and Application Information**
 - OS Build: Any Windows (screenshots from Windows 10 Pro)
 - Architecture: any arcitect\ure
 - Application Version: 11.2409.0.0
 - Region: ja-JP ( but can be any non English)
 - Dev Version Installed: -
 
**Additional context**
In the world, there are about 37 spoken languages. Due to globalization, native speakers can live in countries with different languages.
That creates a situation when people can use 2 or 3 or even more languages with a different level of knowledge.
Due to regional restrictions or non-UTF legacy, it is often safer to have correctly set regional settings in a Windows system.
For example, lining in Japan, but not be a tourist, it is better to have a region set to Japan.
It is good that Microsoft allows the use of a display language that is different from the region.
That creates a situation when a system has Japanese region, English (US) as a display language.

PR #2369, opened 2025-08-15
Add language selection options

ssh4net:
When Windows itself allow to have Display language different from Regional settings UWP apps always forced to use regional settings and/or preferred languages. Some Regions has full English keyboard in own language (Japanese Romaji for example) so English might be removed everywhere except Display language settings.
This fix should allow change app language to user choice

https://github.com/microsoft/calculator/issues/2253

I'm not so familiar with UWP, but no one cared about adding this feature, so this is my attempt. If someone can brush up it, this is really help a lot of expats (or downshifts) in the world.

## Fixes #.
Added language selection support

- LanguageHelper utility to manage application language settings, preferring system display language over regional settings.
- Language selection UI in Settings.

- Manual, app just work


ssh4net:

ssh4net:
@mingxwa can you make Windows respect its own features?

I did not use any hacks or tweaks.
System allow to set English or whatever language as a display language.
System allow me to remove duplicated English keyboard.
System allow me to set Japan as a region, and this is important for different apps to.

But only UWP and QT think that they are know better what user want.

I understand that my fix is more like a bug fix for what is broken for decade in all windows systems.

PR #2421, opened 2025-12-06
Fix button bindings

SiliconA-Z:
Button ids were bounded to the wrong values


Sepharo:
Wasn't it correct before?
If the current button of QWORD is clicked, and that QWORD button has a command with a parameter of 0... then handling the command in OnBitLengthButtonPressed with a parameter of 0, means it needs to move to the next button, which is DWORD.

Issue #574, opened 2019-07-05
Incorrect result for exponational and logarithmic functions with exact results


dovisutu:
**Describe the bug**
The below example shows a case that shown answer (2) doesn't match the real answer
log(e^2)/log(e)-2 != 0

**Steps To Reproduce**

1. Go to `Scientific mode`
2. enter log(e^2)/log(e^1)-2

**Expected behavior**
log(e^2)/log(e^1)-2 should be 0 as *base changing formula for logarithms*:
![LaTex](https://user-images.githubusercontent.com/40313014/60698458-05694400-9f22-11e9-9098-d2d3ca318c51.JPG)

**Screenshots**
![non-zero result](https://user-images.githubusercontent.com/40313014/60698475-17e37d80-9f22-11e9-878a-69efc957b10d.JPG)

**Device and Application Information**
 - OS Build: 10.0.17134.0
 - Architecture: X64
 - Application Version: 10.1905.28.0
 - Region: zh-CN
 - Dev Version Installed: False

**Additional context**
This is a new issue related with #309 and #222 
**Requested Assignment**
I'm just reporting this problem.  I don't know how to fix it.

SAI990323:
This is a matter of precision, I think you may not hope that the calculator will automate simplify the  formula for you and if we use float number, this problem may be inevitable.

dovisutu:
The issue is, if I delete "-2" part, the result is 2.
This may give somebody the thought,"*How strange, why didn't 2-2 = 0?*"
Also, there are two related issues mentioned, both of them are the case a-a != 0.

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

grochocki:
Related to #222 and #309. Relevant article about this class of bugs in the [calc engine](https://devblogs.microsoft.com/oldnewthing/?p=39193).

SAI990323:
> The issue is, if I delete "-2" part, the result is 2.
> This may give somebody the thought,"_How strange, why didn't 2-2 = 0?_"
> Also, there are two related issues mentioned, both of them are the case a-a != 0.

sorry I misunderstand it.

VIGNESHinZONE:
Can i work on this issue and could you pls tell me the files that i can work on.

grochocki:
Thanks for taking a look! @joseartrivera, can you point @VIGNESHinZONE in the right direction here?

Issue #309, opened 2019-03-14
Inaccurate results for trigonometric functions with exact results


imchau:
**Describe the bug**
The below examples of the trigonometric functions show a correct result at first. However, subtracting that result from itself results in a non-zero value, which shows that the results are represented as incorrect approximations internally, when it should be an exact result.

In DEG mode:
arcsin(1) - 90 != 0
sin(30) - 0.5 != 0
In RAD mode:
sin(π/2) - 1 != 0
but interestingly: cos(±π) + 1 == 0

This is a new issue spawned from issue #222. 

**Steps To Reproduce**
<!--Steps to reproduce the behavior:
1. Try the above calculations

**Expected behavior**
Exact internal representation and result 0 for above calculations.

**Device and Application Information (please complete the following information):**
* OS Build: 10 0 17763 0
* Architecture: X64
* Application Version: 10.1812.10048.0

MicrosoftIssueBot:
This is your friendly Microsoft Issue Bot. I've seen this issue come in and have gone to tell a human about it.

uhliksk:
Some situations are not solvable by any conventional way, for example sin ( 42 ) - ( sqrt( 30 + sqrt( 180 ) ) - sqrt( 5 ) + 1 ) / 8 in DEG mode.

grochocki:
Thanks for creating the report! We should defintiely try to squash these.

PR #2424, opened 2025-12-07
Fix #222: Wrong result for e.g. √(2.25) - 1.5 #222

### LINKED ISSUE #222: Wrong result for e.g. √(2.25) - 1.5

guihuaz:
- add a helper function _snaprat to check if a result RAT should be zero with given precision. Apply _snaprat to addrat/subrat/lograt.
- The root causes of issues #222, #309, #574 are the same. For an irrational number result, the approximation introduces small errors beyond precision required, which is expected. The issue was Ratpak tried to take "precision" digits from the calculation result regardless where the result came from, and it exposed digits from approximation errors. Take `sqrt(2.25) - 1.5` for example. `sqrt(2.25)` might be approximated to `1.499...998...` depending on the precision. When this number subtracted `1.5`, the Ratpak got `0.000...001....` Ratpak took up to precision number of digits and converts to scientific format, which resulted `-1.2566515751494548126925332972115e-47`. This PR checks the magnitude difference between the result and the input operands. Since the result is too close to `47` magnitude smaller than original subtraction operands `1.499...998...` and `1.5`, it would be snapped to `0`.
- The same fix applies to `lograt` as well. 
- One further optimization could be a new sqrt function specifically checking perfect square first before using Taylor series approximation.

- Add 1 test case in `CalculatorManagerTestStandard` to verify addrat. 
- Add 2 test cases in `CalculatorManagerTestScientific` to verify addrat and lograt
- Manual test the cases in the linked issues.
- Manual test log case: `log(sqrt(2.25)/1.5)`

### DIFF
--- a/src/CalcManager/Ratpack/rat.cpp
+++ b/src/CalcManager/Ratpack/rat.cpp
@@ -366,9 +366,8 @@ void _snaprat(_Inout_ PRAT* pr, _In_ PRAT a, _In_opt_ PRAT b, int32_t precision)
         }
         else
         {
-            DUPRAT(threshold, absA); 
+            DUPRAT(threshold, absA);
         }
-        
         destroyrat(absA);
         destroyrat(absB);
     }


tian-lt:
> @tian-lt, are you the maintainer of this repo? Could you please take a look at the PR?

@guihuaz Sure. Thank you for proposing the fix.
Let me trigger a CI run for this PR.
I'll review the changes and get back to you later.

tian-lt:
@guihuaz 
Shall we adjust the result in the `Rational` class or a higher context? Because treating the epsilon as zero is more like an application requirement rather than a correct behavior of semi-numerical analysis.

tian-lt (approved):

tian-lt:
The changes look good to me.
@HuanglinXu21 @guominrui, could you help take a look?

axtar:
It seems that this PR has broken lograt. For example log of 9.e+99 now returns 0 instead of 99.954...

guihuaz:
> It seems that this PR has broken lograt. For example log of 9.e+99 now returns 0 instead of 99.54...

@axtar do you have dev env? If so, could you please try to try removing _snaprat inside lograt to see if it fixes the issue? I'm not on my windows machine, right now. I might have time to try this weekend.

axtar:
@guihuaz removing _snaprat in lograt fixes the problem but then of course the problem with log( tan₀( 45 ) ) reappears.

axtar:
@guihuaz As a temporary solution, I added a check in my project so that _snaprat only snaps to zero if absR is also smaller than rat_smallest.

src/CalcManager/Ratpack/rat.cpp
```diff
@@ -285,3 +313,76 @@ bool zerrat(_In_ PRAT a)
...
+        destroyrat(absB);
+    }
+    mulrat(&threshold, rat_smallest, precision);
+
+    PRAT absR = nullptr;
+    DUPRAT(absR, *pr);
+    ABSRAT(absR);
+
+    // if absResult < threshold => snap to zero
+    if (rat_lt(absR, threshold, precision))
```
guihuaz, line 381: (@axtar , move comment here to have a thread)
Do you add a check here? I think this is safe change. Could you please send a PR to add the additional check absR < rat_smallest?

It may expose some other edge cases back, but _snaprat does not likely catch all these cases. In my understanding, the issue is fundamental design flaw in Ratpack. Ratpack, as the name implies, is for arbitrary precision rational math. But it tries to include some irrational math using approximation and rounding. I think this would inevitably introduce errors.

axtar: Yes, I have added an “&&” to that “if”.
I’m sorry but I don’t have the skills to contribute to this repository. I don’t even really know how to make a pull request. I’m simply using ratpak on a microcontroller for a hobbyist hardware calculator project. That’s why I found the issue.
Approximation and rounding issues are difficult to fix but for a calculator arbitrary precision is always better than 64-bit floating point arithmetic. I really like ratpak, it's small and very efficient. But it has a lot of memory leaks that I had to fix to run it on a microcontroller.

axtar: I have opened an issue for this

PR #2430, opened 2026-01-18
Add GetCurrentNumWidth() method to expose current bit width setting

dongfengweixiao:
I am currently developing a [calculator app using Flutter](https://github.com/dongfengweixiao/win_calculator), and I noticed that the current calcmanager only provides a function to set the bit width, but does not provide a function to get the bit width.


dongfengweixiao:

PR #2452, opened 2026-03-30
Fix implicit multiplication: handle cases like (25)2 → 25*2

### LINKED ISSUE #2451: [Bug] Implicit multiplication after closing parenthesis causes preceding operand to be dropped from computation
**Describe the bug**

900+(25)2=50
900+(25)*2=950
if you omit typing the * it ignores the addition.
it is visually the same in the calculator but the result is different.

**Steps To Reproduce**
1. Go to Scientific mode
2. make an equation omitting the use of * in a calculation with parentheses like 900+(25)2
3. Press = button
4. Wrong result

**Expected behavior**
give same result as 900+(25)*2=950

**Screenshots**

Image

Image

 - OS Build: 10     0      26200  0
 - Architecture: x64
 - Application Version: 11.2508.4.0
 - Region: en-US

**Requested Assignment**

adithya-naik:
- Added handling for implicit multiplication after parentheses, e.g., `(25)2` → `(25)*2`.
- Updated parser logic to insert `*` automatically when a number follows a closing parenthesis.
- This ensures expressions like `900+(25)2` evaluate correctly as `950`.

- Full Calculator app was not built due to local environment constraints, but logic matches expected results.

### DIFF
--- a/src/CalcManager/CEngine/scicomm.cpp
+++ b/src/CalcManager/CEngine/scicomm.cpp
@@ -196,6 +196,9 @@ void CCalcEngine::ProcessCommandWorker(OpCode wParam)
         // Check if the last command was a closing parenthesis
         if (m_nLastCom == IDC_CLOSEP)
         {
+            if{
+                ResolveHighestPrecedenceOperation();
+            } 
             // Treat this as an implicit multiplication
             m_nOpCode = IDC_MUL;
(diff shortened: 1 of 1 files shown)


adithya-naik:
Hi Team,

This is just a follow up remainder on this PR

PR #2466, opened 2026-05-22
Refine snapping condition for _snaprat

### LINKED ISSUE #2439: PR #2424 leads to incorrect lograt results in some cases
**Describe the bug**
PR #2424 introduces an issue with lograt that could lead to wrong results.

**Steps To Reproduce**
log of 1.e+50 returns 0 instead of 50

**Expected behavior**
log( 1.e+50 ) = 50

**Application Version**
current main branch

**Requested Assignment**

### LINKED ISSUE #2484: binary logarithm of a factorial larger than 30! outputs "0" or "Result is undefined"
**Describe the bug**

when you enter a number and press exclamation point, you get the factorial, then you should be able to press "2nd" and then the number "2" and the equal sign, you get back how many bits you need to represent that factorial.

**Steps To Reproduce**

"512"
"!"
"2nd"
"Log yX"
"2"
"="

**Expected behavior**

should not say "Result is undefined" or "0", but a normal number.

**Screenshots**

 - OS Build:
Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      26200  0
 - Architecture:
X64
 - Application Version:
11.2605.9.0
 - Region:
en-US
 - Dev Version Installed:
False

**Additional context**

it worked perfectly in the earlier versions, but I don't know when automatic updates actually triggers at the moment, so I don't know how many versions may have released since the error was introduced.

**Requested Assignment**

axtar:
Fixes #2439, Fixes #2484
In the specific case of natural logarithms, _snaprat does not work correctly and returns 0 for values ⪆ 7.9e+33. This affects not only ln, but also log and logyx, as these also use natural logarithms.
Examples:
ln of 7.8e+33 = 78.039 -> ok
ln of 7.9e+33 = 0 -> wrong, should be 78.052
log of 7.9e+33 = 0 -> wrong, should be 33.898
log of 1.e+50 = 0 -> wrong, should be 50
logyx of 7.9e+33 base 2 = 0 -> wrong, should be 112.605
logyx of 7.9e+33 base 7.9e+33 = result is undefined -> wrong, should be 1
logyx of 2 base 7.9e+33 = cannot divide by zero -> wrong, should be 0.009

This PR helps to mitigate these calculation errors by adding an additional check to _snaprat. A result with an absolute value greater or equal than rat_smallest should not be set to zero anyway.

### DIFF
--- a/src/CalcManager/Ratpack/rat.cpp
+++ b/src/CalcManager/Ratpack/rat.cpp
@@ -377,8 +377,8 @@ void _snaprat(_Inout_ PRAT* pr, _In_ PRAT a, _In_opt_ PRAT b, int32_t precision)
     DUPRAT(absR, *pr);
     ABSRAT(absR);
 
-    // if absResult < threshold => snap to zero
-    if (rat_lt(absR, threshold, precision))
+    // if absResult < threshold && absResult < rat_smallest => snap to zero
+    if (rat_lt(absR, threshold, precision) && rat_lt(absR, rat_smallest, precision))
     {
         DUPRAT(*pr, rat_zero);
     }


axtar:
solution proposed by claude:
``` C++
void lograt(_Inout_ PRAT* px, int32_t precision)

{
    _lograt(px, precision);
    _snaprat(px, rat_one, nullptr, precision);
}
```
I close my PR...

Issue #2325, opened 2025-04-28
Memory leaks in ratpak


axtar:
**Describe the bug**
There are some memory leaks in the ratpak library. For example, each non integer factorial calculation leaks about 800 bytes.
Here is a list of PRAT that are not destroyed:

-	Function _gamma: ratRadix, mpy
-	Function factrat for negative integers: fact, frac, neg_rat_one 
-	Function powratNumeratorDenominator:
 	if first call to powratcomp throws an error: yNumerator, yDenominator, pxPow
        if second call to powratcomp throws an error: originalResult, oneoveryDenom, yNumerator, yDenominator, pxPow
        if third call to powratcomp throws an error: originalResult, oneoveryDenom, yNumerator, yDenominator, pxPow, roundedPower, roundedResult
-	Function powratcomp:
        if fBadExponent is true: podd, pxint
        if rattoi32 throws an error: iy, podd, pxint 
-	Function rootrat if call to powrat throws an error: oneovern

**Device and Application Information**
- OS Build: 10     0      26200  0
- Architecture: X64
- Application Version: 11.2508.4.0
- Dev Version Installed: false

**Requested Assignment**

If possible, I would like to fix this.

axtar:
only relevant when using ratpak on a micro controller, not relevant for windows with GB of memory

axtar:
Since work is currently being done on ratpak, perhaps someone can also fix the memory leaks.

axtar:
None of this makes sense anymore. Just let Claude search for memory leaks and he'll find and fix them all. I'm closing this issue.

PR #2467, opened 2026-05-22
Fix memory leaks in _gamma and factrat

axtar:
This PR fixes the memory leaks described in issue #2325 for factorial calculations. Memory leaks occur when calculating factorials of non-integer numbers and of negative integers.


axtar:
None of this makes sense anymore. Just let Claude search for memory leaks and he'll find and fix them all.

PR #2474, opened 2026-06-11
Fix focus after clearing a memory/history item via context menu (a11y)

### LINKED ISSUE #312: Focus moves to "Clear all memory" button after activating "Clear memory item" from the context menu of "Memory" list item.

gebrilfradj:
## Fixes #312
 
 - When a user clears a memory item (“Clear memory item”) or deletes a history item (“Delete”) from a list item’s context menu, keyboard and Narrator focus now consistently transitions to an appropriate and intuitive target. Previously, focus could fall through to the “Clear all memory” or “Clear all history” button, or—if the last remaining item was removed—escape the pane entirely and land on the hamburger menu.
 
 - Focus behavior is now context-aware. When clearing or deleting an item that is not the last in the list, focus moves to the item that takes its place. When removing the only remaining item, focus is redirected based on the pane:
   - **Memory pane:** Focus moves to the M+ (memory add) button, with the flyout closing first in narrow layouts so the control is reachable.
   - **History pane:** Focus moves to the History/Memory pivot in the docked layout, or to the History toggle button in the narrow flyout.
 
 - This update fixes a shared accessibility issue across both the Memory and History panes by addressing a single root cause in the view layer. It introduces new `MemoryEmptied` and `HistoryEmptied` events that are raised before the item is removed, ensuring that focus never briefly lands on a disappearing element such as the trash button. The implementation is contained within the View layer (`Calculator.xaml.cs`, `Memory.xaml.cs`, `HistoryList.xaml.cs`) and requires no changes to XAML, ViewModels, or CalcManager; it follows the existing `Focus(FocusState.Programmatic)` pattern.
 
 - This change also revisits the approach from previously closed PR #1605, which was closed due to staleness during the C++ → C# UI migration rather than a rejection of the approach. The solution has been adapted to the current C# code-behind architecture, expanded to cover the History pane, and strengthened with automated test coverage.
 
 - New WinAppDriver UI tests cover focus behavior after clear/delete in both Memory and History panes across docked and narrow (flyout) layouts. All new and adjacent tests pass locally (9/9):
   - `StandardMemory_Panel_FocusMovesToMemoryPlusAfterClearingOnlyItem`
   - `StandardMemory_Panel_FocusMovesToNextItemAfterClearingFirstOfMultiple`
   - `StandardMemory_Flyout_FocusMovesToMemoryPlusAfterClearingOnlyItem`
   - `StandardHistory_Panel_FocusDoesNotFallBackToClearAllAfterDeletingOnlyItem`
   - `StandardHistory_Flyout_FocusMovesToHistoryButtonAfterDeletingOnlyItem`
 - Full CI on the prior head (Build x64, Run unit tests x64, Run UI tests x64) was green; awaiting CI on the latest commit before flipping to Ready for review.

### DIFF
--- a/src/Calculator/Views/Calculator.xaml.cs
+++ b/src/Calculator/Views/Calculator.xaml.cs
@@ -186,6 +186,23 @@ namespace CalculatorApp
                 };
                 historyVM.HideHistoryClicked += OnHideHistoryClicked;
                 historyVM.HistoryItemClicked += OnHistoryItemClicked;
+                m_historyList.HistoryEmptied += () =>
+                {
+                    // Raised before the last item is removed, so focusing here keeps focus from
+                    // briefly escaping (e.g. to the hamburger) while the list empties.
+                    if (HistoryButton.IsEnabled && HistoryButton.Visibility == Visibility.Visible)
+                    {
+                        // Narrow (flyout) layout: focus the History toggle button, mirroring
+                        // the focus target used when the history flyout closes.
+                        HistoryButton.Focus(FocusState.Programmatic);
+                    }
+                    else
+                    {
+                        // Docked layout: no History toggle button exists, so keep focus
+                        // within the docked History/Memory pivot instead of escaping.
+                        DockPivot.Focus(FocusState.Programmatic);
+                    }
+                };
             }
         }
 
@@ -742,6 +759,7 @@ namespace CalculatorApp
         private CalculatorApp.HistoryList m_historyList;
         private bool m_fIsHistoryFlyoutOpen;
         private bool m_fIsMemoryFlyoutOpen;
+        private bool m_focusMemoryPlusOnFlyoutClose;
 
         private void OnMemoryFlyoutOpened(object sender, object args)
         {
@@ -762,7 +780,14 @@ namespace CalculatorApp
         {
             m_fIsMemoryFlyoutOpen = false;
             EnableControls(true);
-            if (MemoryButton.IsEnabled)
+            if (m_focusMemoryPlusOnFlyoutClose)
+            {
+                // The flyout was closed because its last memory item was cleared; move focus
+                // to M+ on the now-visible keypad rather than the (disabled) memory button.
+                m_focusMemoryPlusOnFlyoutClose = false;
+                MemPlus?.Focus(FocusState.Programmatic);
+            }
+            else if (MemoryButton.IsEnabled)
             {
                 MemoryButton.Focus(FocusState.Programmatic);
             }
@@ -796,6 +821,21 @@ namespace CalculatorApp
             if (m_memory == null)
             {
                 m_memory = new Memory();
+                m_memory.MemoryEmptied += () =>
+                {
+                    // Raised before the last item is removed. Focus M+ up front so focus never
+                    // escapes to the hamburger while the list empties.
+                    MemPlus?.Focus(FocusState.Programmatic);
+
+                    if (m_fIsMemoryFlyoutOpen)
+                    {
+                        // Narrow (flyout) layout: the keypad sits behind the open flyout. Close
+                        // it and re-assert M+ focus once it has fully closed (OnMemoryFlyoutClosed),
+                        // so focus isn't handed to the now-disabled memory button.
+                        m_focusMemoryPlusOnFlyoutClose = true;
+                        MemoryFlyout.Hide();
+                    }
+                };
                 VisualStateManager.GoToState(m_memory, GetCurrentLayoutState(), true);
             }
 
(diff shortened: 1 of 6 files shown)


gebrilfradj:
@microsoft-github-policy-service agree company="Microsoft"

src/Calculator/Views/Memory.xaml.cs
```diff
@@ -60,8 +66,28 @@ private void MemoryListItemClick(object sender, ItemClickEventArgs e)
...
         {
             var memoryItem = GetMemoryItemForCurrentFlyout();
-            if (memoryItem != null)
+            if (memoryItem == null)
+            {
+                return;
+            }
+
+            // Capture the position before clearing; the container is torn down on removal.
+            var clearedIndex = MemoryListView.IndexFromContainer(MemoryContextMenu.Target);
```
veler: We could execute this line only when `itemsRemainingAfterClear > 0` is true, since we don't use this variable in the `else`

src/Calculator/Views/HistoryList.xaml.cs
```diff
@@ -67,7 +73,25 @@ private void OnDeleteMenuItemClicked(object sender, RoutedEventArgs e)
             var listViewItem = HistoryContextMenu.Target;
             if (HistoryListView.ItemFromContainer(listViewItem) is HistoryItemViewModel itemViewModel)
             {
-                Model.DeleteItem(itemViewModel);
+                // Capture the position before deleting; the container is torn down on removal.
+                var deletedIndex = HistoryListView.IndexFromContainer(listViewItem);
```
veler: Same thing here

src/Calculator/Views/Calculator.xaml.cs
```diff
@@ -186,6 +186,23 @@ public void InitializeHistoryView(CalculatorApp.ViewModel.HistoryViewModel histo
                 };
                 historyVM.HideHistoryClicked += OnHideHistoryClicked;
                 historyVM.HistoryItemClicked += OnHistoryItemClicked;
+                m_historyList.HistoryEmptied += (s, e) =>
```
veler: Let's move this into its own method instead of making a lambda event handling.

src/CalculatorUITests/MemoryFunctionalTests.cs
```diff
@@ -136,6 +136,92 @@ public void StandardMemory_Flyout()
             Assert.IsNotNull(CalculatorDriver.Instance.CalculatorSession.FindElementByAccessibilityId("MemoryPaneEmpty"));
         }
 
+        /// <summary>
+        /// Bug 20774908: verifies focus lands on the M+ keypad button (not "Clear all
```
veler: Let's reference GitHub issue 312 instead of the Azure DevOps work item.

emmche: +1 Or remove reference to specific bugs and just have a description of the scenario

src/Calculator/Views/Calculator.xaml.cs
```diff
@@ -796,6 +823,21 @@ private Memory GetMemory()
             if (m_memory == null)
             {
                 m_memory = new Memory();
+                m_memory.MemoryEmptied += (s, e) =>
```
veler: Same thing here please. Let's create a method instead of lambda

veler (approved):

src/Calculator/Views/Memory.xaml.cs
gebrilfradj: Fixed in afcd36d: moved the index capture inside the > 0

src/Calculator/Views/HistoryList.xaml.cs
gebrilfradj: Fixed in afcd36d: same fix from last comment

src/Calculator/Views/Calculator.xaml.cs
gebrilfradj: Fixed in afcd36d: extracted to OnHistoryEmptied next to the other named handlers

emmche (approved):

src/CalculatorUITests/MemoryFunctionalTests.cs
gebrilfradj: Fixed in afcd36d: all five new test docstrings now reference issue #312 instead of the internal work item.

src/Calculator/Views/Calculator.xaml.cs
gebrilfradj: Fixed in c1dc614: extracted to OnMemoryEmptied, mirroring the OnHistoryEmptied extraction.

PR #2497, opened 2026-08-31
Fixes 594

### LINKED ISSUE #594: Collapse unit converters into single page and refresh layout
**Problem Statement**
Today, we treat each converter type as a high-level navigation item. This means that if you want to switch between volume and weight converters, you need to navigate to an entirely different page. While we do offer shortcuts to switch between modes, they are not very discoverable (#157), and parsing a mixed list of calculators and converters is unnecessarily more complex than it needs to be. Furthermore, the converter page interface is laid out differently than other calculator modes and feedback suggests that the interface could be made more intuitive.

image

**Evidence or User Insights**
We currently have a flat navigation structure. According to [app navigation guidance](https://docs.microsoft.com/en-us/windows/uwp/design/basics/navigation-basics), this structure is appropriate when there are less than 8 pages in the group. We already have 13 converters, and #515 is looking at evaluating which converters and units we include, and suggests we may actually be _missing_ some converter types (increase to 17). In this case, hierarchical or an alternative navigation would be preferred.

Some other UX issues have been identified as well. It is unclear how to quickly switch units (#172 and #266). There is some additional conversation in #382 and #379 as well.

**Proposal**
Collapse converters into a single "Converter" page where you can select the type of conversion in addition to the units. There may be an opportunity to streamline input even further with #589 to allow users to input any unit they like without having to explicitly select a converter type first.

**Goals**
* Collapse top-level converter navigation into a single Converter (open question: does currency converter feel different enough to remain separate?)
* Redesign page layout for greater consistency with other calculator modes
* Redesign value input to align with other converter patterns (e.g., found in translation apps)

**Non-Goals**
* Determining which converters/units to include is out-of-scope (covered by #515)

**Low-Fidelity Concept**
image

Bing Translator/Converter UX (for inspiration)
bing

image

**Requested Assignment**

veler:


mmdparsa-dev:
The best part is that its written in c#

